PR #49, opened 2014-06-10
a=imageattr negotiation specification. Fixes #5

### LINKED ISSUE #5: Handling of a=imageattr
Section 5.2.1
   o  [OPEN ISSUE:  Handling of a=imageattr]

alvestrand:
This change adds a=imageattr generation, parsing, and guidelines for usage.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -535,97 +535,6 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           with a timer; at this point, the application could reapply the
           existing remote description as a final answer.</t>
         </section>
-        <section anchor="sec.imageattr" title="Video size negotiation">
-          <t>Video size negotiation is the process of using the "a=imageattr"
-          SDP attribute<xref target="RFC6236"/> to indicate:</t>
-
-          <t><list style="symbols">
-            <t>From the sender to the receiver: What video frame sizes it is
-            capable of sending, and what video frame sizes it prefers to
-            send</t>
-
-            <t>From the receiver to the sender: What video frame sizes it is
-            capable of receiving, and what video frame sizes it prefers to
-            receive</t>
-          </list>Since "a=imageattr" is defined to be an advisory field, this
-          does not absolutely constrain the video formats that the sender can
-          use, but gives an indication of the preferred values.</t>
-
-          <section title="Choice of imageattr sub-attributes">
-            <t>The "a=imageattr" field is described as payload type specific.
-            When all video codecs supported have the same capabilities, use of
-            the wildcard payload type (*) is RECOMMENDED.</t>
-
-            <t>If video codecs have differing capabilities, the specific payload
-            types are used, and there is one "a=imageattr" line per payload
-            type.</t>
-
-            <t>The "a=imageattr" field can contain multiple sets of parameters,
-            each with a specified q value (a preference). In this specification,
-            the preferences are used in the following manner:</t>
-
-            <t><list style="symbols">
-              <t>q=0.1 is used to indicate the decode capabilities of the
-              device, if known. Using image sizes outside these parameters
-              risks decode failure.</t>
-
-              <t>q=0.5 is used to indicate the limits to image resolution
-              caused by mandatory constraints set by the user (if any).</t>
-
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -576,6 +576,135 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+
+            <t><list style="symbols">
+              <t>a=imageattr:* recv [x=[16:1920],y=[16:1080],q=0.1]
+              [x=[320:640],y=[16:1080],q=0.5]</t>
+              </list>and if the stream is known to be displayed on a surface of
+            dimension 640x480, the result might be</t>
+
+            <t><list style="symbols">
+              <t>a=imageattr:* recv [x=[16:1920],y=[16:1080],q=0.1]
+              [x=[320:640],y=[16:1080],q=0.5] [x=640,y=480,q=1.0]</t>
```
juberti: This breaks when the display resolution is larger than the constraints or decoder limits. Consider a decoder that does max 720p, and then the video element is made fullscreen (say, 1080p). Then, since q=1.0 for 1080p, the sender sends 1080p and the decoder dies.

Reversing the q values of course doesn't solve this problem, since if the display window is 480p, we don't want the sender to send 720p (recall, the decoder limit).

It's not clear to me how the sender should interpret multiple imageattrs with different q= values - i.e. should it try to satisfy the most preferred one, and then work backward, or should it try to satisfy them all? My interpretation has been the former, so I think that what we would need to do is to do some form of ANDing, e.g in the fullscreen/1080p case
q=0.1 is max 720p
q=1.0 is max 720p (1080p ANDed with 720p)

or in the 480p case
q=0.1 is max 720p
q=1.0 is max 480p (720p ANDed with 480p)

draft-ietf-rtcweb-jsep.xml
```diff
@@ -576,6 +576,135 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          <section title="Choice of imageattr send attributes">
+            <t>There are two cases for generation of the send
+            attributes: When the correspondent's recv attribute is
+            unknown, and when the recv attribute is known.
+            </t>
+            <t>When the recv attribute is unknown, such as for an
+            initial offer, the send attribute is generated as follows:
+            </t>
+            <t><list style="symbols">
+              <t>q=0.1 is used to indicate the range of video sizes the
```
juberti: This may not be known well in the case of a software encoder, so we may just pick some arbitrary max (e.g. 4K)

draft-ietf-rtcweb-jsep.xml
```diff
@@ -576,6 +576,135 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            attributes, the lower-priority configurations might not be
+            subsets of the higher-priority ones, so they need to be
+            applied as "advanced" constraints, in order from highest
+            to lowest priority, in order to give a reasonable result
+            for all cases, and the application of some constraints
+            will likely fail.
+            </t>
+            <t>This will generate an "a=imageattr send" attribute
+            where the q=0.5 tier reflects how well the preferences of
+            the receiver could be accomondated, and where the q=1.0
```
juberti: accomodated

draft-ietf-rtcweb-jsep.xml
```diff
@@ -576,6 +576,135 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            before generating the send attribute. Note
+            that in the case of non-WebRTC-generated "recv"
+            attributes, the lower-priority configurations might not be
+            subsets of the higher-priority ones, so they need to be
+            applied as "advanced" constraints, in order from highest
+            to lowest priority, in order to give a reasonable result
+            for all cases, and the application of some constraints
+            will likely fail.
+            </t>
+            <t>This will generate an "a=imageattr send" attribute
```
juberti: I'm not sure I understand the purpose of a send attribute. The recv is pretty clear, but what good does it do to say what you are sending?

fluffy:
Justing is going to take a run at fixing up this PR to match last meeting

juberti:
Superseded by #150

PR #104, opened 2015-02-07
Use TCP/DTLS and UDP/TDTLS/SCTP in profiles to avoid confusion. Needs up...

### LINKED ISSUE #101: Profile names: TCP/DTLS/RTP/SAVPF and TCP/DTLS/SCTP
As per discussion in mmusic, to make it abundantly clear that DTLS is to be used even when running over ICE-TCP, the profile name should be TCP/DTLS/RTP/SAVPF. We'll leave the UDP/TLS/RTP/SAVPF one alone, since that ship has sailed, and it's less ambiguous.

ekr:
...date of draft-nandakumar. Fixes #101 

This is a merge of a previously approved PR


juberti:
lgtm

fluffy:
LGTM

PR #105, opened 2015-02-08
Remove special case for {local,remote}Description after close. Fixes #88

### LINKED ISSUE #88: localDescription and RemoteDescription should not return NULL, if RTCPeerConnection is closed.
In section 4.1.7 and 4.1.8, JSEP draft specifies that localDescription and remoteDescription should return NULL respectively, if RTCPeerConnection is closed.
But WebRTC working group discrards it as per the discussion in bug 21082.
(https://www.w3.org/Bugs/Public/show_bug.cgi?id=21082)

Inorder to make both the specs to be in sync, I think the text in JSEP draft should be modified from 

   "A null object will be returned if the local description has not yet
   been established, or if the PeerConnection has been closed."

to

   "A null object will be returned if the local description has not yet
   been established."

ekr:


juberti:
lgtm

fluffy:
lgtm

PR #106, opened 2015-02-09
Checkpoint: First cut at separate RTCP setting

ekr:
Justin, this isn't done, but PTAL and see if you think it's going in the right direction.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -733,13 +733,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           application can specify its preferred policy regarding use of RTP/RTCP
           multiplexing <xref target="RFC5761"/> using one of the following policies:
           <list style="hanging">
-            <t hangText="max-compat:">The browser will gather both RTP and
+            <t hangText="negotiate:">The browser will gather both RTP and
             RTCP candidates but also will offer "a=rtcp-mux", thus allowing
             for compatibility with either multiplexing or non-multiplexing endpoints.</t>
 
-            <t hangText="max-mux:">The browser will only gather RTP candidates
+            <t hangText="require:">The browser will only gather RTP candidates
             and when acting as answerer will reject any m= section that does not
-            provide an "a=rtcp-mux" attribute.</t>
+            provide an "a=rtcp-mux" attribute. This halves the number of candidates
+            that the offerer needs to gather.</t>
           </list>
           </t>
           
@@ -1072,9 +1073,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             immediately; if increased, additional candidates are pre-gathered;
             if decreased, the now-superfluous candidates are discarded.
             </t>
-            <t>Any changes to the BUNDLE or RTCP multiplexing policies take effect immediately, i.e.
-            any future tracks added to the PeerConnection will have their
-            bundle-only and RTP/RTCP multiplexing state marked accordingly.
+            <t>The BUNDLE and RTCP-multiplexing policies MUST NOT be
+            changed after the construction of the PeerConnection.
             </t>
             
           </list></t>
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1903,7 +1903,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
               <t>If present in the offer, an "a=rtcp-mux" line, as specified
               in <xref target="RFC5761"></xref>, Section 5.1.1. If the
-              "max-mux" RTCP multiplexing policy is set and no "a=rtcp-mux"
+              "require" RTCP multiplexing policy is set and no "a=rtcp-mux"
               line is present in the offer,  then the m=line MUST be
               marked as rejected by setting the port in the m= line to zero, as
               indicated in <xref target="RFC3264"></xref>, Section 6.
@@ -2301,7 +2301,7 @@ a=end-of-candidates
      <t> This section shows a typical example of a session between two
      browsers setting up an audio channel and a data channel. Trickle
      ICE is used in full trickle mode with a bundle policy of max-bundle,
-     an RTCP mux policy of max-mux, and a single TURN server.
+     an RTCP mux policy of require, and a single TURN server.
      Later, two video flows, one for the
      presenter and one for screen sharing, are added to the session.
      This example shows Alice's browser initiating the
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -689,7 +689,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           candidate pool size MUST be zero.
           </t>
 
-          <t>Lastly, the
+          <t>The
           application can specify its preferred policy regarding use of BUNDLE,
           the multiplexing mechanism defined in <xref
           target="I-D.ietf-mmusic-sdp-bundle-negotiation"></xref>.
@@ -723,18 +723,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               candidate gathering and maximize multiplexing, at the cost of less
               compatibility with legacy endpoints.
               </t>
-              <t></t>
-              <t hangText="max-bundle-and-rtcp-mux:">Similar to max-bundle,
-              but RTCP candidates are not gathered. This policy reduces the
-              candidates that must be gathered to the absolute minimum,
-              but will not be compatible with legacy endpoints that do not
-              support RTCP mux.
-              </t>
             </list>
             As it provides the best tradeoff between performance and
             compatibility with legacy endpoints, the default BUNDLE policy
             MUST be set to "balanced".
           </t>
+
+          <t>The
+          application can specify its preferred policy regarding use of RTP/RTCP
+          multiplexing <xref target="RFC5761"/> using one of the following policies:
+          <list style="hanging">
+            <t hangText="max-compat:">The browser will gather both RTP and
+            RTCP candidates but also will offer "a=rtcp-mux", thus allowing
+            for compatibility with either multiplexing or non-multiplexing endpoints.</t>
+
+            <t hangText="max-mux:">The browser will only gather RTP candidates
+            and when acting as answerer will reject any m= section that does not
+            provide an "a=rtcp-mux" attribute.</t>
+          </list>
+          </t>
+          
         </section>
         <section title="createOffer" anchor="sec.createoffer">
           <t>The createOffer method generates a blob of SDP that contains a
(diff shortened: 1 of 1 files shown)


fluffy:
looks on right path to me

draft-ietf-rtcweb-jsep.xml
```diff
@@ -723,18 +723,24 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           </t>
+
+          <t>The
+          application can specify its preferred policy regarding use of RTCP-MUX
+          <xref target="RFC5761"/> using one of the following policies:
+          <list style="hanging">
+            <t hangText="compat:">The browser will gather both RTP and
+            RTCP candidates but also will offer "a=rtcp-mux", thus allowing
+            for compatibility with either muxing or non-muxing endpoints.</t>
+            <t hangText="mux-only:">The browser will only gather RTCP candidates.
```
juberti: Assume you mean RTP here

draft-ietf-rtcweb-jsep.xml
```diff
@@ -723,18 +723,24 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
             As it provides the best tradeoff between performance and
             compatibility with legacy endpoints, the default BUNDLE policy
             MUST be set to "balanced".
           </t>
+
+          <t>The
+          application can specify its preferred policy regarding use of RTCP-MUX
+          <xref target="RFC5761"/> using one of the following policies:
+          <list style="hanging">
+            <t hangText="compat:">The browser will gather both RTP and
```
juberti: I don't love these names since they aren't analogous with the bundle policies, but I don't have any other suggestions yet. Will think about it.

draft-ietf-rtcweb-jsep.xml
ekr: yes

draft-ietf-rtcweb-jsep.xml
ekr: I changed to "max-compat" and "max-mux" which are more analogous but also more silly since this is a boolean.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1880,7 +1889,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               the "active" role is RECOMMENDED.</t>
 
               <t>If present in the offer, an "a=rtcp-mux" line, as specified
-              in <xref target="RFC5761"></xref>, Section 5.1.1.</t>
+              in <xref target="RFC5761"></xref>, Section 5.1.1. If the
+              "max-mux" RTCP multiplexing policy is set and no "a=rtcp-mux"
+              line is present in the offer,  then the m=line MUST be
+              marked as rejected by setting the port in the m= line to zero, as
+              indicated in <xref target="RFC3264"></xref>, Section 6.
+              </t>
```
juberti: This makes sense, as we discussed today, but this raises an interesting point, namely what should the behavior be for an endpoint marked as max-bundle or balanced, when it receives an offer without BUNDLE? If we extend this to its logical conclusion, any m= lines that the endpoint expected to be bundled would have to be rejected, and since balanced is the default, this would mean that two non-bundled video streams would by default be rejected.

The alternative would simply be to say that if the remote side doesn't support BUNDLE, you just try to carry on without bundling, for current and future m= lines. This is a bit different than rtcp-mux, in that it is global, and therefore you can't get into weird cases depending on who is offering. But then the symmetry with the rtcp-mux policy is lost.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -723,18 +723,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
             As it provides the best tradeoff between performance and
             compatibility with legacy endpoints, the default BUNDLE policy
             MUST be set to "balanced".
           </t>
+
+          <t>The
+          application can specify its preferred policy regarding use of RTP/RTCP
+          multiplexing <xref target="RFC5761"/> using one of the following policies:
+          <list style="hanging">
+            <t hangText="max-compat:">The browser will gather both RTP and
```
juberti: I see what you were trying to do here, but I agree it seems silly. After thinking about it a bit, I think it's a distinct animal from the bundle policy. How about "negotiate" and "require" as policies?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -723,18 +723,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            RTCP candidates but also will offer "a=rtcp-mux", thus allowing
+            for compatibility with either multiplexing or non-multiplexing endpoints.</t>
+
+            <t hangText="max-mux:">The browser will only gather RTP candidates
+            and when acting as answerer will reject any m= section that does not
+            provide an "a=rtcp-mux" attribute.</t>
```
juberti: Suggest pointing out specifically that this halves the number of candidates that need to be gathered.

draft-ietf-rtcweb-jsep.xml
ekr: SGTM.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1064,10 +1072,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             immediately; if increased, additional candidates are pre-gathered;
             if decreased, the now-superfluous candidates are discarded.
             </t>
-            <t>Any changes to the BUNDLE policy take effect immediately, i.e.
+            <t>Any changes to the BUNDLE or RTCP multiplexing policies take effect immediately, i.e.
             any future tracks added to the PeerConnection will have their
-            bundle-only state marked accordingly.
+            bundle-only and RTP/RTCP multiplexing state marked accordingly.
```
juberti: What does marking RTP/RTCP mux state mean exactly? This flag doesn't really control any signaling, it controls the RTCP gathering state and what happens when the answer is received.

draft-ietf-rtcweb-jsep.xml
ekr: Willdo.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1064,10 +1072,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             immediately; if increased, additional candidates are pre-gathered;
             if decreased, the now-superfluous candidates are discarded.
             </t>
-            <t>Any changes to the BUNDLE policy take effect immediately, i.e.
+            <t>Any changes to the BUNDLE or RTCP multiplexing policies take effect immediately, i.e.
```
juberti: FWIW I think frobbing these states post-construction is likely to lead to tears (both for implementors and app developers), so we should consider disallowing such changes altogether.

draft-ietf-rtcweb-jsep.xml
ekr: Well, what I thought it meant in this context was that if we added a new track, we would gather RTCP or not as indicated here. Does that not make sense?

draft-ietf-rtcweb-jsep.xml
ekr: I would be totally fine with that.

draft-ietf-rtcweb-jsep.xml
juberti: OK, I was thinking about in terms of what goes into SDP, and a=rtcp-mux always goes into SDP. I would be fine with this as "... and RTCP multiplexing state set accordingly."

draft-ietf-rtcweb-jsep.xml
juberti: Let's just do that then.

fluffy: +1 one not allowing this

draft-ietf-rtcweb-jsep.xml
fluffy: If you get an offer that does not support FOO, you should just generate an answers that is consistent with no FOO and not fail. The idea one would fail because of this sounds nuts to me. (and same with bundle.

fluffy:
My only issue with this is what happens if you get an offer that does not do mux, I think it should be accepted regardless of what this is set to. I think this should be purely about what is generated in offer. Other than that it looks good.

ekr:
We should discuss this on today's call. Justin and I spent a while on it last time and
one of the relevant arguments was consistency. Specifically, say you have two
endpoints, one of which is old (i.e., pre-WebRTC) and doesn't support mux at
all and the other of which is set to "require", i.e., only do mux.

With the current text, any attempt between these two to communicate will fail.

If you only apply these polices to the offer, however:

- If the old endpoint is the offerer, then it will not offer mux but the WebRTC
  endpoint will accept it and the call will succeed.
- If the new endpoint is the offerer, then it will only provide mux candidates
  and the call will fail in a fairly ugly fashion.

The point being that the current text allows you to fail fast.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1880,7 +1889,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               the "active" role is RECOMMENDED.</t>
 
               <t>If present in the offer, an "a=rtcp-mux" line, as specified
-              in <xref target="RFC5761"></xref>, Section 5.1.1.</t>
+              in <xref target="RFC5761"></xref>, Section 5.1.1. If the
+              "max-mux" RTCP multiplexing policy is set and no "a=rtcp-mux"
```
juberti: max-mux -> require

draft-ietf-rtcweb-jsep.xml
juberti: The RTCP case breaks in the case where the answerer wants to force RTCP mux, receives an offer from a non-muxer (which results in a non-mux m= section), and then later reoffers with a new m= section. Unless the answerer is somehow keeping track that the remote side might not want to do mux, it will send its reoffer with the new m= section being mux-only.

I now think that the BUNDLE case is distinct, there never is any ambiguity on either side. If the answerer doesn't support BUNDLE, you just won't use BUNDLE, and that's easy to deal with.
Anyway, TL;DR - I think this text LG as-is. It is a corner case.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2272,8 +2286,8 @@ a=end-of-candidates
 
      <t> This section shows a typical example of a session between two
      browsers setting up an audio channel and a data channel. Trickle
-     ICE is used in full trickle mode with a policy of max-bundle-and-rtcp-mux
-     and a single TURN server.
+     ICE is used in full trickle mode with a bundle policy of max-bundle,
+     an RTCP mux policy of max-mux, and a single TURN server.
```
juberti: max-mux->require

juberti:
lgtm

ekr:
Push -f due to rebase screwup.

Cullen and I discussed this PR and he is unconvinced that we should fail when
the offer does not include mux but the answerer has set "require". In the 
interest of landing the rest of this, I have added an issue and an open issue
section in the draft.

https://github.com/rtcweb-wg/jsep/issues/114

Cullen, Justin, if this is acceptable LMK and I will merge this PR in.

juberti:
This seems fine to me - good topic to discuss in Dallas, if we are still of two minds at that point.

ekr:
@fluffy can you confirm this matches your expectations?

fluffy:
LGTM and this open issue is not a huge deal one way or the other

PR #113, opened 2015-03-08
Start of setLD/RD

juberti:
@ekr @fluffy PTAL and let me know if you think this is on the right track. If so, I will flesh out in whatever time I have on Sunday.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2227,7 +2227,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         <t>Next, media sections are processed. For each media section,
         the following steps MUST be performed; if any parameters are out of
         bounds, or cannot be applied, processing MUST stop and an error MUST be
-        returned.</t>
+        returned.
           <list style="symbols">
             <t>TODO</t>
           </list>
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2047,13 +2047,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         The SDP is read,
         line-by-line, and converted to a data structure that contains the
         deserialized information. However, SDP allows many types of lines, not
-        all of which are relevant to JSEP applications many of them
-        have no useful function in JSEP contexts.
+        all of which are relevant to JSEP applications.
         For each line, the implementation will first ensure it is syntactically
-        correct according its defining ABNF [TODO: reference], and check
+        correct according its defining ABNF [TODO: reference], check
         that it conforms to <xref target="RFC4566"/> and <xref target="RFC3264"/>
-        semantics and then process it (discarding, checking and discarding, or
-        storing) as described below.
+        semantics, and then either parse and store or discard the provided value,
+        as described below.
         [TODO: ensure that every line is listed below.]
         If the line is not well-formed, or cannot be parsed as described, the parser MUST stop
         with an error and reject the session description. This ensures that
@@ -2068,11 +2067,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             and their ordering is not meaningful.</t>
           <t>For non-attribute (non-"a=") lines, their sequencing, syntax,
             and semantics, are checked, as mentioned above. The following
-            lines ([TODO]) are not meaningful in the JSEP context and
-            MAY be discarded once they have been checked. The remaining
-            lines are processed as follows:
+            lines are not meaningful in the JSEP context and
+            MAY be discarded once they have been checked.
+            <list>
+              <t>TODO</t>
+            </list>
           </t>
-          <t>
+          <t>The remaining lines are processed as follows:
             <list>
               <t>The "c=" line MUST be parsed and stored.</t>
               <t>[OPEN ISSUE: For example, because session-level bandwidth is
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2227,29 +2227,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         <t>Next, media sections are processed. For each media section,
         the following steps MUST be performed; if any parameters are out of
         bounds, or cannot be applied, processing MUST stop and an error MUST be
-        returned.
+        returned.</t>
           <list style="symbols">
-            <t>If the ICE ufrag and password values changed, trigger the ICE Agent
-              to start an ICE restart and begin gathering new candidates, as defined
-              in <xref target="RFC5245"/>, Section X.Y.</t>
-          </list>
-        </t>
-        <t>If the media section proto value indicates use of RTP:
-          <list style="symbols">
-            <t>If the RTCP mux parameter is set, discard any RTCP transport,
-             and begin muxing RTCP over the RTP transport, as specified in
-             <xref target="RFC5761"/>, Section X.Y. [if answer]
-            </t>
-            <t>Set RTP header extension receive state, ensure all are known</t>
-            <t>Set PT demux</t>
-            <t>Set receive codecs, ensure all are known</t>
-            <t>Set up RTX (verify apt)</t>
-            <t>Set local directional state</t>
-          </list>
-        </t>
-        <t>If the media section proto value indicates use of SCTP:
-          <list style="symbols">
-            <t>Set local SCTP pseudo-port</t>
+            <t>TODO</t>
           </list>
         </t>
         <t>Finally, if this description is of type "pranswer" or "answer", follow
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2169,7 +2169,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                 <xref target="RFC5285"/>, Section 5, and their values stored.</t>
               <t>Any "a=rtcp-fb" attributes MUST be parsed as specified in
                 <xref target="RFC4585"/>, Section 4.2., and their values stored.</t>
-              <t>Any "a=rtcp-mux" attribute MUST be parsed as specified in
+              <t>If present, a single "a=rtcp-mux" line MUST be parsed as specified in
                 <xref target="RFC5761"/>, Section 5.1.1, and its presence or
                 absence flagged and stored.</t>
               <t>TODO: a=rtcp-rsize, a=rtcp, a=msid, a=candidate, a=end-of-candidates</t>


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                <t>If the type is "pranswer" or "answer", the PeerConnection
+                      state MUST be either "have-remote-offer" or
+                      "have-local-pranswer".</t>
+              </list>
+            </t>
+            <t>If the type is not correct for the current state, processing MUST
+            stop and an error MUST be returned.</t>
+            <t>Next, the SessionDescription is parsed into a data structure,
+              as described in the
+              <xref target="sec.parsing-a-desc"/> section below.
```
ekr: Do you want to say something about what happens on parse failures?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+      </section>
 
-      <section title="Applying a Local Description" anchor="sec.applying-a-local-descript"></section>
+      <section title="Parsing a Session Description" anchor="sec.parsing-a-desc">
+        <t>When a SessionDescription of any type is supplied to
+        setLocal/RemoteDescription, the implementation must parse it and
+        reject it if it is invalid. The exact details of this
+        process are explained below.</t>
+        <t>The SDP contained in the session description object consists of a
+        sequence of text lines, each containing a key=value expression.
```
ekr: This seems like a good place to reference 4566. Also that says <type>=<value>

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        The SDP is read,
+        line-by-line, and converted to a data structure that contains the
+        deserialized information. However, SDP supports various fields that are not relevant to JSEP applications, and so these fields can be ignored.
```
ekr: s/supports/allows/

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        For each line, the implementation will first ensure it is syntactically
+        correct, and then either parse the data to store for later use,
+        or discard the line and move on. If the line is not well-formed, or
+        cannot be parsed as described, the specified the parser MUST stop with an error and reject the session decription. This ensures that implementations do not accidentally misinterpret ambiguous SDP.</t>
```
ekr: s/the specified//

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        reject it if it is invalid. The exact details of this
+        process are explained below.</t>
+        <t>The SDP contained in the session description object consists of a
+        sequence of text lines, each containing a key=value expression.
+        The SDP is read,
+        line-by-line, and converted to a data structure that contains the
+        deserialized information. However, SDP supports various fields that are not relevant to JSEP applications, and so these fields can be ignored.
+        For each line, the implementation will first ensure it is syntactically
+        correct, and then either parse the data to store for later use,
+        or discard the line and move on. If the line is not well-formed, or
```
ekr: To be clear, I think you're saying that lines which are to be discarded must still be syntactically valid, right?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        cannot be parsed as described, the specified the parser MUST stop with an error and reject the session decription. This ensures that implementations do not accidentally misinterpret ambiguous SDP.</t>
+
+        <section title="Session-Level Parsing">
+          <t>First, the session-level lines are checked and parsed. These
+            lines MUST occur in a specific order, and with a specific syntax, as defined in <xref target="RFC4566"/>, Section 5. </t>
+          <t>For non-attribute (a=) lines, their sequencing and syntax are checked,
+            as mentioned above, but all values MUST be discarded. For example,
+            because the use of ICE is mandatory, the c= line is superfluous and its value MUST be discarded; similarly, because session-level bandwidth is ambiguous when multiple media streams are present, a b= line at session
+            level is not useful and its value MUST be discarded.</t>
```
ekr: Are you going to provide a list below of which values must be discarded? If so, probably one example suffices.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              <t>If present, an a=ice-options is parsed as specified in
+                [TBD], and, if valid, the set of specified options is stored.</t>
+              <t>If present, any a=fingerprint lines are parsed as specified in
+                [TBD], and, if valid, the set of fingerprint and algorithm
+                values is stored.</t>
+              <t>If present, any a=setup attribute is parsed as specified in
+                [TBD], and, if valid, the value is stored.</t>
+              <t>TODO: msid-semantic and related values</t>
+            </list>
+            Once all the session-level lines have been parsed, processing
```
ekr: I wonder if this much detail is useful here. It seems like kind of a recipe for accidental inconsistencies with the existing RFCs. I understand the desire to have this spec be self-contained, but I think there's a tension with making it copy-by-value.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          <t>First, the session-level lines are checked and parsed. These
+            lines MUST occur in a specific order, and with a specific syntax, as defined in <xref target="RFC4566"/>, Section 5. </t>
+          <t>For non-attribute (a=) lines, their sequencing and syntax are checked,
+            as mentioned above, but all values MUST be discarded. For example,
+            because the use of ICE is mandatory, the c= line is superfluous and its value MUST be discarded; similarly, because session-level bandwidth is ambiguous when multiple media streams are present, a b= line at session
+            level is not useful and its value MUST be discarded.</t>
+          <t>Specific processing is applied for the following session-level
+            attribute (a=) lines:
+            <list style="symbols">
+              <t>Any a=group lines are parsed as specified in [TBD], and, if
```
ekr: The convention in this document seems to be to quote "a=<foo>"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                is stored.</t>
+              <t>If present, an a=ice-options is parsed as specified in
+                [TBD], and, if valid, the set of specified options is stored.</t>
+              <t>If present, any a=fingerprint lines are parsed as specified in
+                [TBD], and, if valid, the set of fingerprint and algorithm
+                values is stored.</t>
+              <t>If present, any a=setup attribute is parsed as specified in
+                [TBD], and, if valid, the value is stored.</t>
+              <t>TODO: msid-semantic and related values</t>
+            </list>
```
ekr: Is this an exhaustive list? What about a=identity?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        deserialized information. However, SDP supports various fields that are not relevant to JSEP applications, and so these fields can be ignored.
+        For each line, the implementation will first ensure it is syntactically
+        correct, and then either parse the data to store for later use,
+        or discard the line and move on. If the line is not well-formed, or
+        cannot be parsed as described, the specified the parser MUST stop with an error and reject the session decription. This ensures that implementations do not accidentally misinterpret ambiguous SDP.</t>
+
+        <section title="Session-Level Parsing">
+          <t>First, the session-level lines are checked and parsed. These
+            lines MUST occur in a specific order, and with a specific syntax, as defined in <xref target="RFC4566"/>, Section 5. </t>
+          <t>For non-attribute (a=) lines, their sequencing and syntax are checked,
```
ekr: Can't "a=" lines  occur in any order?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                </list>
+              </t>
+              <t>Handle SCTP attributes:
+                <list style="symbols">
+                  <t>Read sctpmap/port attributes</t>
+                </list>
+              </t>
+            </list>
+          </t>
+        </section>
```
ekr: This seems superficially plausible.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              </list>
+          </t>
+          <t>If this session description is of type "pranswer" or "answer", the
+          following additional checks are applied:
+            <list style="symbols">
+              <t>The number of m= sections MUST exactly match the number of m=
+              sections in the associated offer.</t>
+              <t>For each m= section, the protocol value MUST exactly match the
+              protocol value in the corresponding m= section in the associated
+              offer.</t>
```
ekr: Good!

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        </section>
+      </section>
 
-      <section title="Applying a Remote Description" anchor="sec.applying-a-remote-descrip"></section>
+      <section title="Applying a Local Description" anchor="sec.applying-a-local-desc">
+       <t>The following steps are performed at the media engine level to apply a local
+       description. For each m= section, do the following: [TODO - expand]
+         For all m= sections:
+         <list style="symbols">
+          <t>If ufrag/pwd changed, trigger ICE restart</t>
```
ekr: Since you can't produce your own values for this, you need to validate that these are consistent with what you produced.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,205 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+         </list>
+         For RTP m= sections:
+         <list style="symbols">
```
ekr: Some sort of consistency check needs to be done here. I.e., that only changes permitted in Section 6 were performed.

draft-ietf-rtcweb-jsep.xml
juberti: Yes, lines to be ignored (other than syntax checking) are to be ignored (i.e. no further processing as far as JSEP is concerned).

ekr:
This seems like it's on the right track. My biggest concern is whether we really want to require people to ignore values which we also prescribe values for. For instance, you say that the impl should discard the c= line, but S 5.2.1 is very prescriptive about the c= line contents. I recognize that valid (non-JSEP)  implementations might produce other values for the c= line but I wonder if we really want to ignore them.

draft-ietf-rtcweb-jsep.xml
juberti: I think we need to say what gets stuffed into the data structure which is then used in the 'apply' sections below. The createoffer/answer stuff is pretty specific, so I think we're already heading down this path a bit. But I held off on rewriting 4566 inside this spec.

draft-ietf-rtcweb-jsep.xml
juberti: no, still some TODOs. but it should be exhaustive. also missing a=extmap.

draft-ietf-rtcweb-jsep.xml
juberti: yes, but the letters themselves must be in a specific order. i.e. a only after c=, etc

draft-ietf-rtcweb-jsep.xml
juberti: agreed

juberti:
What would one do with the c= line? We make our c= lines in a certain way for the benefit of dumb remote endpoints. JSEP endpoints aren't so encumbered.

draft-ietf-rtcweb-jsep.xml
ekr: OK.

draft-ietf-rtcweb-jsep.xml
ekr: OK, well I'm willing to try going down this path, but if we get too bogged down in inconsistencies I may re-raise this point later

draft-ietf-rtcweb-jsep.xml
ekr: Agreed. It's just an editorial issue.

ekr:
I guess you would validate that it properly reflected one of the candidates (if any) in the m= section.

I'm torn here, because it seems like if we totally ignore these values we contribute to them just getting more and more bogus.

ekr:
With that said, Firefox seems to basically ignore these values.

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml
juberti: Rewrote this part a bit.

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml
juberti: Clarified this.

draft-ietf-rtcweb-jsep.xml
juberti: I was originally just going to have these as examples, but I think that specifically saying b=/c= etc are ignored will be necessary to have consistency. Haven't yet done that though.

draft-ietf-rtcweb-jsep.xml
juberti: Ack.

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml
juberti: Still TODO, will come back to this once I hit some of the bigger things

draft-ietf-rtcweb-jsep.xml
juberti: Clarified (or tried to)

draft-ietf-rtcweb-jsep.xml
juberti: Agreed. Will speak to that when I flesh this out. Added TODO for now; fortunately we have Section 6 which I can just reference specifically.

juberti:
It wouldn't be hard to verify c=; we would have to store the value, and then deal with it during the "Verifying Consistency" section. However this could be a slippery slope; should we validate that t= values are not garbage, etc?

ekr:
"It wouldn't be hard to verify c=; we would have to store the value, and then deal with it during the "Verifying Consistency" section. However this could be a slippery slope; should we validate that t= values are not garbage, etc?"

Fair enough. OTOH there's a hygiene argument about rejecting bogus stuff early so it
doesn't get into the ecosystem is not contaminated. Maybe we should say that these
lines *can* be checked for consistency? I don't feel strongly about this.

juberti:
I am strongly on the side of "be strict in what you accept, or else people will turn your ecosystem into a cesspool", but I have a hard time getting excited about extensively validating stuff that is going to be ignored anyway.

Saying that the application MAY do additional consistency checks seems like a reasonable balance; I'll see if I can craft some text around that.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,224 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        reject it if it is invalid. The exact details of this
+        process are explained below.</t>
+        <t>The SDP contained in the session description object consists of a
+        sequence of text lines, each containing a key-value expression,
+        as described in <xref target="RFC4566"/>, Section 5.
+        The SDP is read,
+        line-by-line, and converted to a data structure that contains the
+        deserialized information. However, SDP allows many types of lines, not
+        all of which are relevant to JSEP applications, and the ones that are
+        not relevant can be ignored.
```
fluffy: I'd remove the part about not relevant can be ignored here and instead just deal with it when we get to the specifics. When we get to the specifics, I would phrase it more as not JSEP was changing the definiton of SDP but more in the example of the k= lines is describes a SDP situation that is now allowed in the WebRTC profile of RTP security thus a browser will never generate and any SDP that contained it would not be a capability the browser had thus by it's definition it would be ignored.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,224 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        For each line, the implementation will first ensure it is syntactically
+        correct, and then either parse the data value to store for later use,
+        or ignore the value and move on. The implementation MAY store ignored
+        lines for diagnostic purposes or other reasons, but the data contained
+        within will not be used in any future JSEP processing. If the line is
```
fluffy: I don't really care but this gets into too much detail for my taste about how one must implement it - clearly one could implement things differently and it would still all work fine. Not asking for any change here but I would do a pass over this to see how much you can remove the ignore language. It is true that in many 3264 Offer/Answer cases, there are bits in the SDP that are basically ignored and that is fine. But if this comes across as one is doing a subset of 3264 offer/answer, then it's not so fine. ( And with the possibility of a very few small exceptions, I don't think you are suggesting that we are not doing 3264 )

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,224 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+
+        <section title="Session-Level Parsing">
+          <t>First, the session-level lines are checked and parsed. These
+            lines MUST occur in a specific order, and with a specific syntax, as
+            defined in <xref target="RFC4566"/>, Section 5. Note that while
+            the specific line types (e.g. "v=", "c=") MUST occur in the defined
+            order, lines of the same type (typically "a=") can occur in any order,
+            and their ordering is not meaningful.</t>
+          <t>For non-attribute (non-"a=") lines, their sequencing and syntax are checked,
+            as mentioned above, but all values MUST be discarded. For example,
```
fluffy: uh, saying they must be discarded is not OK. First I don't think they are, second, this is the extensibility mechanism of SDP and you can't change that here.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,224 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            because the use of ICE is mandatory, the "c=" line is superfluous and its
+            value MUST be discarded; similarly, because session-level bandwidth is
```
fluffy: I'd think you would be better to say this as the value of c line is not read when doing ICE but remember it needs to be still needs to be correctly generated - particularly in reoffers - as in that case indicates the address in use for SBCs.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,224 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            ambiguous when multiple media streams are present, a "b=" line at session
+            level is not useful and its value MUST be discarded.</t>
```
fluffy: uh not sure about b= here. I think we still need to be able to deal with it, it's just the more nuanced version at finer grai are more userfull and override it. That said I would need to go back and look at the specs - perhaps this has been deprecated.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,224 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                mandatory-to-use features enumerated in <xref target="sec.usage-requirements"/> MUST be present. These values MAY either be present at the
+                media level, or inherited from the session level.
+                  <list style="symbols">
+                    <t>ICE ufrag and password values</t>
+                    <t>DTLS fingerprint and setup values</t>
+                  </list>
+                </t>
+                <t>If a BUNDLE group is present, the referenced m= sections MUST
+                exist. If multiple BUNDLE groups exist, each referenced m=
+                section MUST only exist in a single BUNDLE group. [TODO: should this apply to all groups?]</t>
```
fluffy: This type of stuff should be in the bundle spec not here

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,224 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                the ufrag value is stored.</t>
+              <t>If present, an "a=ice-pwd" line is parsed as specified in
+                <xref target="RFC5245"/>, Section 15.4, and, if valid, the pwd
+                value is stored.</t>
+              <t>If present, an "a=ice-options" line is parsed as specified in
+                <xref target="RFC5245"/>, Section 15.5, and, if valid, the set of specified options is stored.</t>
+              <t>If present, any "a=fingerprint" lines are parsed as specified in
+                <xref target="RFC4572"/>, Section 5, and, if valid, the set of fingerprint and algorithm values is stored.</t>
+              <t>If present, an "a=setup" line is parsed as specified in
+                <xref target="RFC4145"/>, Section 4, and, if valid, the value
```
fluffy: Do we really want to say each of these or just say something more like "process ICE attributes as described in 5245"

draft-ietf-rtcweb-jsep.xml
juberti: yeah, this is probably overzealous. The point here is that this spec won't make use of any of the session-level non-a= lines. But I guess a new one could be added in the future that we might want to care about, so we shouldn't prevent that. Eric and I discussed this briefly and I think I have text that will work here.

draft-ietf-rtcweb-jsep.xml
juberti: OK. In the createOffer section, I do explain that this must be very carefully generated. I could add some text to avoid confusion here if you think that would help.

draft-ietf-rtcweb-jsep.xml
juberti: One could imagine setting it here at session level for an audio-only call. But overall, do we want to encourage that? I think JSEP should be considered "SDP: The Good Parts" and we should try to deprecate known-problematic ways of doing things.

draft-ietf-rtcweb-jsep.xml
fluffy: yah, I think some of this is just subtle flavor of how to phrase it. The varios SDP RFC need to be the actual defining spec or else we will just get it wrong here and be too long but agree with browser would currently never read some things

draft-ietf-rtcweb-jsep.xml
ekr: I believe that Firefox does ignore the c= line, but it appears that 5245 thinks you ought not to:

https://tools.ietf.org/html/rfc5245#section-9.1.1.1

"  These rules imply that setting the IP address in the c line to
   0.0.0.0 will cause an ICE restart.  Consequently, ICE implementations
   MUST NOT utilize this mechanism for call hold, and instead MUST use
   a=inactive and a=sendonly as described in [RFC3264]."


And:
https://tools.ietf.org/html/rfc5245#section-9.2.2.3

  The agent forms a candidate pair for each component of the media
   stream by:

   o  Setting the remote candidate equal to the offerer's default
      destination for that component (e.g., the contents of the m and c
      lines for RTP, and the a=rtcp attribute for RTCP)

draft-ietf-rtcweb-jsep.xml
juberti: I couldn't find a good reference in BUNDLE but it is covered in 5888, I can just refer to that section instead.

draft-ietf-rtcweb-jsep.xml
juberti: We could probably be terser here and say "read the ICE ufrag, pwd, options, ice-lite" as defined in 5245, Sectiond 15.3-15.5. I do think we need to be prescriptive here since we are going to later use these parsed values in the apply section, and it needs to be clear where they came from.

draft-ietf-rtcweb-jsep.xml
fluffy: As much as I would love to see SDP get a major clean up, that would need to happen over in mmusic not webrtc. That said, I'g got no objects with "implementers hints" here but for the most part I don't think we can change SDP. We do have something that like any SDP that is not ICE and DTLS/SRTP is invalid but that more the usage than a change of SDP.  But all that said, I don't think I am disagreeing much with what is being said here, just think we need to say some of it a bit differenetly.

draft-ietf-rtcweb-jsep.xml
juberti: For 9.1.1.1, the ufrag/pwd also need to change in this case, so I would still argue reading c= is not necessary. For 9.2.2.3 I would need to go do some more thinking, but I suspect that this only is important if you generate the updated offer of dubious value.

That said, perhaps it would just be simplest to say that we do read the c= and store it, and then we can go think through what exactly we want to do with it later.

juberti:
I am currently working on an edit that tries to address the points raised here, look for it in ~30 min.

juberti:
OK, I took another shot at this. Now, I sort lines into things that JSEP specifically cares about and MUST be acted on, and things that JSEP doesn't say anything about, and implementations MAY act on them or ignore them.

juberti:
I finished fleshing out the SDP parsing stuff. It definitely could be tightened up with another editorial pass, but I think it mostly defines what needs to happen at the parse stage, and is worth keeping; we can polish it later.

The apply-local-description stuff is still quite primitive and we might want to pull it for this rev.

juberti:
I will be in the office by 8:30 PT tomorrow - ping me and we can figure out how to land this from the current state.

fluffy:
working my way thought the latest ...

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,265 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        reject it if it is invalid. The exact details of this
+        process are explained below.</t>
+        <t>The SDP contained in the session description object consists of a
+        sequence of text lines, each containing a key-value expression,
+        as described in <xref target="RFC4566"/>, Section 5.
+        The SDP is read,
+        line-by-line, and converted to a data structure that contains the
+        deserialized information. However, SDP allows many types of lines, not
+        all of which are relevant to JSEP applications, and the ones that are
+        not specifically referenced below MAY be ignored. Implementations MAY
```
fluffy: I'm not OK with "and the ones that are not specifically referenced below MAY be ignored. Implementations MAY" because we have zero chance of getting that right here and it is not future upgradable. If a new thing is added to SDP to support a feature like simulcast, is should only need a change in the relevant SDP spec not here. Really what we need to call our in this section is which SDP extentions RFC / drafts the browser needs to support.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,265 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        also act on such lines, in accordance with the specification that
+        defines them, instead.
+        For each line, the implementation will first ensure it is syntactically
```
fluffy: syntactically correct is a bit vague here - keep in mind the browser will receive SDP with extensions it does not understand and that is OK and should not cause an error. However if the SDP does not meat the requirements of the SDP ABNF, then yes an error is appropriate. I think the best way to fix this would just be very crisp about the syntactically correct meaning.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,265 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        correct, and then either parse the data value to store for later use,
+        or ignore the value and move on. If the line is
+        not well-formed, or cannot be parsed as described, the parser MUST stop
```
fluffy: I seem to recall the error was going to constrained something about which line of the SDP had failed and why.

draft-ietf-rtcweb-jsep.xml
juberti: I don't think that can work. We spell out exactly what has to happen in createOffer, I think we have to do the same here. Yes, new things will be added in the future, and we need to at a minimum not choke on such things, but for the things we know about I don't think we can just say "consult the relevant RFCs" since there is so much ambiguity already.

ekr: I'm not sure I understand the objection. Isn't MAY be ignored just a basic statement of SDP interop? To take the example of simulcast, imagine we didn't get around to defining it before we finished JSEP. It would then be the case that JSEP implementations were permitted to ignore it. If we want JSEP implementations to process pieces of SDP they need to be called out in the document or be part of the now-current SDP processing rules.

draft-ietf-rtcweb-jsep.xml
ekr: I'm not following this comment.

draft-ietf-rtcweb-jsep.xml
juberti: Have a suggestion for text here? Perhaps

"For each line, if the implementation understands it, the implementation will first ensure it is syntactically valid according to the ABNF that defines it, and then ..."

ekr: Without trying to write text, here is the rule that I think should be applied:

- For SDP lines which you don't know about, e.g., a=monkey, they just need to follow the basic SDP rules.
- For SDP lines which you do know about, e.g., e=candidate, they need to match the syntax for that specific line.

draft-ietf-rtcweb-jsep.xml
juberti: I agree with you, but the details of the error seem to be a W3C document thing and not a JSEP thing. For here, I think just saying "stop with an error" is sufficient.

draft-ietf-rtcweb-jsep.xml
ekr: Justin's proposed text above seems fine as long as we also have a statement about following the general bnf of the form:       <type>=<value>

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,265 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+
+        <section title="Session-Level Parsing" anchor="sec.session-level-parse">
+          <t>First, the session-level lines are checked and parsed. These
+            lines MUST occur in a specific order, and with a specific syntax, as
+            defined in <xref target="RFC4566"/>, Section 5. Note that while
+            the specific line types (e.g. "v=", "c=") MUST occur in the defined
+            order, lines of the same type (typically "a=") can occur in any order,
+            and their ordering is not meaningful.</t>
+          <t>For non-attribute (non-"a=") lines, their sequencing and syntax are checked,
+            as mentioned above, but the only line whose value MUST be parsed and
```
fluffy: I don't think that only c= line  is what SDP says. I'm fine with saying as a guid to implementers that currently the information in the i= and k= are not used by the current WebRTC algorithm. I suspect this comes to roughly the same thing as what you are trying to say here.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,265 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            stored is the "c=" line. Note that because session-level bandwidth is
+            ambiguous when multiple media streams are present, a "b=" line at session
+            level is not useful and its value SHOULD be ignored.</t>
```
fluffy: I looked and did not see this in the notes - instead it looked like the discutions was we should support the finer grain mechanisms. I don't really guy the b= at session level is not useful.

draft-ietf-rtcweb-jsep.xml
fluffy: I'm fine with them as they are

draft-ietf-rtcweb-jsep.xml
juberti: I am fine to mark b= at session level as an open issue and discuss in Dallas.

juberti:
Based on the comments right now, I am planning on making the following changes:
- Specifically mention SDP general rules and specific ABNF in "syntactically correct" section
- mark session-level b= as open issue
- clean up "if valid" text
- clarify a couple places as to whether multiple values are OK (e.g. rtcp-mux)

LMK if you disagree or have additional things you want to see.

ekr:
I suggest that you also put in some sort of general "This is a draft" disclaimer. Other than that, I'm fine with proceeding as you propose.

I'll be around for another 30-60 minutes and then out for lunch. Back around 2-2:30ish.

draft-ietf-rtcweb-jsep.xml
fluffy: Yep - what both of you said sounds like good path to me

draft-ietf-rtcweb-jsep.xml
fluffy: Good point - I'm fine with that

draft-ietf-rtcweb-jsep.xml
fluffy: Just from my point of view, I'm happy to discuss this in Dallas but as a general rule of thumb for how we approach writing this, my view is we already agreed we were doing 3264 so I'm happy to deal with exception to that but I view this is likely in that category.

fluffy:
I'm around next two hours then in meeting till late. To me the most important thing is to make this clear we are not picking a few parts of SDP to implement but instead we are doing SDP. So I am nok OK with things like "ignore everything but what is mentioned here". I am OK with the way we use SDP here means that X,Y, are never used. I view what I am saying here matches WG consensus

ekr:
This doesn't seem philosophically unreasonable, but at this point I suggest we focus on what's actionable. Are there specific things that are still in this PR that you object to? If so, let's call
them out and fix them and merge it.

fluffy:
As a check list of things I think we need to mention ... somewhere in this text I think we need to say browser need to support 

RFC2833
RFC3264  
RFC4566 
ietf-mmusic-msid
ietf-mmusic-sdp-bundle-negotiation
RFC5761
RFC3605
ietf-payload-rtp-opus
RFC3551
RFC4145
RFC5245
RFC5576
RFC5506
RFC6465
RFC6464
ietf-mmusic-trickle-ice
ietf-mmusic-sctp-sdp
ietf-rtcweb-data-channel
ietf-payload-vp8
ietf-mmusic-sdp-simulcast
RFC7022

I've missed a few there (like 264) but I think that is bunch of them

juberti:
I expect to talk about each of these in detail in the parse/apply sections, but the laundry list already exists in https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-08#section-5.1.1

juberti:
Cullen, regarding "I am nok OK with things like "ignore everything but what is mentioned here". I am OK with the way we use SDP here means that X,Y, are never used":

The current text says that this spec defines how JSEP deals with the specific lines X,Y,Z, but implementations MAY process other lines in accordance with their definitions. Does that work for you?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,13 +1977,265 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            attribute lines MUST be processed:
+            <list style="symbols">
+              <t>The "m=" fmt value MUST be parsed as specified in
+                <xref target="I-D.ietf-mmusic-sctp-sdp"/>, Section 4.3, and
+                the application protocol value stored.</t>
+              <t>An "a=sctp-port" attribute MUST be present, and it MUST be
+                parsed as specified in <xref target="I-D.ietf-mmusic-sctp-sdp"/>,
+                Section 5.2, and the value stored.</t>
+            </list>
+          </t>
```
fluffy: I think here we should put a TODO to include text to deal with the SDP extensions defined in  

RFC2833
RFC3605
RFC5506
RFC6465
RFC6464
ietf-mmusic-msid
ietf-mmusic-sdp-bundle-negotiation
ietf-payload-rtp-opus
ietf-mmusic-trickle-ice
ietf-payload-vp8

I don't care that we add the actually references for those if that turns out to be a PITA. I just seems we need a reminder to go deal with them as well.

juberti: I am fine with that (aside; I think -opus and 2833 are already covered (via a=fmtp);  -bundle is already covered via a=group; 3605 isn't directly used by us (a=candidate used instead), and 6464/6465 are handled by extmap)

So that just leaves 5506, msid, trickle-ice, and vp8.

fluffy: I forgot about the https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-08#section-5.1.1 text which has 5506 and you have a todo for msid. I don't think extmap covers 6464/6465  - they use extmap but I don't think extmap requires them. Seems like they should be mentioned somewhere. The 3605 seems like it is in sort of same category as c line but not sure what needed with it.

juberti:
OK. Working on editing this all now - expect an update circa 3:00 PM PT

draft-ietf-rtcweb-jsep.xml
ekr: v= is a good example here. Is it your position that if v=20, that we should reject the SDP?

juberti: It's a gray area, but I would say yes, since it can be done trivially as part of the parse.

ekr: Perhaps
"their sequencing, syntax, and semantic conformance with RFC3264 and RFC4566 are checked" would get us what Cullen is after. However, I suspect based on what you just said about trivially, that's not going to work for you.

Another example would be validating that o= incremented by 0 or 1.

juberti: I could live with that.

juberti:
New version has been pushed that addresses (I believe) all points discussed today.

fluffy:
reading

juberti:
Overall I think this has uncovered a bunch of stuff that I think we need to discuss in Dallas, so I'd like to find a way to land this, with whatever disclaimers are appropriate.

I plan to remove the apply-local stuff before submitting, will work on removing it now.

juberti:
apply-local stuff gutted

fluffy:
talking to EKR and trying to resolve something

ekr:
Justin, Cullen and I talked and his principal concern is defaulting to ignoring stuff not listed.
I think the way to fix this is to just list all (15 or so) non a= lines and their disposition. i didn't have time here, but PR #118  puts in the framework and then we should be able to resolve these quickly, leaving the controversial ones for IETF. wdyt?

See

PR #117, opened 2015-03-09
Set xd3 comments2

ekr:
Justin,

I got with Cullen and his objection seems primarily to be defaulting to "everything not documented is ignored". I think we can resolve this just by exhaustively listing the treatment for the 15 or so different kinds of SDP lines there are and whether they are ignored or not. I didn't have time to do that, but this PR puts in the framework for us to do that and I think we could nail down this list before and during IETF and move on.

wdyt?


ekr:
Oops.

PR #120, opened 2015-03-21
Clarify bundle policy behavior. Fixed #119

### LINKED ISSUE #119: Confusion about how many BUNDLE groups there are
5.2.1 seems clear that there is one:
Once all m= sections have been generated, a session-level "a=group" attribute MUST be added as specified in [RFC5888]. This attribute MUST have semantics "BUNDLE", and MUST include the mid identifiers of each m= section. The effect of this is that the browser offers all m= sections as one BUNDLE group. However, whether the m= sections are bundle-only or not depends on the BUNDLE policy.

4.1.1 seems clear that "balanced" means >1
The application will BUNDLE all media streams of the same type together. That is, if there are multiple audio and multiple video MediaStreamTracks attached to a PeerConnection, all but the first audio and video tracks will be marked as bundle-only, and candidates will only be gathered for N media streams, where N is the number of distinct media types. When talking to a non-BUNDLE-aware endpoint, only the non-bundle-only streams will be negotiated. This policy balances desire to multiplex with the need to ensure basic audio and video still works in legacy cases. ***Data channels will be in a separate bundle group.***

Justin and EKR believe the number should always be 1. Need to fix the text to make this clear.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -697,8 +697,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           control how aggressively it will try to BUNDLE media streams together.
           The set of available policies is as follows:
           <list style="hanging">
-              <t hangText="balanced:">The application will BUNDLE all media streams of
-              the same type together. That is, if there are multiple audio and
+              <t hangText="balanced:">The application will bundle all media streams into
+              a single bundle group. All but the first media stream of each
+              type will be marked as bundle-only. That is, if there are multiple audio and
```
martinthomson: This first sentence doesn't match the second.  The first says all media is bungled.  The second says that all media *of the same type* is bungled.

martinthomson: Oh, I see what you are doing here.  That's a little confusing.  You are offering to unbundle different types, but allowing them to be ALL bundled.

ekr: Yes. Do you have a preferred wording?

martinthomson: ---
The offer includes a single bundle group across all media sections.  The first media section of each type (audio, video, or application) MUST(?) contain transport parameters, which will allow an answerer to unbundle that section.  The second and any subsequent media section of each type will be marked bundle-only. 

Consequently, candidates are gathered for at most three media sections.  If there are multiple streams of a video type, the only option given to an answerer is to accept the bundling of those streams, or reject them entirely.

When negotiating with a non-BUNDLE aware [...]
---
I note that if there is an offer with two audio and two video streams, unbundling the first video will leave the answerer with a choice of bundling the second video with the audio, or rejecting it.  That's shitty, but I guess that's the design we have to deal with

draft-ietf-rtcweb-jsep.xml
```diff
@@ -706,8 +707,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               talking to a non-BUNDLE-aware endpoint, only the non-bundle-only
               streams will be negotiated. This policy balances desire to
               multiplex with the need to ensure basic audio and video still
-              works in legacy cases. Data channels will be in a separate bundle
-              group.
+              works in legacy cases.
```
martinthomson: "work", singular

martinthomson: Or "can be negotiated in"

fluffy:
LGTM

juberti:
Holding my review until Martin's comments are addressed

draft-ietf-rtcweb-jsep.xml
ekr: I took some of this but left off some of the explanation.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -697,17 +697,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-              multiple video MediaStreamTracks attached to a PeerConnection, all
-              but the first audio and video tracks will be marked as
-              bundle-only, and candidates will only be gathered for N media
-              streams, where N is the number of distinct media types. When
-              talking to a non-BUNDLE-aware endpoint, only the non-bundle-only
-              streams will be negotiated. This policy balances desire to
-              multiplex with the need to ensure basic audio and video still
-              works in legacy cases. Data channels will be in a separate bundle
-              group.
+              <t hangText="balanced:">The offer will include a single bundle
```
juberti: Let's either capitalize BUNDLE everywhere, or nowhere. (Q: since BUNDLE isn't an acronym, why do we capitalize it?)

draft-ietf-rtcweb-jsep.xml
```diff
@@ -697,17 +697,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              group across all media sections. The first media section of each type
```
juberti: Since the handling of BUNDLE groups is common across all these policies, I recommend putting this first statement into the preceding paragraph.

juberti:
lgtm with comments above

Issue #4, opened 2014-05-08
Generation of CNAMEs for MSTs


ekr:
Section 5.2.1:
      in accordance with [RFC7022].  [OPEN ISSUE:  How are CNAMEs
      specified for MSTs?  Are they randomly generated for each
      MediaStream?  If so, can two MediaStreams be synced?]

juberti:
What are next steps here?

fluffy:
Propose browser generates one CNAME (or more if multiple clock sources). Later API may add ways to set the CNAME from JS.

martinthomson:
That would be consistent with the RTP usage conclusions.

PR #127, opened 2015-04-22
Specify how many CNAME get generated

fluffy:
This fixes https://github.com/rtcweb-wg/jsep/issues/4

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1443,11 +1443,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               attribute, as specified in Section 6.1, indicating the CNAME for
               the source. The CNAME must be generated in accordance with
               <xref target="RFC7022"></xref>.
-              Each PeerConnection MUST use a different CNAME. All of the
-              MediaStreamTracks SHOULD have the same CNAME unless they have
-              different clock sources in which case the MediaStreamTracks with
-              different clocks sources MUST have different
-              CNAMEs. Future specification may define an Javascript API to set
+              Any MediaStreamTracks which have different clock sources MUST
+              have different CNAMEs. All MediaStreamTracks which have the same
+              clock source SHOULD have the same CNAME.
+              Future specification may define an Javascript API to set
               or copy the CNAME but that is not required in the API.</t>
 
               <t>If RTX is supported for this media type, another "a=ssrc"
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -405,6 +405,15 @@ setRemote(ANSWER) |                                   |
         APIs to allow the application to control various session parameters,
         which will provide the necessary information to the browser about what
         sort of SessionDescription to produce.</t>
+
+        <section title="CNAME Semantics" anchor="sec.cname">
+          
+          <t> Any MediaStreamTracks which have different clock sources MUST have
+          different CNAMEs. All MediaStreamTracks which have the same clock
+          source SHOULD have the same CNAME. </t>
+          
+        </section>
+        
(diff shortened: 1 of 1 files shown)


fluffy:
If this looks reasonable, we probably need some similar text in the answer.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1442,10 +1442,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-              specified for MSTs? Are they randomly generated for each
-              MediaStream? If so, can two MediaStreams be synced?
-              See: https://github.com/rtcweb-wg/jsep/issues/4]</t>
+              <xref target="RFC7022"></xref>.
+              Each PeerConnection MUST use a different CNAME. All of the
+              MediaStreamTracks SHOULD have the same CNAME unless they have
+              different clock sources in which case the MediaStreamTracks with
+              different clocks sources MUST have different
+              CNAMEs. Future specification may define an Javascript API to set
+              or copy the CNAME but that is not required in the API.</t>
```
martinthomson: I would not bother with forward-looking statements.  We all know how well that works in practice.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1442,10 +1442,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-              <xref target="RFC7022"></xref>. [OPEN ISSUE: How are CNAMEs
-              specified for MSTs? Are they randomly generated for each
-              MediaStream? If so, can two MediaStreams be synced?
-              See: https://github.com/rtcweb-wg/jsep/issues/4]</t>
+              <xref target="RFC7022"></xref>.
+              Each PeerConnection MUST use a different CNAME. All of the
+              MediaStreamTracks SHOULD have the same CNAME unless they have
+              different clock sources in which case the MediaStreamTracks with
+              different clocks sources MUST have different
+              CNAMEs. Future specification may define an Javascript API to set
```
ekr: Suggested rewrite:
"Any MediaStreamTracks which have different clock sources MUST have different CNAMEs. All MediaStreamTracks which have the same clock source SHOULD have the same CNAME."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1442,10 +1442,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               used for sending media, along with the mandatory "cname" source
               attribute, as specified in Section 6.1, indicating the CNAME for
               the source. The CNAME must be generated in accordance with
-              <xref target="RFC7022"></xref>. [OPEN ISSUE: How are CNAMEs
-              specified for MSTs? Are they randomly generated for each
-              MediaStream? If so, can two MediaStreams be synced?
-              See: https://github.com/rtcweb-wg/jsep/issues/4]</t>
+              <xref target="RFC7022"></xref>.
+              Each PeerConnection MUST use a different CNAME. All of the
```
juberti: This guidance is buried pretty deeply here in this section that tries to explain how to generate an offer. Especially if we are going to need similar text in the answer, it might make sense to pull this guidance on CNAMEs out to its own 3.x subsection.

Then this text can just say to use the appropriate CNAME as indicated in that subsection.

draft-ietf-rtcweb-jsep.xml
juberti: Do you mean MSTs with different clock sources on the same PC? Because clearly MSTs on different PCs should have distinct CNAMEs. As indicated above, I think this is complex enough to warrant its own section.

fluffy:
I think this got everyones comments - give it a quick read.

fluffy:
Replaced by PR https://github.com/rtcweb-wg/jsep/pull/135

Issue #130, opened 2015-04-24
Should "media description" not be "session description"?


stefhak:
JSEP on two occasions says "SDP media description" where it seems "SDP session description" or similar would be more appropriate. "media description" seem to reference m-lines rather than the full SDP in other documents (4566, 3264).

fluffy:
agree - see if that pull request fixes it

stefhak:
to my understanding it does fix it!

PR #131, opened 2015-04-24
Fix typo on SDP name

fluffy:


fluffy:
Fixes #130

stefhak:
LGTM!

draft-ietf-rtcweb-jsep.xml
```diff
@@ -225,7 +225,7 @@
     <section title="Semantics and Syntax" anchor="sec.semantics-and-syntax">
       <section title="Signaling Model" anchor="sec.signaling-model">
         <t>JSEP does not specify a particular signaling model or state
-        machine, other than the generic need to exchange SDP media
+        machine, other than the generic need to exchange SDP session
```
juberti: SDP and "session description" seem redundant. I would remove "SDP" and just say "session description" here and below.

stefhak: That would work fine as far as I'm concerned (I was after changing "media description" to "session description" - does not really matter if there is "SDP" before or not).

ekr: I agree with Justin here.

PR #136, opened 2015-05-10
inital cut at adding fast start answer example

### LINKED ISSUE #15: Provide example of how to do an inactive stream
S 4.1.4.1 reads:

"   Most web applications will not need to create answers using the
   "pranswer" type.  While it is good practice to send an immediate
   response to an "offer", in order to warm up the session transport and
   prevent media clipping, the preferred handling for a web application
   would be to create and send an "inactive" final answer immediately
   after receiving the offer.  Later, when the called user actually
   accepts the call, the application can create a new "sendrecv" offer
   to update the previous offer/answer pair and start the media flow.
   While this could also be done with an inactive "pranswer", followed
   by a sendrecv "answer", the initial "pranswer" leaves the offer-
   answer exchange open, which means the caller cannot send an updated
   offer during this time.
"

We need an example for this.

fluffy:
Fixes #15

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3848,17 +3848,134 @@ a=end-of-candidates
        <figure>
          <artwork><![CDATA[
 v=0
-o=- 7729291447651054566 1 IN IP4 0.0.0.0
+o=- 7729291447651054566 2 IN IP4 0.0.0.0
 s=-
 t=0 0
 a=msid-semantic:WMS
 a=group:BUNDLE a1 d1 v1 v2
-m=audio 64532 UDP/TLS/RTP/SAVPF 96 0 8 97 98
-c=IN IP4 55.66.77.88
-a=rtcp:64532 IN IP4 55.66.77.88
+m=audio 52546 UDP/TLS/RTP/SAVPF 96 0 8 97 98 
+c=IN IP4 11.22.33.44 
+a=rtcp:52546 IN IP4 11.22.33.44
 a=mid:a1
-a=msid:QI39StLS8W7ZbQl1sJsWUXkr3Zf12fJUvzQ1
+a=msid:57017fee-b6c1-4162-929c-a25110252400 
+       e83006c5-a0ff-4e0a-9ed9-d3e6747be7d9 
+a=sendrecv
+a=rtpmap:96 opus/48000/2
+a=rtpmap:0 PCMU/8000
+a=rtpmap:8 PCMA/8000
+a=rtpmap:97 telephone-event/8000
+a=rtpmap:98 telephone-event/48000
+a=maxptime:120
+a=ice-ufrag:ATEn1v9DoTMB9J4r
+a=ice-pwd:AtSK0WpNtpUjkY4+86js7ZQl
+a=ice-options:trickle
+a=fingerprint:sha-256
+              19:E2:1C:3B:4B:9F:81:E6:B8:5C:F4:A5:A8:D8:73:04
+             :BB:05:2F:70:9F:04:A9:0E:05:E9:26:33:E8:70:88:A2
+a=setup:actpass
+a=rtcp-mux
+a=rtcp-rsize
+a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
+a=extmap:2 urn:ietf:params:rtp-hdrext:sdes:mid
+a=candidate:109270923 1 udp 2122194687 192.168.1.2 51556 typ host 
+a=candidate:4036177503 1 udp 1685987071 11.22.33.44 52546 typ srflx 
+            raddr 192.168.1.2 rport 51556 
+a=candidate:3671762466 1 udp 41819903 22.33.44.55 61405 typ relay 
+            raddr 11.22.33.44 rport 52546 
+a=end-of-candidates
+m=application 52546 UDP/DTLS/SCTP webrtc-datachannel
+c=IN IP4 11.22.33.44
+a=mid:d1
+a=fmtp:webrtc-datachannel max-message-size=65536
+a=sctp-port 5000
+a=ice-ufrag:ATEn1v9DoTMB9J4r
+a=ice-pwd:AtSK0WpNtpUjkY4+86js7ZQl
+a=ice-options:trickle
+a=fingerprint:sha-256 19:E2:1C:3B:4B:9F:81:E6:B8:5C:F4:A5:A8:D8:73:04
+                     :BB:05:2F:70:9F:04:A9:0E:05:E9:26:33:E8:70:88:A2
+a=setup:passive
+a=candidate:109270923 1 udp 2122194687 192.168.1.2 51556 typ host
+a=candidate:4036177503 1 udp 1685987071 11.22.33.44 52546 typ srflx
+            raddr 192.168.1.2 rport 51556
+a=candidate:3671762466 1 udp 41819903 22.33.44.55 61405 typ relay
+            raddr 11.22.33.44 rport 52546
+            a=end-of-candidates
+m=video 52546 UDP/TLS/RTP/SAVPF 100 101
+c=IN IP4 11.22.33.44
+a=rtcp:52546 IN IP4 11.22.33.44
+a=mid:v1
+a=sendrecv
+a=rtpmap:100 VP8/90000
+a=rtpmap:101 rtx/90000
+a=fmtp:101 apt=100
+a=ice-ufrag:ATEn1v9DoTMB9J4r
+a=ice-pwd:AtSK0WpNtpUjkY4+86js7ZQl
+a=ice-options:trickle
+a=fingerprint:sha-256
+              19:E2:1C:3B:4B:9F:81:E6:B8:5C:F4:A5:A8:D8:73:04
+             :BB:05:2F:70:9F:04:A9:0E:05:E9:26:33:E8:70:88:A2
+a=setup:passive
+a=rtcp-mux
+a=rtcp-rsize
+a=extmap:2 urn:ietf:params:rtp-hdrext:sdes:mid
+a=rtcp-fb:100 ccm fir
+a=rtcp-fb:100 nack
+a=rtcp-fb:100 nack pli
+a=candidate:109270923 1 udp 2122194687 192.168.1.2 51556 typ host
+a=candidate:4036177503 1 udp 1685987071 11.22.33.44 52546 typ srflx
+          raddr 192.168.1.2 rport 51556
+a=candidate:3671762466 1 udp 41819903 22.33.44.55 61405 typ relay
+          raddr 11.22.33.44 rport 52546
+          a=end-of-candidates
+m=video 52546 UDP/TLS/RTP/SAVPF 100 101
+c=IN IP4 11.22.33.44
+a=rtcp:52546 IN IP4 11.22.33.44
+a=mid:v2
+a=sendrecv
+a=rtpmap:100 VP8/90000
+a=rtpmap:101 rtx/90000
+a=fmtp:101 apt=100
+a=ice-ufrag:ATEn1v9DoTMB9J4r
+a=ice-pwd:AtSK0WpNtpUjkY4+86js7ZQl
+a=ice-options:trickle
+a=fingerprint:sha-256
+              19:E2:1C:3B:4B:9F:81:E6:B8:5C:F4:A5:A8:D8:73:04
+             :BB:05:2F:70:9F:04:A9:0E:05:E9:26:33:E8:70:88:A2
+a=setup:passive
+a=rtcp-mux
+a=rtcp-rsize
+a=extmap:2 urn:ietf:params:rtp-hdrext:sdes:mid
+a=rtcp-fb:100 ccm fir
+a=rtcp-fb:100 nack
+a=rtcp-fb:100 nack pli
+a=candidate:109270923 1 udp 2122194687 192.168.1.2 51556 typ host
+a=candidate:4036177503 1 udp 1685987071 11.22.33.44 52546 typ srflx
+            raddr 192.168.1.2 rport 51556
+a=candidate:3671762466 1 udp 41819903 22.33.44.55 61405 typ relay
+            raddr 11.22.33.44 rport 52546
+a=end-of-candidates
+]]></artwork>
+       </figure>
+       </t>
+         
+     <t>
+       The SDP for |answer-C4| which is a answer looks like:
+     </t>
+       <t>
+       <figure>
+         <artwork><![CDATA[
+v=0
+o=- 4962303333179871723 2 IN IP4 0.0.0.0
+s=-
+t=0 0
+a=msid-semantic:WMS
+a=group:BUNDLE a1 d1 v1 v2
+m=audio 64532 UDP/TLS/RTP/SAVPF 96 0 8 97 98 
+c=IN IP4 55.66.77.88 
+a=rtcp:64532 IN IP4 55.66.77.88 
+a=msid:QI39StLS8W7ZbQl1sJsWUXkr3Zf12fJUvzQ1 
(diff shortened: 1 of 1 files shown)


fluffy:
Prob. needs some work but initial pass at this

juberti:
#15 talks about adding an example where the initial answer is a=inactive, as opposed to a pranswer. I think the point of this sample was to show how one could send an inactive answer (followed by a later sendrecv offer) rather than a pranswer, as JSEP discourages use of pranswer.

I think your example could still fit this scenario with modest changes.

fluffy:
Just to be clear, you would have to send an new offer to change the inactive if it had already be sent in a answer. You thinking an A sends offer to B and B sends answer with inactive. Then B sends offer changing to active and A sends answer accepting that.

juberti:
Exactly. That's what the text in 4.1.4.1 suggests.

fluffy:
What's needed to merge this one? Given it just and example we could mark it with "needs review or something" and add it in

ekr:
I had understood on our last call that the idea was to restructure this
without PRAnswer,
so I haven't even looked at it. I would suggest deferring this to post the
next draft.

fluffy:
that has been done. The old pr-answer stuff is all commented out.

ekr:
Unfortunately, I missed the review request. Kind of a misfeature in github UI. Unless it seems urgent, I suggest we leave this PR for the next cycle.

fluffy:
I don't care if we add this now or later and I forgot to add the r?ekr after I changed the example

juberti:
I concur with Eric.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,933 @@ a=end-of-candidates
...
+a=candidate:4036177503 1 udp 1685987071 11.22.33.44 52546 typ srflx
+            raddr 192.168.1.2 rport 51556
+a=candidate:3671762466 1 udp 41819903 22.33.44.55 61405 typ relay
+            raddr 11.22.33.44 rport 52546
+a=end-of-candidates
+         ]]></artwork>
+       </figure>
+     </t>
+    </section>
+    -->
```
ekr: Can you remove this large commented out chunk and then reland it when you have it done. It's weird to have it in the doc but commented out.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+      audio and video streams. The advantage of such a
+      flow is that as soon as the first answer is received, the browser can
+      proceed with ICE and DTLS negotiation and have the transport set up and
+      working. If the transport setup completes before the second offer is
+      sent -- and bundle is used -- then media can be transmitted immediately
+      upon completion of the offer/answer exchange. This technique allows
+      the application to parallelize call setup with user input, thus reducing
+      perceived post-dial-delay. The second offer/answer exchange can change
+      the selected codec or other information. </t>
+
```
ekr: Looking at this example, why bother with inactive? Recall that this is generally in the context of a Web application. Why not offer *just* the data channel and then add the two new channels? The application can easily convey what sort of call to ask for to the user out of band of the SDP.

This change would have the example that we could probably actually get Firefox or Chrome to easily emit it, thus saving potential errors.

@juberti, @fluffy wdyt?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+         ]]></artwork>
+       </figure>
+     </t>
+    </section>
+    -->
+
+    <section title="Inactive Media Answer Example" anchor="sec.inactive-examples" >
+
+      <t> In this example, browser A sends an offer to browser
+      B with an offer for data, audio, and two video tracks. Browser B immediately
```
juberti: Having two video tracks is probably unnecessary complexity. I even wonder if we should include the data channel here.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+a=end-of-candidates
+         ]]></artwork>
+       </figure>
+     </t>
+    </section>
+    -->
+
+    <section title="Inactive Media Answer Example" anchor="sec.inactive-examples" >
+
+      <t> In this example, browser A sends an offer to browser
```
juberti: Suggest referencing 4.1.4.1, which first mentions this 2-offer-answer scenario.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+      as inactive. This allows the browser to start negotiating ICE and DTLS
+      immediately. B then takes some time, perhaps prompting the user to
+      answer the call or for permission to use the microphone and/or camera.
+      At some future point in time B, sends a second offer which enables the
+      audio and video streams. The advantage of such a
+      flow is that as soon as the first answer is received, the browser can
+      proceed with ICE and DTLS negotiation and have the transport set up and
+      working. If the transport setup completes before the second offer is
+      sent -- and bundle is used -- then media can be transmitted immediately
+      upon completion of the offer/answer exchange. This technique allows
```
juberti: I think media transmission can start immediately upon the callee's 2nd setLocal. The caller is ready to receive at this point.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+      the application to parallelize call setup with user input, thus reducing
+      perceived post-dial-delay. The second offer/answer exchange can change
```
juberti: I don't think this is relevant to post-dial-delay, which is time between dial and ringing. This minimizes perceived call setup time for the callee (and possibly the caller), and greatly reduces the chances of voice clipping for the callee.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+      the selected codec or other information. </t>
+      
+     <t> 
+       The SDP for |offer-C1| looks like:
+        </t>
+       <t>
+       <figure>
+         <artwork><![CDATA[
+v=0
```
juberti: To avoid errors, we should probably make this SDP identical to the "complex example" we currently have in the document (or at least the relevant parts of this SDP should be identical).

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+a=ice-options:trickle
+a=fingerprint:sha-256 6B:8B:F0:65:5F:78:E2:51:3B:AC:6F:F3:3F:46:1B:35
+                     :DC:B8:5F:64:1A:24:C2:43:F0:A1:58:D0:A1:2C:19:08
+a=setup:actpass
+a=rtcp-mux
+a=rtcp-rsize
+a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
+a=extmap:2 urn:ietf:params:rtp-hdrext:sdes:mid
+a=ssrc:4429951804 cname:Q/NWs1ao1HmN4Xa5
+a=candidate:109270924 1 udp 2122194687 192.168.2.3 61665 typ host
```
juberti: Initial offer shouldn't contain candidates

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+t=0 0
+a=msid-semantic:WMS
+a=group:BUNDLE a1 d1 v1 v2
+m=audio 52546 UDP/TLS/RTP/SAVPF 96 0 8 97 98
+c=IN IP4 11.22.33.44
+a=rtcp:52546 IN IP4 11.22.33.44
+a=mid:a1
+a=msid:57017fee-b6c1-4162-929c-a25110252400
+       e83006c5-a0ff-4e0a-9ed9-d3e6747be7d9
+a=inactvie
```
juberti: typo on inactive. This should probably be the same SDP as the previous complex example, just with a=inactive.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+a=fingerprint:sha-256
+              19:E2:1C:3B:4B:9F:81:E6:B8:5C:F4:A5:A8:D8:73:04
+             :BB:05:2F:70:9F:04:A9:0E:05:E9:26:33:E8:70:88:A2
+a=setup:passive
+a=rtcp-mux
+a=rtcp-rsize
+a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
+a=extmap:2 urn:ietf:params:rtp-hdrext:sdes:mid
+a=ssrc:1732846380 cname:FocUG1f0fcg/yvY7
+a=candidate:109270923 1 udp 2122194687 192.168.1.2 51556 typ host
```
juberti: again no candidates in initial answer

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+a=candidate:3671762466 1 udp 41819903 22.33.44.55 61405 typ relay
+            raddr 11.22.33.44 rport 52546
+a=end-of-candidates
+         ]]></artwork>
+       </figure>
+       </t>
+
+
+         <t>
+       The SDP for |offer-C3| which is a answer looks like:
```
juberti: This terminology is confusing... it's still an offer, just emitted by the callee upon accepting the call.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+     </t>
+       <t>
+       <figure>
+         <artwork><![CDATA[
+v=0
```
juberti: This SDP should basically be the same as the offer, except no longer marked as inactive (we should say so explicitly)

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+       </t>
+         
+     <t>
+       The SDP for |answer-C4| which is a answer looks like:
+     </t>
+       <t>
+       <figure>
+         <artwork><![CDATA[
+v=0
+o=- 4962303333179871723 2 IN IP4 0.0.0.0
```
juberti: This should basically be identical to C1 (and we should say to explicitly).

draft-ietf-rtcweb-jsep.xml
ekr: Hmm... Are you allowed to just change a stream from inactive to sendrecv without receiving an answer like that?

draft-ietf-rtcweb-jsep.xml
ekr: Isn't the data channel needed so you set up one stream?

draft-ietf-rtcweb-jsep.xml
ekr: This may be true, but I think a citation to where this is allowed in 3264 would be good.

@martinthomson, @fluffy do you have one to hand?

draft-ietf-rtcweb-jsep.xml
ekr: Well, it depends when you do the call to action, but I agree your phrasing is better. post-answer delay is the right term here.

draft-ietf-rtcweb-jsep.xml
ekr: Generally, I'm starting to think we should make these examples with Chrome or Firefox.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3163,6 +3164,935 @@ a=end-of-candidates
...
+
+    <section title="Inactive Media Answer Example" anchor="sec.inactive-examples" >
+
+      <t> In this example, browser A sends an offer to browser
+      B with an offer for data, audio, and two video tracks. Browser B immediately
+      responds with an answer that accepts the data channel but marks the rest
+      as inactive. This allows the browser to start negotiating ICE and DTLS
+      immediately. B then takes some time, perhaps prompting the user to
+      answer the call or for permission to use the microphone and/or camera.
+      At some future point in time B, sends a second offer which enables the
```
martinthomson: confusing comma

draft-ietf-rtcweb-jsep.xml
martinthomson: I don't think that it needs citation, it's physics.  Citing 3264 would suffice.  In O/A, the A party can send to the O party as soon as it receives the O.  There is no further communication upon which the A party might condition its behaviour.

martinthomson: Hmm, re-reading this, I'm not sure if this is right either.  It's unclear which offer/answer exchange this last refers to.  If you have negotiated an m= section as inactive, then I don't believe that you can unilaterally activate that section, you have to wait for an answer.  Nothing in the definition of `a={sendrecv|sendonly|recvonly|inactive}` in RFC 4566 supports that of course (it's amazingly terse).

RFC 3264 suggests that you can't unilaterally send on inactive in the sense that the new offer has to be compatible with the previous session state, but RFC 3264 never says anything like that, so I guess that an alternative interpretation is *technically* valid; crazier interpretations of omissions have been proposed.  [Paragraph 2, Section 6.1 of ](https://tools.ietf.org/html/rfc3264#section-6.1) would seem to support the notion that anything other than `inactive` is incompatible with an `inactive` state.

draft-ietf-rtcweb-jsep.xml
juberti: Even if a=inactive, the transport should still be established. RTCP will still flow even if RTP is inactive.

draft-ietf-rtcweb-jsep.xml
juberti: I don't get that impression from that paragraph. Two arguments for my interpretation below.

Argument #1: Let's first agree that the callee could return a a=sendonly answer to the caller (indicating that the incoming stream, from the callee's perspective, is disabled). The offer was a=sendrecv, so returning a=sendonly in the answer is clearly legal. And in this state, the callee could still send media any time it wanted to, and sending a new offer with a=sendrecv would indicate that now the *caller* should send media.

This is essentially what we want, except for the fact that providing a=sendonly to setLocalDescription on the callee side will result in the callee sending media prematurely. We could try to prevent that by setting a=inactive on the callee setLocal, but sending a=sendonly to the caller in the first answer; the second offer from the callee would then change this to a=sendrecv and media would flow in both directions. 

However, it would certainly be simplest to just allow a=inactive to be used for both the callee setLocal and sent to the caller in the answer (where it is used in setRemote).

Argument #2: changing from a=inactive to a=sendrecv is essentially ending local and remote hold (as described in RFC3264, S 8.4). My reading of 8.4 is that the update of the directionality attribute is advisory; one doesn't need to wait for the answer to start/stop sending media.

fluffy:
The more I look at this I would like to move all the example to the example draft

juberti:
Overall I think this is still worth finishing. I think we need to provide a bit more prose that explains the high-level flow here, and that there is some other mechanism (i.e. dummy tracks) that prevents the callee from sending to the caller until the callee accepts, as well as the fact that the caller is in a=recvonly mode until it gets the second answer.

Note also that the callee will likely initially only gather relay candidates for the early warmup, and only go to full gathering with the second setLocal, so this example also illustrates that policy from S 3.4.3.

Lastly, I think we should try to strip out as much as possible from this example just to keep it compact. We already have an example with a data channel, so perhaps we could simply make this a one-audio/one-video example (or perhaps even omit video altogether).

I will see if I can hack on this a bit and push an update reflecting the above to the branch.

fluffy:
I think this fixes #15 too

juberti:
Closing this for now until I can polish it further

PR #137, opened 2015-05-10
Lipsync4

### LINKED ISSUE #31: How do we signal synchronization
Cullen wants LS groups because that is what legacy gear understands.

Magnus wants to use MSID, which carries this.

fluffy:
Fixes #31

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1304,7 +1304,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
 	  <t>The next step is to generate lip sync groups as defined
 	  in <xref target="RFC5888"/>, Section 7.  For each
-	  MediaStream with more than one MediaStreamTracks, an group
+	  MediaStream with more than one MediaStreamTrack, a group
 	  of type "LS" MUST be added that contains the mid values for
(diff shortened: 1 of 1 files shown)


ekr:
I cleaned up the language.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1302,6 +1302,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <xref target="I-D.ietf-mmusic-msid"></xref>, Section 4.</t>
             </list></t>
 
+	  <t>The next step is to generate lip sync groups as defined
```
juberti: Since the mids are not all known at this point (only after m= sections are generated), I think this needs to happen down at line 1493, next to where we create the BUNDLE group.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1770,6 +1779,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           session-level attributes. The process here is identical to that
           indicated in the Initial Offers section above.</t>
 
+	  <t>The next step is to generate lip sync groups as defined
```
juberti: See above

juberti:
lgtm

PR #141, opened 2015-06-05
Remove a=ice-lite in m= sections. Fixes #126

### LINKED ISSUE #126: ice-lite is not valid at media level
http://rtcweb-wg.github.io/jsep/#rfc.section.5.6.2 mentions its use at media level; this should be clarified.

ekr:


fluffy:
LGTM

PR #142, opened 2015-06-06
Ignore c= line. Fixes #121

### LINKED ISSUE #121: Explicitly prohibit ICE mismatch processing
When JSEP recommends that `c=` lines are ignored (absent some syntax validation), it violates a rule in RFC 5245 regarding mismatch processing.  5245 requires that the values for port and IP address in `c=` and `m=` correspond to a candidate that appears in the media section.

Consensus in Dallas was to prohibit checking of the value in `c=` against `m=` and `a=candidate:`, which requires an explicit mention in the draft, since this is different to what 5245 says.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2151,7 +2151,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           </t>
           <t>The remaining lines are processed as follows:
             <list>
-              <t>The "c=" line MUST be parsed and stored.</t>
+              <t>The "c=" line MUST be checked for syntax but its value SHOULD be ignored.
```
juberti: should this go into the TODO section at line 2149?

ekr: Sure. Will move it.

ekr:
@juberti is this OK? @fluffy

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2146,12 +2146,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             lines are not meaningful in the JSEP context and
             MAY be discarded once they have been checked.
             <list>
+              <t>The "c=" line MUST be checked for syntax but its value SHOULD be ignored.
+              This supersedes the guidance in <xref target="RFC5245"/> Section 6.1,
```
juberti: style in this document is
[RFC5245], Section 6.1,

(i.e. comma before Section)

juberti:
lgtm with the one comment above

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2146,12 +2146,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             lines are not meaningful in the JSEP context and
             MAY be discarded once they have been checked.
             <list>
+              <t>The "c=" line MUST be checked for syntax but its value SHOULD be ignored.
+              This supersedes the guidance in <xref target="RFC5245"/>, Section 6.1,
+              to use "ice-mismatch" to indicate mismatches between "c=" and
+              the candidate lines. Because JSEP always uses ICE--and generally
+              uses trickle ICE--"ice-mismatch" is not useful in this case.
```
fluffy: Give the state of trickle ice, I should like to remove the "--and generally uses trickle ICE--".

fluffy:
Ignored is the wrong word to use here. The value is not used is true but ignored is just simply not true.

draft-ietf-rtcweb-jsep.xml
ekr: All right.

ekr:
Hmm.... We're not checking it and not doing anything with it. I'm happy to
use another
word, but I'm going to need you to suggest one.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2146,12 +2146,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             lines are not meaningful in the JSEP context and
             MAY be discarded once they have been checked.
             <list>
+              <t>The "c=" line MUST be checked for syntax but its value SHOULD be ignored.
```
fluffy: suggestions ... change "SHOULD be ignored" to "is not used"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2205,8 +2210,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <t>Following the "m=" line, specific processing MUST be applied for the
             following non-attribute lines:
             <list style="symbols">
-              <t>The "c=" line, if present, MUST be parsed as specified in
-                <xref target="RFC4566"/>, Section 5.7, and its contents stored.</t>
+              <t>
+                As with the "c=" line at the session level, 
+                the "c=" line MUST be parsed according to
+                <xref target="RFC4566"/>, Section 5.7, but its value SHOULD be
+              ignored.</t>
```
fluffy: suggestion ... change "SHOULD be ignored" to "is not used".

draft-ietf-rtcweb-jsep.xml
ekr: Fine with me. @juberti, if you agree, say so and I will make the change and
merge.

ekr:
@fluffy please r+ if you are happy.
@juberti please merge if you are happy.

fluffy:
I really don't care but it like 1 of the 3 things I raised got fixed

ekr:
I see I missed the trickle thing. What else is there?

fluffy:
Around line 2216 I had another change "SHOULD be ignored" to "is not used". I think we are running into the limitations of using github for review here.

ekr:
Indeed, I think I missed it.

ekr:
@juberti: I think this is ready to merge if you are still happy.

fluffy:
LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2146,12 +2146,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             lines are not meaningful in the JSEP context and
             MAY be discarded once they have been checked.
             <list>
+              <t>The "c=" line MUST be checked for syntax but its
+              value is not used.  This supersedes the guidance in <xref
+              target="RFC5245"/>, Section 6.1, to use "ice-mismatch" to indicate
+              mismatches between "c=" and the candidate lines. Because JSEP always
```
juberti: would sound better as 
"... and the candidate lines; because JSEP always uses ICE, "ice-mismatch" is not useful in this context."

juberti:
lgtm with minor change

PR #150, opened 2015-06-12
Overview of a=imageattr handling

### LINKED ISSUE #5: Handling of a=imageattr

juberti:
Still needs updates to createOffer/Answer. Fixes #5

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -582,18 +582,25 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             hard limits with any mandatory constraints
             will be used as the x= and y= values for the generated "a=imageattr"
             attribute for the associated MediaStreamTrack.
-            With only one set of preferences expressed, the q= value is not
-            important; it SHOULD be set to 1.0.
           </t>
           <t>If the intersection is the null
             set, i.e., there are no resolutions that
             are permitted by both the decoder and the mandatory constraints,
-            this SHOULD be represented using x=0 and y=0 values.</t>
+            this SHOULD be represented by an "a=imageattr" attribute with
+            x=0 and y=0 values.</t>
+          <t>
+            If the decoder limits
+            are unknown, e.g. when using a software decoder, the mandatory
+            constraints are used directly. If there are no known decoder limits
+            or mandatory constraints, the "a=imageattr" attribute SHOULD be
+            omitted.</t>
+          <t>The rules here express a single set of preferences, and therefore,
+            the "a=imageattr" q= value is not important. It SHOULD be set to 1.0.</t>
           <t>The "a=imageattr" field is payload type specific.
             When all video codecs supported have the same capabilities, use of a
             single attribute, with the wildcard payload type (*), is RECOMMENDED.
             However, when the supported video codecs have differing capabilities,
-            a specific "a=imageattr" attribute is inserted for each payload
+            specific "a=imageattr" attributes MUST be inserted for each payload
             type.
           </t>
           <t>As an example, consider a system with a HD-capable, multiformat
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -569,7 +569,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <t>In order to determine the limits on what video resolution a
             receiver wants to receive, it will intersect its decoder hard
             limits with any mandatory constraints that have been applied to
-            the associated MediaStreamTrack. For the answerer, these mandatory
+            the associated MediaStreamTrack. If the decoder limits
+            are unknown, e.g. when using a software decoder, the mandatory
+            constraints are used directly. For the answerer, these mandatory
             constraints can be applied to the remote MediaStreamTracks that
             are created by a setRemoteDescription call, and will affect the
             output of the ensuing createAnswer call. Any constraints set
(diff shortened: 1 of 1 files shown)


ekr:
Travis still seems sad.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -556,6 +556,64 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          constrain what it receives due to application preferences, e.g. a
+          specific size for the window in which the video will be displayed.
+        </t>
+
+        <section title="Creating an imageattr Attribute">
+          <t>In order to determine the limits on what video resolution a
+            receiver wants to receive, it will intersect its decoder hard
+            limits with any mandatory constraints that have been applied to
+            the associated MediaStreamTrack. The resulting resolution space
+            will be used as the x= and y= values for the generated "a=imageattr"
```
ekr: This seems unclear. Is your idea that this was you do applyConstraints in ontrackadded. Perhaps some more text would help.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -556,6 +556,64 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        </section>
+
+        <section title="Interpreting an imageattr Attribute">
+          <t>
+            <xref target="RFC6236"/> defines "a=imageattr" to be an advisory
+            field. This means that it does not absolutely constrain the video
+            formats that the sender can use, but gives an indication of the
+            preferred values.</t>
+            <t>This specification attempts to define more specific behavior.
+            When a sender of a given MediaStreamTrack, which is producing video
```
ekr: s/attempts to define/prescribes/?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -556,6 +556,64 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            of a certain resolution, receives an "a=imageattr recv"
+            attribute, it will first check to see if the original resolution
+            meets the criteria specified in the attribute, and transmit it
```
ekr: s/will/MUST/

draft-ietf-rtcweb-jsep.xml
```diff
@@ -556,6 +556,64 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            untouched if so. If the original resolution is too large for the
+            attribute criteria, the sender SHOULD apply downscaling to the
+            output of the MediaStreamTrack in order to satisfy the criteria.
+            In rare cases, where a receiver requires a specific resolution, the
+            sender MAY apply upscaling in order to provide that resolution.
+            However, if there is no appropriate scaling mechanism that allows
```
ekr: We should probably either encourage or discourage this not MAY it.

ekr:
Since I won't be on the call, I give this my preemptive LGTM, provided you at least look at my comments (and maybe reject them) and fix the build

draft-ietf-rtcweb-jsep.xml
juberti: yes, or really any time prior to SetLD(answer).

On the offerer side, this is problematic, as the track doesn't exist prior to setRD. While this could be addressed by a separate offer-answer exchange, this suggests that this mechanism is not ideal.

Agree the text should discuss this.

draft-ietf-rtcweb-jsep.xml
juberti: yeah, 'prescribes' is probably more appropriate

draft-ietf-rtcweb-jsep.xml
juberti: OK. SHOULD it is.

juberti:
Eric, did you want to take another look at this?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -556,6 +556,82 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            receiver wants to receive, it will intersect its decoder hard
+            limits with any mandatory constraints that have been applied to
+            the associated MediaStreamTrack. For the answerer, these mandatory
+            constraints can be applied to the remote MediaStreamTracks that
+            are created by a setRemoteDescription call, and will affect the
+            output of the ensuing createAnswer call. For the offerer, since it
+            does not know about any remote MediaStreamTracks until it receives
+            theanswer, this implies that it can only set these constraints after
+            the initial offer-answer exchange is complete, and as a result, a
+            new offer-answer is needed to communicate them.</t>
```
fluffy: A sends offer to B. Lets say it has 

   a=imageattr:97 send * recv *

B wants to limit it was it receives to 360p and sends back an pr-answer with 
    
    a=imageattr:* recv [x=[16:640],y=[16:360],q=1.0]

and A starts sending 360p to B when it gets this pr-answer. 

I'm not seeing why an 2nd offer/answer is needed. You do an answer of pr-answer before the constraint can take place but given an pr-answer will work, I don't think it is right to say that the offer-answer needs to complete. 

It seems like replacing "the initial offer-answer exchange is complete, and as a result, a new offer-answer is needed to communicate them"  with "receiving the initial answer or pr-answer. " would fix this.

fluffy:
There one sentence I think we need to deal with (see other comment) but rest of this LGTM

draft-ietf-rtcweb-jsep.xml
ekr: I think you may be interpreting the text here different from what is intended.

Say that we take the example you give except that once A calls setRD(answer) on B's answer, the JS applies a constraint to the resulting media stream that restricts it to 360p. There is no way for A to communicate that to B in this exchange, because it has concluded, and B is left thinking that it can send at any rate. In order to adjust B's ideas, A needs to send a new offer with a new a=imageattr. Do you disagree with this?

fluffy: OK - I think I'm starting to get one this is getting at. Let's separate two cases here. Case 1 is A is mobile phone that the actually browser knows that due to hardware limitations it can not receive more than 360p, and case 2 where it's a desktop browser that can receive big stuff but the JS app wants to limit this particular video stream to 360p.

In case 1, the limit that A can only receive 360p would have been in the initial Offer. 

In case 2, what you are saying is that as the JS API layer, we have no way to set this limit until after an answer or pr-answer is received by A? It seems we need to fix this API layer not so much for this but more for simulcast. 

Agree of course that any time A wants to change it's constraints after sending the Offer, it's going to need to do new O/A exchange.

ekr: I agree with your analysis here, but it seems like for now we should be landing effectively what Justin proposed. Is there something you think would make this clearer?

ekr: Yes, I thin this case analysis is correct, and yes, I think this does reveal a problem in the
API, which you would presumably need to resolve by having some sort of template in
offerToReceiveFoo.

fluffy:
It still seems like my proposed change of replacing "the initial offer-answer exchange is complete, and as a result, a new offer-answer is needed to communicate them" with "receiving the initial answer or pr-answer. " would make this more or less right for both cases and as the API developers, or does not develop, for this then we can fill in more details

ekr:
Well, as far as I can tell, you can't do it after the pr-answer, since you can't do an offer at that stage.

ekr:
@juberti PTAL at PR #153. I think with this change we are ready to go. If that makes you happy, merge in PR #153 into here and then merge this PR.

fluffy:
With EKR's #153 this LGTM

juberti:
OK, I will push the buttons now.

juberti:
OK, I worked this over a bit to incorporate PR #153 and also handle the case where the decoder hard limits are not known. The changes are minor but I will hold off for the next few hours to get a sanity check from one of you.

ekr:
LGTM

fluffy:
I will read it and get back to you I  30 minutes

fluffy:
LGTM

ekr:
Fixed merge conflict and merged.

PR #153, opened 2015-06-15
Clean up text a bit

ekr:


fluffy:
LGTM - I think that makes it clear what can and can not work.

juberti:
lgtm

PR #164, opened 2015-06-29
Make clear that a=imagettr is optional. Consistency. Fixes #160

### LINKED ISSUE #160: Contradicting statements about a=imageattr between Section 3.6 and 5.2.1
G. Section 5.2.1:

This section contains the following:

   Each m= section MUST include the following attribute lines:

   o  If this m= section is for video media, an "a=imageattr" line, as
      specified in Section 3.6.

But, Section 3.6 does not mandate that a=imageattr always is included. Thus the text must be modified to allow the selective inclusion of the attribute based on what needs to be said.

ekr:


fluffy:
LGTM

Issue #162, opened 2015-06-18
LipSync groups violating RFC5888


gloinul:
I. Section 5.3.1:

   The next step is to generate lip sync groups as defined in [RFC5888],
   Section 7.  For each MediaStream with more than one MediaStreamTrack,
   a group of type "LS" MUST be added that contains the mid values for
   each MediaStreamTrack in that MediaStream.  In some cases this may
   result in adding a mid to a given LS group that was not in that LS
   group in the associated offer.  Although this is not allowed by
   [RFC5888], it is allowed when implementing this specification.

So, do I understand it correctly the violation this paragraph is discussing is the following from Section 9.2 of RFC5888:

   The
   identification-tags contained in this "a=group" line MUST be the same
   as those received in the offer, or a subset of them (zero
   identification-tags is a valid subset).

So when will these violations occur for a LS group in JSEP. I assume due to that offer and answer has different structure of how MediaStreams uses the different m= lines. But, isn't this just another case of when an answerer will have to do an follow up Offer to be able to add all the MediaStreamTracks and their MediaStreams due to different sets on the two peers?

I don't really think this specification should override RFC 5888. The obvious solution is to make a new offer by the answering part who couldn't express its desired configuration in the first round.

fluffy:
I suspect we have two paths here.  A) make a 5888 bis that allows what is currently in draft (and this may turn into bis go group semantics too)   B) just use 5888.   B is slightly less effecent in certian cases than A but it seems like that case will not come up often so I'm not sure it matters much. We should talk about how relevant that case is and if it is not a big deal, B is the fastest path to done

juberti:
Agree that B seems like the easiest path, and we can definitely extend it to A later without too much trouble. We will note the issues that exist and how they could be dealt with by future generations.

The overall fundamental question is do we stay with LS or do something else (e.g. MSID) and for now we are going to stick with LS.

juberti:
We probably need to have some text indicating that LS is what determines what MediaStream a given MediaStreamTrack goes into. Ergo, this reduces the applicability of MSID (which is still used to identify that there is a sending track for a m= section)

ekr:
I think this needs to be done by @fluffy. I am happy to wordsmith afterwards

fluffy:
Have a look at https://github.com/rtcweb-wg/jsep/pull/167

alvestrand:
My suggestion for a simple (???) track assignment algorithm here:

[...]

PR #167, opened 2015-07-02
changed to not violate 5888 for LS groups

fluffy:
Fixes https://github.com/rtcweb-wg/jsep/issues/162

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1600,7 +1600,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>The next step is to generate session-level lip sync groups as
           defined in <xref target="RFC5888"/>, Section 7.  For each MediaStream
-          with more than one MediaStreamTrack, a group of type "LS" MUST be
+          with more than one MediaStreamTrack, a group of type LS MUST be
           added that contains the mid values for each MediaStreamTrack in that
           MediaStream. </t>
 
@@ -1885,15 +1885,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           session-level attributes. The process here is identical to that
           indicated in the Initial Offers section above.</t>
 
-	  <t>The next step is to generate lip sync groups as defined
-	  in <xref target="RFC5888"/>, Section 7.  For each
-	  MediaStream with more than one MediaStreamTrack, a group
-	  of type "LS" MUST be added that contains the mid values for
-	  each MediaStreamTrack in that MediaStream. In some cases
-	  this may result in adding a mid to a given LS group that
-	  was not in that LS group in the associated offer. Although
-	  this is not allowed by <xref target="RFC5888"/>, it is
-	  allowed when implementing this specification. </t>
+	  <t>The next step is to generate session-level lip sync
+	  groups as defined in <xref target="RFC5888"/>, Section
+	  7. All Tracks in the same MediaStream need to be placed in
+	  the same LS group.  RFC5888 only allows the groups in the
+	  answer to be a subset of what was in the offer. If the
+	  answer wishes to have two things synchronized that were not
+	  in the same LS group in the in the offer, another offer
+	  exchange is needed to negotiate this. </t>
 
           <t>The next step is to generate m= sections for each m= section that
           is present in the remote offer, as specified in <xref
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1896,10 +1896,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 	      answer to be a subset of what was in the offer. If there are
 	      tracks that the same MediaStream that can not be added to
 	      the "LS" group, once the state moves back to "stable" the
-	      browser SHOULD initiate another offer/answer negotiation put
-	      all the tracks in the MediaStream into the "LS" group. [[OPEN ISSUE:
-          https://github.com/rtcweb-wg/jsep/issues/162]]</t>
-
+	      browser MAY initiate another offer/answer negotiation put
+	      all the tracks in the MediaStream into the "LS" group or they can be
+          left in a state that though answer wishes them to be synchronized, the
+          answer can not signal them as synchronized because the offer did put
+          them in the same LS group. The working group consider varios solution
+          to this rare corner condition and decided not to try to automatically
+          solve it. Applications that run into can do an additional round of
+          signaling reversing the offer / answer roles to resolve the issue if
+          it is critical for their use case. </t>
+          
           <t>The next step is to generate m= sections for each m= section that
           is present in the remote offer, as specified in <xref
           target="RFC3264"></xref>, Section 6. For the purposes of this
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1600,7 +1600,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>The next step is to generate session-level lip sync groups as
           defined in <xref target="RFC5888"/>, Section 7.  For each MediaStream
-          with more than one MediaStreamTrack, a group of type LS MUST be
+          with more than one MediaStreamTrack, a group of type "LS" MUST be
           added that contains the mid values for each MediaStreamTrack in that
           MediaStream. </t>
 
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1600,7 +1600,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>The next step is to generate session-level lip sync groups as
           defined in <xref target="RFC5888"/>, Section 7.  For each MediaStream
-          with more than one MediaStreamTrack, a group of type "LS" MUST be
+          with more than one MediaStreamTrack, a group of type LS MUST be
```
juberti: we used quotes in the preceding paragraph, so I think we should keep them here too. We may want to say "with semantics "LS"" instead of "of type LS" as well.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1885,15 +1885,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-	  was not in that LS group in the associated offer. Although
-	  this is not allowed by <xref target="RFC5888"/>, it is
-	  allowed when implementing this specification. </t>
+	  <t>The next step is to generate session-level lip sync
+	  groups as defined in <xref target="RFC5888"/>, Section
+	  7. All Tracks in the same MediaStream need to be placed in
+	  the same LS group.  RFC5888 only allows the groups in the
+	  answer to be a subset of what was in the offer. If the
+	  answer wishes to have two things synchronized that were not
+	  in the same LS group in the in the offer, another offer
```
juberti: I don't think this is sufficient guidance; consider the case where the caller does not want to sync audio/video, but the callee does. The caller offers an empty LS group, and because 5888, the callee can only reply with an empty LS group. But when the callee reoffers a LS group with audio and video, the caller will reject that LS group, since it doesn't match its expectations.

I think we need to simply say that such a case doesn't work right now, and is left as a problem for future study. (In reality I expect all implementations to just implement declarative behavior.)

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1885,15 +1885,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-	  MediaStream with more than one MediaStreamTrack, a group
-	  of type "LS" MUST be added that contains the mid values for
-	  each MediaStreamTrack in that MediaStream. In some cases
-	  this may result in adding a mid to a given LS group that
-	  was not in that LS group in the associated offer. Although
-	  this is not allowed by <xref target="RFC5888"/>, it is
-	  allowed when implementing this specification. </t>
+	  <t>The next step is to generate session-level lip sync
+	  groups as defined in <xref target="RFC5888"/>, Section
+	  7. All Tracks in the same MediaStream need to be placed in
```
juberti: I don't think this gives the answerer enough guidance in what it should generate.

I think we should keep the previous guidance, but simply add that if the set of mids formed does not comprise a subset of the offered LS mids, any mids not in the offered group MUST be omitted, per 5888, and therefore left unsynchronized.

ekr: Justin, do you want to update this PR?

ekr: @juberti

fluffy:
OK - I took another stab at addressing the what to do. Obviosly more detail could be added over time but I think this gets it close to publish and tlak about

draft-ietf-rtcweb-jsep.xml
juberti: I have limited access, but let me try to type something up here on Monday. I see you posted a new version, but perhaps we could still close out this PR in the next day or two.

juberti: Text me or email me at my gmail address if you need to get hold of me before the deadline.

ekr: Sure, I'm happy to post -12 if we have the changes in time :)

-Ekr

fluffy: I did put some updated text - if it looks ok to you, or we can easily fix it, the deadline is 5 pm tomorrow

Sent with my thumbs - sorry

On Jul 5, 2015, at 8:13 PM, Justin Uberti <notifications@github.com<mailto:notifications@github.com>> wrote:


In draft-ietf-rtcweb-jsep.xml<https://github.com/rtcweb-wg/jsep/pull/167#discussion_r33903866>:

> @@ -1885,15 +1885,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
>            session-level attributes. The process here is identical to that
>            indicated in the Initial Offers section above.</t>
>
> -       <t>The next step is to generate lip sync groups as defined
> -       in <xref target="RFC5888"/>, Section 7.  For each
> -       MediaStream with more than one MediaStreamTrack, a group
> -       of type "LS" MUST be added that contains the mid values for
> -       each MediaStreamTrack in that MediaStream. In some cases
> -       this may result in adding a mid to a given LS group that
> -       was not in that LS group in the associated offer. Although
> -       this is not allowed by <xref target="RFC5888"/>, it is
> -       allowed when implementing this specification. </t>
> +       <t>The next step is to generate session-level lip sync
> +       groups as defined in <xref target="RFC5888"/>, Section
> +       7. All Tracks in the same MediaStream need to be placed in


I have limited access, but let me try to type something up here on Monday. I see you posted a new version, but perhaps we could still close out this PR in the next day or two.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1885,15 +1885,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	  the same "LS" group as long as theses tracks were in the
+	  offered to by synchronized in the offer.
```
alvestrand: Suggest alternate text:

If all tracks of a MediaStream that the answerer wants to send fit within the media sections of an LS group in the offer, these media sections (and only these media sections) are used for the MediaStream. If there are media sections in the offer that are not part of an LS group, these can be used by any track belonging to any stream.
In all cases, the LS groups in the answer are identical to the LS groups in the offer.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1885,15 +1885,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	  RFC5888 only allows the groups in the
+	  answer to be a subset of what was in the offer. If there are
+	  tracks that the same MediaStream that can not be added to
+	  the "LS" group, once the state moves back to "stable" the
```
alvestrand: This seems nearly right to me. Suggest:

If there are tracks that are placed in the answer, but not within a suitable LS group, or tracks that cannot be added to the answer because of a lack of suitable media sections, once the state moves back to "stable" the browser SHOULD initiate another offer/answer.....

ekr:
Consensus from today's call: Justin to try to come up with some examples of how the proposed algorithm is problematic and then surface in Yokohama.

fluffy:
latest plan for cullen to redo based on idea tha if we got in the wird case, it was just too bad and did not get

fluffy:
@ekr - Could you take an editorial pass over this. Text is short.

aboba:
How about this? 

+ 	      <t>The next step is to generate session-level lip sync
 +	      groups as defined in <xref target="RFC5888"/>, Section
 +	      7. All Tracks in the same MediaStream need to be placed in
 +	      the same "LS" group, as long as these tracks were intended
 +	      to be synchronized in the offer.  RFC5888 only allows LS groups in the      
 +	      answer to be a subset of what was in the offer. If there are tracks
 +	      in the same MediaStream that the Answerer desires to add to
 +	      the "LS" group, once the state moves back to "stable" the
 +	      answering browser MAY initiate another offer/answer negotiation, putting
 +	      all the tracks in the MediaStream into the "LS" group.  Without initiating
 +           another O/A exchange, the answering browser cannot add tracks to the LS group
 +           that were not in the Offer. Rather than adding complexity to address this corner
 +           condition in a single O/A exchange, the Working Group recommends that applications
+            needing to solve this problem utilize an additional round of signaling, reversing the 
+            offer / answer roles. </t>

juberti:
I agree that if the callee prefs don't match the caller sync prefs, it's too bad - that's the conclusion we came to as a group.

But any discussion of the additional round of signaling needs to touch on transceivers. Otherwise, there's nothing the callee can do if its sync preferences don't match what's offered; the automatic assignment of MSTs to m= lines will simply ignore the callee's preferences, so the only way to do the additional signaling is to create new transceivers at the callee side, and explicitly stuff the callee tracks into the new transceivers (as opposed to the offered transceivers).

cc @pthatcherg

juberti:
Justin to take another shot at this and leave a TODO to reference the section on transceivers once it exists.

fluffy:
This has been replace with new PR at https://github.com/rtcweb-wg/jsep/pull/263

PR #172, opened 2015-08-21
Add additional clarity on ICE default candidates. Fixes #171

### LINKED ISSUE #171: What should c= be during trickling?
The question has come up about what the default candidate should be set to during trickle ICE
but after at least one candidate has been gathered so at least in principle you could set c= to one of those candidates. See also;
https://bugzilla.mozilla.org/show_bug.cgi?id=1192813

ekr:


fluffy:
LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1642,7 +1642,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
               <t>Each "m=" and c=" line MUST be filled in with the port, protocol,
               and address of the default candidate for the m= section, as described
-              in <xref target="RFC5245"></xref>, Section 4.3. Each
+              in <xref target="RFC5245"></xref>, Section 4.3. If ICE checking
+              has already completed for one or more candidate pairs and a candidate
+              pair is in active use, then that pair MUST be used, even if ICE
+              has not yet completed. Each
```
juberti: Do you want to add a reference to 5245, S 9.1.2.2 here, or mention how this slightly differs from the guidance offered there (which refers to ICE completed)?
https://tools.ietf.org/html/rfc5245#section-9.1.2.2

juberti:
lgtm

PR #177, opened 2015-08-21
Describe how to handle OfferToReceiveAudio/Video. Fixes #140

### LINKED ISSUE #140: Clarify m= section for "offerToReceive"
For the case where the offer should contain m-lines dedicated for incoming media (reconly lines), section 5.2.3.1 and 5.2.3.1 says nothing about what should be in those m= sections. Should be clarified.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1403,7 +1403,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           MediaStreamTrack is added to a PeerConnection, the resultant m=audio
           section will precede the m=video section. If a second MediaStream
           containing an audio MediaStreamTrack was added, it would follow
-          the m=video section.</t>
+          the m=video section. If any extra m= sections are required because
+          OfferToReceiveAudio (<xref target="sec.offertoreceiveaudio"/>
```
fluffy: I think it would make more sense to change 
   OfferToReceiveAudio were used 
to something like 
  value set for the the OfferToReceiveAudio option. 
And them a similar change for the other places in the PR

fluffy:
have a look at my trivial nit comment and see what you think but either way I'm OK with this

draft-ietf-rtcweb-jsep.xml
ekr: I don't understand this suggestion. Can you provide a rewritten sentence?

ekr:
Consensus from today's call: Park this and let's discuss in Japan.

ekr:
Closing because we need to remove OfferToReceive*

PR #180, opened 2015-10-16
Apply LD/RD

### LINKED ISSUE #44: Define "most recent gathering phase"
The text relies on the phrase "most recent gathering phase" to describe which candidates are present in SDP.

Justin enumerated a few cases where the browser gathers a completely new set of candidates for an m= section (or component thereof):
- when a new m= section is added (which happens for a new RTCPeerConnection obviously)
- when an m= section did have a port of 0, but is repurposed for new media (basically the same scenario as above)
- ICE restart

This needs a clearer definition, or maybe just a single clear description of what it means to do this.  This can probably tie into the whole process of describing when candidate events fire and the interaction between that and setLocalDescription.

### LINKED ISSUE #54: Add sections on parsing and interpreting SDP
Outline: https://docs.google.com/a/google.com/document/d/1XzhL55Z9PibChNz7nUwE7p0tJYhItTTw0USRUirtTMg/edit

juberti:
Fixes #54

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -744,7 +744,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             that the offerer needs to gather.</t>
           </list>
           </t>
-          
+
         </section>
         <section title="createOffer" anchor="sec.createoffer">
           <t>The createOffer method generates a blob of SDP that contains a
@@ -1077,7 +1077,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>The BUNDLE and RTCP-multiplexing policies MUST NOT be
             changed after the construction of the PeerConnection.
             </t>
-            
+
           </list></t>
           <t>This call may result in a change to the state of the ICE Agent,
           and may result in a change to media state if it results in
@@ -2315,7 +2315,27 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         bounds, or cannot be applied, processing MUST stop and an error MUST be
         returned.
           <list style="symbols">
-            <t>TODO</t>
+            <t>If the ICE ufrag and password values changed, trigger the ICE Agent
+              to start an ICE restart and begin gathering new candidates, as defined
+              in <xref target="RFC5245"/>, Section X.Y.</t>
+          </list>
+        </t>
+        <t>If the media section proto value indicates use of RTP:
+          <list style="symbols">
+            <t>If the RTCP mux parameter is set, discard any RTCP transport,
+             and begin muxing RTCP over the RTP transport, as specified in
+             <xref target="RFC5761"/>, Section X.Y. [if answer]
+            </t>
+            <t>Set RTP header extension receive state, ensure all are known</t>
+            <t>Set PT demux</t>
+            <t>Set receive codecs, ensure all are known</t>
+            <t>Set up RTX (verify apt)</t>
+            <t>Set local directional state</t>
+          </list>
+        </t>
+        <t>If the media section proto value indicates use of SCTP:
+          <list style="symbols">
+            <t>Set local SCTP pseudo-port</t>
           </list>
         </t>
         <t>Finally, if this description is of type "pranswer" or "answer", follow
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2290,14 +2290,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     <t>DTLS setup value, which MUST be set according to the rules
                       specified in <xref target="RFC5763"/>, Section 5, and MUST
                       be consistent with the selected role of the current DTLS
-                      connection, if one exists.[TODO: may need revision, i.e. use of actpass</t>
+                      connection, if one exists.[TODO: may need revision, i.e., use of actpass</t>
                     <t>DTLS fingerprint values, where at least one fingerprint
                       MUST be present.</t>
                   </list>
                 </t>
                 <t>Each m= section is also checked to ensure prohibited features
-                  are not used. If this is a local description, the ICE Lite
-                  setting MUST NOT be specified.</t>
+                  are not used. If this is a local description, the "ice-lite"
+                  attribute MUST NOT be specified.</t>
               </list>
           </t>
           <t>If this session description is of type "pranswer" or "answer", the
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2315,27 +2315,42 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         bounds, or cannot be applied, processing MUST stop and an error MUST be
         returned.
           <list style="symbols">
-            <t>If the ICE ufrag and password values changed, trigger the ICE Agent
-              to start an ICE restart and begin gathering new candidates, as defined
-              in <xref target="RFC5245"/>, Section X.Y.</t>
-          </list>
-        </t>
-        <t>If the media section proto value indicates use of RTP:
-          <list style="symbols">
-            <t>If the RTCP mux parameter is set, discard any RTCP transport,
-             and begin muxing RTCP over the RTP transport, as specified in
-             <xref target="RFC5761"/>, Section X.Y. [if answer]
+            <t>Basic sanity checking is applied for the following attributes in the media section:
+              <list style="symbols">
+                <t>ICE lite: MUST NOT be present.</t>
+                <t>ICE ufrag: MUST be present, and MUST meet the length criteria from [TODO: ref]</t>
+                <t>ICE pwd: MUST be present, and MUST meet the length criteria from [TODO: ref]</t>
+                <t>DTLS fingerprint: MUST be present, and, for existing media sections, MUST be unchanged.</t>
+                <t>DTLS setup: for new media sections, MUST follow the guidance from [TODO: ref]; for existing media sections, MUST be unchanged. [TODO: check whether this is what we agreed on (vs actpass)]</t>
+              </list>
             </t>
-            <t>Set RTP header extension receive state, ensure all are known</t>
-            <t>Set PT demux</t>
-            <t>Set receive codecs, ensure all are known</t>
-            <t>Set up RTX (verify apt)</t>
-            <t>Set local directional state</t>
-          </list>
-        </t>
-        <t>If the media section proto value indicates use of SCTP:
-          <list style="symbols">
-            <t>Set local SCTP pseudo-port</t>
+            <t>If this media section is new, begin gathering candidates for it,
+             as defined in <xref target="RFC5245"/>, Section X.Y.,
+             unless it has been marked as bundle-only.</t>
+            <t>Or, if the ICE ufrag and password values have changed, trigger the ICE Agent
+              to start an ICE restart and begin gathering new candidates for the
+              media section, as defined in <xref target="RFC5245"/>, Section X.Y.,
+              unless it has been marked as bundle-only.</t>
+            <t>If the media section proto value indicates use of RTP:
+              <list style="symbols">
+                <t>If RTCP mux is indicated, prepare to demux RTP and RTCP from the RTP
+                  ICE component. [TODO: ref]</t>
+                <t>For each specified RTP header extension, establish a mapping between
+                  the extension ID and URI. If any RTP header extension is unknown, this
+                  MUST result in an error. [TODO: ref]</t>
+                <t>Prepare to demux RTP data intended for this media section based on
+                  the MID header extension. [TODO: ref]</t>
+                <t>For each specified payload type, establish a mapping between the
+                  payload type ID and the actual media format. [TODO: ref] If any indicated payload
+                  type is unknown, this MUST result in an error.</t>
+                <t>For each specified "rtx" media format, establish a mapping between
+                  the RTX payload type and its associated primary payload type. [TODO: ref] If any
+                  referenced primary payload types are not present, this MUST result in an error.</t>
+                <t>If the directional attribute is of type "sendrecv" or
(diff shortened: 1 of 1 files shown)


juberti:
Still have a bunch of stuff to flesh out, but wanted to get this out for initial review. @ekr @fluffy

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2315,7 +2315,42 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         bounds, or cannot be applied, processing MUST stop and an error MUST be
         returned.
           <list style="symbols">
-            <t>TODO</t>
+            <t>Basic sanity checking is applied for the following attributes in the media section:
+              <list style="symbols">
+                <t>ICE lite: MUST NOT be present.</t>
+                <t>ICE ufrag: MUST be present, and MUST meet the length criteria from [TODO: ref]</t>
+                <t>ICE pwd: MUST be present, and MUST meet the length criteria from [TODO: ref]</t>
+                <t>DTLS fingerprint: MUST be present, and, for existing media sections, MUST be unchanged.</t>
```
fluffy: If a new fingerprint is added, the browser would need a way of finding the cert and private key for that fingerprint so unless we want to get into all that, we might want to just say it has to be one of the fingerprints that was in the original.

fluffy:
Looks like good stuff to me. My only concerns is I don't want this to be taken as something that overrides what the normatively referenced specs say to do - we might already have text that covers that when read in full context but I did not got look. This seems like a reasonable level of detail that it helps folks while at the same time being concise enough one can read it.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2315,7 +2315,42 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                <t>DTLS setup: for new media sections, MUST follow the guidance from [TODO: ref]; for existing media sections, MUST be unchanged. [TODO: check whether this is what we agreed on (vs actpass)]</t>
+              </list>
```
ekr: What happens if this sanity checking fails? You reject with an error.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2315,7 +2315,42 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            </t>
+            <t>If this media section is new, begin gathering candidates for it,
+             as defined in <xref target="RFC5245"/>, Section X.Y.,
+             unless it has been marked as bundle-only.</t>
+            <t>Or, if the ICE ufrag and password values have changed, trigger the ICE Agent
+              to start an ICE restart and begin gathering new candidates for the
+              media section, as defined in <xref target="RFC5245"/>, Section X.Y.,
+              unless it has been marked as bundle-only.</t>
+            <t>If the media section proto value indicates use of RTP:
```
ekr: This conditional seems weird, because you could have the ufrag/password change *and* have a session-level ICE ufrag/password change. So maybe the opposite order?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2315,7 +2315,42 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              <list style="symbols">
+                <t>If RTCP mux is indicated, prepare to demux RTP and RTCP from the RTP
+                  ICE component. [TODO: ref]</t>
```
ekr: I thought we had sort of agreed that JSEP endpoints had to offer mux: the only legal policies always offer mux and as of PR#178 you can't change it at the app layer. So I think that this does not need a conditional.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2315,7 +2315,42 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                <t>For each specified RTP header extension, establish a mapping between
+                  the extension ID and URI. If any RTP header extension is unknown, this
+                  MUST result in an error. [TODO: ref]</t>
```
ekr: This seems a big vague. What does "establish a mapping" mean in practice.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2315,7 +2315,42 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                <t>Prepare to demux RTP data intended for this media section based on
+                  the MID header extension. [TODO: ref]</t>
```
ekr: Is MID required?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2315,7 +2315,42 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            <t>Or, if the ICE ufrag and password values have changed, trigger the ICE Agent
+              to start an ICE restart and begin gathering new candidates for the
+              media section, as defined in <xref target="RFC5245"/>, Section X.Y.,
+              unless it has been marked as bundle-only.</t>
+            <t>If the media section proto value indicates use of RTP:
+              <list style="symbols">
+                <t>If RTCP mux is indicated, prepare to demux RTP and RTCP from the RTP
+                  ICE component. [TODO: ref]</t>
+                <t>For each specified RTP header extension, establish a mapping between
+                  the extension ID and URI. If any RTP header extension is unknown, this
```
ekr: "any indicated" as below.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2315,7 +2315,42 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                <t>Prepare to demux RTP data intended for this media section based on
+                  the MID header extension. [TODO: ref]</t>
+                <t>For each specified payload type, establish a mapping between the
+                  payload type ID and the actual media format. [TODO: ref] If any indicated payload
+                  type is unknown, this MUST result in an error.</t>
+                <t>For each specified "rtx" media format, establish a mapping between
+                  the RTX payload type and its associated primary payload type. [TODO: ref] If any
+                  referenced primary payload types are not present, this MUST result in an error.</t>
+                <t>If the directional attribute is of type "sendrecv" or
+                  "recvonly", enable receipt, decoding, and playout of media.
```
ekr: This seems like it's assuming too much: i.e., we wouldn't receive prior to DTLS and we don't playout until it's attached to something. Trying to figure out the best way to put this.

Perhaps:
"Configure the media decoding subsystem to receive media with the offered parameters"?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2324,11 +2359,107 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        bounds, or cannot be applied, processing MUST stop and an error MUST be
+        returned.
+          <list style="symbols">
+            <t>Basic sanity checking is applied for the following attributes in the media section:
+              <list style="symbols">
+                <t>ICE ufrag: MUST be present, and MUST meet the length criteria from [TODO: ref]</t>
+                <t>ICE pwd: MUST be present, and MUST meet the length criteria from [TODO: ref]</t>
+                <t>DTLS fingerprint: MUST be present.</t>
+                <t>DTLS setup: for new media sections, MUST follow the guidance from [TODO: ref]; for existing media sections, MUST be unchanged.</t>
+              </list>
```
ekr: This section seems like the thing you could refactor.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2324,11 +2359,107 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                  enable them on the associated payload types.</t>
+                <t>For each specified RTCP feedback mechanism that is supported by the local implementation,
+                  enable them on the associated payload types.</t>
+                <t>For any specified "b=AS" or "b=TIAS" bandwidth value, record this value
+                  as the maximum value that the remote side wishes to receive for this media section.
+                  [TODO: convert AS to TIAS].</t>
+                <t>[TODO: handling of CN, telephone-event, "red"]</t>
+                <t>If the media section if of type audio:
+                  <list style="symbols">
+                    <t>For any specified "ptime" value, configure the avaialble payload types to
```
ekr: spelling on available.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2324,11 +2359,107 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                   <xref target="RFC5761"/>, Section X.Y. Otherwise, transmit RTCP over
+                   the RTCP component; if no RTCP component exists, because RTCP mux
+                   was previously enabled, this MUST result in an error. [TODO: ref]</t>
+                <t>If the media section has reduced-size RTCP enabled, configure the
+                  RTCP transmission for this media section to use reduced-size RTCP,
+                  as specified in [TODO: ref]</t>
+                <t>If the directional attribute in the answer is of type "sendrecv" or
+                  "sendonly", start transmitting media using the specified primary SSRC
+                  and one of the selected payload types, once the underlying transport
+                  layers have been established. Otherwise, stop transmitting RTP media,
```
ekr: Maybe "prepare to start transmitting", since it's like "start... but only when ICE is done"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2324,11 +2359,107 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            <t>If the media section proto value indicates use of RTP:
+              <list style="symbols">
+                <t>If the media section has RTCP mux enabled, discard any RTCP component,
+                   and begin or continue muxing RTCP over the RTP component, as specified in
+                   <xref target="RFC5761"/>, Section X.Y. Otherwise, transmit RTCP over
+                   the RTCP component; if no RTCP component exists, because RTCP mux
+                   was previously enabled, this MUST result in an error. [TODO: ref]</t>
+                <t>If the media section has reduced-size RTCP enabled, configure the
+                  RTCP transmission for this media section to use reduced-size RTCP,
+                  as specified in [TODO: ref]</t>
```
ekr: TODO: you should be sad if there is an answer with non-matching PTs.

juberti:
Thanks for the comments. Ran out of time to resolve them today but will take another shot at it Saturday.

draft-ietf-rtcweb-jsep.xml
juberti: Yes; I think that is what I am saying.

draft-ietf-rtcweb-jsep.xml
juberti: Didn't quite follow what you were saying here.

draft-ietf-rtcweb-jsep.xml
juberti: If the other side said no to mux, and you did a reoffer, you could end up not offering mux (I think).

draft-ietf-rtcweb-jsep.xml
juberti: Agreed. Will resolve in next PR.

draft-ietf-rtcweb-jsep.xml
juberti: Fixed.

draft-ietf-rtcweb-jsep.xml
juberti: Fixed.

draft-ietf-rtcweb-jsep.xml
juberti: I think it's more than just decoding; it's the ability to receive any RTP media at all.

I couldn't think of anything better right now, so I just removed the mention of playout.

draft-ietf-rtcweb-jsep.xml
juberti: Fixed.

draft-ietf-rtcweb-jsep.xml
juberti: Fixed.

draft-ietf-rtcweb-jsep.xml
juberti: Fixed.

draft-ietf-rtcweb-jsep.xml
juberti: This is covered in the semantics validation section above, but it was buried a bit, so I expanded it to make it more obvious.

juberti:
Updated based on comments. I'd like to go ahead and merge this ASAP, and then hopefully still get another rev in before the deadline.

draft-ietf-rtcweb-jsep.xml
ekr: Hmm...  i guess that's true. Probably needs to be written down somewhere.

draft-ietf-rtcweb-jsep.xml
ekr: The logic here is:

if (media section new) {
   begin gathering
} else (if ICE restart) {
   trigger ice restart and begin gathering
}

but what if (media section new && ice restart)? This text suggests that you would just begin gathering.

draft-ietf-rtcweb-jsep.xml
fluffy: Yah, I think we should fill out the nuances of this

fluffy:
I'm fine with merging what we have here right now.

draft-ietf-rtcweb-jsep.xml
juberti: For this media section, I think you would just begin gathering, as there's nothing to restart.

draft-ietf-rtcweb-jsep.xml
juberti: Some of this is covered in https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-10#section-5.2.2, namely only offer rtcp-mux if we previously agreed to use it.

However, it should probably be an error to stop using RTCP mux if we were previously using it. I can add text to that effect.

juberti: Added text.

PR #183, opened 2015-11-02
Set default mux policy to require

ekr:


martinthomson:
LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -837,7 +837,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             not provide an "a=rtcp-mux" attribute.
             </t>
           </list>
-          </t>
+          <t>The default multiplexing policy MUST be set to "require".</t>
```
pthatcherg: +1

The default should have been specified all along, and I think this is the right default value.

PR #184, opened 2015-11-03
Fix editorial mess left by IETF 94 mux discussion

martinthomson:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -837,9 +837,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             not provide an "a=rtcp-mux" attribute.
             </t>
           </list>
+          </t>
+          <t>
           The default multiplexing policy MUST be set to
           "require". Implementations MAY choose to reject attempts by the
-          application to set the multiplexing policy to "negotiate"
+          application to set the multiplexing policy to "negotiate".
```
ekr: If you think this is important, please also fix the <t></t> bracing on lines 820-823 above.

PR #189, opened 2015-11-03
Fix ice-passwd in ice-pwd

dontcallmedom:


ekr:
This change is correct.

martinthomson:
Yes it is.

PR #195, opened 2015-12-15
Using template

martinthomson:
Should work better than the TLS change.


juberti:
Discussed and plan is to get this debugged for TLS first and then we can revisit it here.

ekr:
Fixed separately.

PR #200, opened 2016-01-08
Clear up ICE restart logic

### LINKED ISSUE #179: Need clarification on ICE restart logic
From deadbeef@google.com:
_Question 1: Should this be interpreted literally, to mean "the next call, and only the next call to createOffer will generate new ICE credentials"? If so, there are various situations wherein one could encounter unexpected behavior. For example:
Before createOffer is called, a remote offer is received that triggers an ICE restart. Should the next call to createOffer trigger another, superfluous ICE restart?
createOffer is called twice, without setting the local description (and kicking off a new gathering phase). Should the second call generate new credentials, or only the first?
I would propose a slight change, such as: "this will cause subsequent calls to createOffer to generate new ICE credentials, until a new gathering phase has started." This would more clearly represent the intent of the original rule, in my opinion._

_Question 2: The next paragraph states "Any changes to the ICE candidate policy also affect the next gathering phase, in similar fashion to the server changes described above." Does this mean that changes to the candidate policy will also cause an ICE restart, according to the same rules? I assume so, but figure it's worth asking._

juberti:
Fixes #179

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -421,7 +421,7 @@ setRemote(ANSWER) |                                   |
           gathering phase, a 'needs-ice-restart' bit is set. When this bit is
           set, calls to the createOffer API will generate new ICE credentials.
           This bit is cleared by a call to the setLocalDescription API with
-          new ICE credentials from either an offer or an answer, i.e. from either
+          new ICE credentials from either an offer or an answer, i.e., from either
           a local- or remote-initiated ICE restart.</t>
           <t>When a new
           gathering phase starts, the ICE Agent will notify the application that
@@ -1164,8 +1164,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           parameters are changed:</t>
          <t><list style="symbols">
             <t>Any changes to the STUN/TURN servers to use affect
-            the next gathering phase. If gathering has already occurred, the
-            'needs-ice-restart' bit mentioned in <xref target="sec.ice-gather-overview"/>
+            the next gathering phase. With the exception of the initial state
+            where ICE gathering has not yet occurred, the 'needs-ice-restart'
+            bit mentioned in <xref target="sec.ice-gather-overview"/>
             will be set. This will cause the next call to createOffer to
             generate new ICE credentials, for the purpose of forcing an ICE
             restart and kicking off a new gathering phase, in which the new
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -417,6 +417,12 @@ setRemote(ANSWER) |                                   |
           be triggered explicitly by the application, or implicitly by the
           browser in response to changes in the ICE configuration.
           </t>
+          <t>When the ICE configuration changes in a way that requires a new
+          gathering phase, a 'needs-ice-restart' bit is set. When this bit is
+          set, calls to the createOffer API will generate new ICE credentials.
+          This bit is cleared by a call to the setLocalDescription API with
+          new ICE credentials from either an offer or an answer, i.e. from either
```
ekr: nit: i.e.,

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1158,20 +1164,23 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            any existing candidates will be discarded, and new candidates will
+            be gathered from the new servers.
             </t>
             <t>Any changes to the ICE candidate policy also affect the next
-            gathering phase, in similar fashion to the server changes described
-            above. Note though that changes to the policy have no effect on
-            the candidate pool, because pooled candidates are not surfaced to
-            the application until a gathering phase occurs, and so any
-            necessary filtering can still be done on any pooled candidates.
+            gathering phase. If the policy is being made more liberal (i.e.
```
ekr: nit: i.e.,

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1158,20 +1164,23 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
             <t>Any changes to the STUN/TURN servers to use affect
-            the next gathering phase. If gathering has already occurred, this
-            will cause the next call to createOffer to generate new ICE
-            credentials, for the purpose of forcing an ICE restart and kicking
-            off a new gathering phase, in which the new servers will be used.
-            If the ICE candidate pool has a nonzero size, any existing
-            candidates will be discarded, and new candidates will be gathered
-            from the new servers.
+            the next gathering phase. If gathering has already occurred, the
+            'needs-ice-restart' bit mentioned in <xref target="sec.ice-gather-overview"/>
```
ekr: Maybe
"If gathering is complete, any changes to the STUN/TURN servers affect the next gathering phase..."

and then at the end
"If STUN/TURN servers are changed during gathering, then no ICE restart is needed"

juberti: Note to self - ensure that whatever I write here addresses the following paragraph as well.

PR #201, opened 2016-01-08
Add reference to IP handling doc

### LINKED ISSUE #194: Ice Candidate Policy should reference ip-handling draft
https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-12#section-3.4.3 talks about browser local policy. It should make this more concrete by referring to https://tools.ietf.org/html/draft-shieh-rtcweb-ip-handling-00 as the source of this policy.

juberti:
Fixes #194

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -497,9 +497,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           location information as possible (note that these choices come with
           corresponding operational costs). To accomplish this, the browser
           MUST allow the application to restrict which ICE candidates are used
-          in a session. In addition, administrators may also wish to control
-          the set of ICE candidates, and so the browser SHOULD also allow
-          control via local policy, with the most restrictive policy prevailing.
+          in a session. Note that this filtering is applied on top of any
+          restrictions the browser chooses to enforce regarding which IP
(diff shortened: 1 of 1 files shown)


fluffy:
Not sure if this is done but what is there LGTM

juberti:
Yes, I think this is complete.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -766,9 +767,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                 even global level.</t>
           </list>
           </t>
-          <t>Although it can be overridden by local policy, the default ICE
-          candidate policy MUST be set to allow all candidates, as this
-          minimizes use of application STUN/TURN server resources.</t>
+          <t>The default ICE candidate policy MUST be set to allow all
+          candidates, as this is generally the desired policy, and also
```
ekr: suggest "all types of candidates" because if you apply the draft-shieh rules, then it's not all.

ekr:
LGTM

draft-ietf-rtcweb-jsep.xml
juberti: We use "all candidates" elsewhere, so I think this tweak would be insufficient. But perhaps I could speak to the exact meaning of "all candidates" on line 754, and allude to the just-mentioned browser filtering.

PR #202, opened 2016-01-12
clarify VAD stuff

fluffy:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1859,6 +1859,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             This option allows the endpoint to significantly reduce the amount of
             audio bandwidth it receives, at the cost of some fidelity, depending on
             the quality of the remote VAD algorithm.</t>
+            <t>If the "VoiceActivityDetection" option is specified, with a value
+            of "false", the browser MUST negotiate the audio to not suppress any
+            audio. For opus this would use an "usedtx=0" parameter, while for
+            G.729 DTX would be not be used, and ("CN") would not be
+            negotiated. </t>
+            <t> The "VoiceActivityDetection" option does not have any impact on
+            the setting of the vad value in the client to mixer audio level
```
juberti: suggest quoting "vad"

Perhaps use the word signaling here, e.g.

"the setting of the "vad" value in the signaling of the client to mixer audio level header extension..."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1859,6 +1859,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             This option allows the endpoint to significantly reduce the amount of
             audio bandwidth it receives, at the cost of some fidelity, depending on
             the quality of the remote VAD algorithm.</t>
+            <t>If the "VoiceActivityDetection" option is specified, with a value
+            of "false", the browser MUST negotiate the audio to not suppress any
```
juberti: Would prefer this to mirror the section above, e.g.

"the offer MUST indicate that it does not want to use silence suppression, by not including any comfort noise codecs and by setting any codec-specific settings appropriately, e.g. usedtx=0 for Opus."

juberti:
We reviewed the handling for G.729 and it's similar to Opus; there's a dtx=foo fmtp parameter to control receipt of DTX.

I suggest we leave this parameter to createOffer as an indication on whether you want the remote side to use DTX, and add a frob on RtpSender (TBD exactly where) to directly control whether you use DTX. That frob is out of scope for this PR.

For the PSAP/E911 case, the call will typically be over PCMU, and so it's covered through either of a) setting this parameter (which drops CN from the offered codecs), or b) the PSAP not responding with CN in the answer. So that's not an issue.

We may need to add some text in the handling-a-remote-description section to make it clear that if the remote side says to use DTX (either through presence of CN, or usedtx=1), DTX MUST be engaged by the browser. We could also spell out the no-DTX case, if desired.

fluffy:
So plan is make this just about SDP, both offer and answer, along lines of Justin text. Then  file issue to deal with happens with RTP sender.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1859,6 +1859,20 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
             audio bandwidth it receives, at the cost of some fidelity, depending on
             the quality of the remote VAD algorithm.</t>
+            <t>If the "VoiceActivityDetection" option is specified, with a value
+            of "false", the browser MUST NOT omit "CN" codecs.
+            For codecs that have their own internal silence suppression support,
+            the appropriate fmtp parameters for that codec MUST be
+            specified to indicate that silence suppression for received audio is
+            not desired. For example, when using the Opus codec, the
+            "usedtx=0" parameter would be specified in the offer.</t>
+            <t>Note that setting the "VoiceActivityDetection" parameter on an
```
juberti: suggest "... parameter when generating an offer is a request to receive audio with silence suppression. It has no impact on whether the local endpoint does silence suppression for the audio it sends."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1859,6 +1859,20 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            endpoint has no impact on whether that endpoint does silence
+            suppression for the audio it sends.</t>
+            <t> The "VoiceActivityDetection" option does not have any impact on
```
juberti: extra space before 'The'

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2186,8 +2200,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <section title="VoiceActivityDetection" anchor="sec.voiceactivitydetection2">
             <t>Silence suppression in the answer is handled as described in
-              <xref target="sec.voiceactivitydetection1"></xref>.
-            </t>
+            <xref target="sec.voiceactivitydetection1"></xref>.</t>
```
juberti: Since you can't answer with CN if it wasn't offered, we probably need a note here to the effect of:
"If silence suppression was not indicated in the offer, via the omission of CN codecs or relevant fmtp parameters, the VoiceActivityDetection parameter has no effect, and the answer should be generated as if VoiceActivityDetection was set to false."

fluffy:
Put in all Justin's comments from prev review.

PR #204, opened 2016-01-26
a=imageattr clarification - Issue157

suhasHere:
Adds clarification to the imageattr example

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -603,10 +603,23 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             type.
           </t>
           <t>As an example, consider a system with a HD-capable, multiformat
-          video decoder, where the application has constrained the received
-          track to at most 360p. In this case, the implemention would generate this
-          attribute:</t>
+            video decoder, where the application has constrained the received
+            track to at most 360p. In this case, the implemention would generate this
+            attribute:</t>
           <t>a=imageattr:* recv [x=[16:640],y=[16:360],q=1.0]</t>
+          <t>Also to note, the above example implies that the receiver is capable 
+            of decoding all the image screen resolutions ranging from 16x16 
+            pixels up to 640x360 pixels.</t>
+          <t>In scenarios where the receiver would want to constrain the possible
+            image screen resolutions, it does so by including the 'step' and/or the
+            'par' imageattr parameters as shown:</t> 
+          <t>a=imageattr:* recv [x=[16:16:640],y=[16:8:360],q=1.0, par=1.2-1.8]</t> 
+          <t>Here, the x-axis resolution can take values from 16 up to 640 in 16 
+            pixels steps. The y-axis resolution similarly can take values from 16 
+            up to 360 in 8 pixels step. The par parameter limits the set of possible 
+            x and y screen resolution combinations such that 640x360 (ratio=1.777) is 
+            a valid combination while 640x16 (ratio=40) or 160x320 (ratio=0.5) are 
+            invalid combinations.</t>
         </section>
 
         <section title="Interpreting an imageattr Attribute">


draft-ietf-rtcweb-jsep.xml
```diff
@@ -603,10 +603,23 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           <t>As an example, consider a system with a HD-capable, multiformat
-          video decoder, where the application has constrained the received
-          track to at most 360p. In this case, the implemention would generate this
-          attribute:</t>
+            video decoder, where the application has constrained the received
+            track to at most 360p. In this case, the implemention would generate this
+            attribute:</t>
           <t>a=imageattr:* recv [x=[16:640],y=[16:360],q=1.0]</t>
+          <t>Also to note, the above example implies that the receiver is capable 
+            of decoding all the image screen resolutions ranging from 16x16
```
juberti: I think the term 'screen' is confusing. Perhaps just 'image resolution'?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -603,10 +603,23 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          video decoder, where the application has constrained the received
-          track to at most 360p. In this case, the implemention would generate this
-          attribute:</t>
+            video decoder, where the application has constrained the received
+            track to at most 360p. In this case, the implemention would generate this
+            attribute:</t>
           <t>a=imageattr:* recv [x=[16:640],y=[16:360],q=1.0]</t>
+          <t>Also to note, the above example implies that the receiver is capable 
+            of decoding all the image screen resolutions ranging from 16x16 
+            pixels up to 640x360 pixels.</t>
```
juberti: Perhaps: "This declaration indicates that the receiver is capable of decoding any image resolution from 16x16 up to 640x360 pixels."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -603,10 +603,23 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          <t>In scenarios where the receiver would want to constrain the possible
```
juberti: The paragraph above mentions how a receiver might want to constrain the possible resolutions, so this sentence feels out of place.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -603,10 +603,23 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            image screen resolutions, it does so by including the 'step' and/or the
+            'par' imageattr parameters as shown:</t> 
+          <t>a=imageattr:* recv [x=[16:16:640],y=[16:8:360],q=1.0, par=1.2-1.8]</t>
```
juberti: Is par= necessary to express decoder limits?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -603,10 +603,23 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           <t>a=imageattr:* recv [x=[16:640],y=[16:360],q=1.0]</t>
+          <t>Also to note, the above example implies that the receiver is capable 
+            of decoding all the image screen resolutions ranging from 16x16 
+            pixels up to 640x360 pixels.</t>
+          <t>In scenarios where the receiver would want to constrain the possible
+            image screen resolutions, it does so by including the 'step' and/or the
+            'par' imageattr parameters as shown:</t> 
+          <t>a=imageattr:* recv [x=[16:16:640],y=[16:8:360],q=1.0, par=1.2-1.8]</t> 
+          <t>Here, the x-axis resolution can take values from 16 up to 640 in 16 
+            pixels steps. The y-axis resolution similarly can take values from 16
```
juberti: We haven't specified how a step value should be handled. If the generated image doesn't match the allowed steps, should the sender pad, scale, crop, ?

Unless we have a specific use case in mind here, I think we should omit this.

draft-ietf-rtcweb-jsep.xml
suhasHere: This looks like a much better sentence construction.

draft-ietf-rtcweb-jsep.xml
suhasHere: My understanding it par attribute ensures that only a subset of possible resolutions be proposed and this picked.  It also provides a nicer way to avoid combinations like 17x191 pixels or 640x16 or combinations that might end up in unpleasant user experience if negotiated.

draft-ietf-rtcweb-jsep.xml
suhasHere: I tend to agree with you needing a concrete requirement to handle the mismatch between steps and the generated image resolution. One motivating factor that i thought as applicable for the reasons what par attribute does - reduce the possible options in a given range to a more manageable subset and added benefit of moving uniformly in the range (in terms of step pixels).

Also to note, omitting the step value implies a unit step value.

I am open to options here.

draft-ietf-rtcweb-jsep.xml
juberti: Right, but how would a browser fill this in? There's no API to express this.

juberti: We may need to say what happens when PAR limits are exceeded - probably we just want to fail in that case - but I don't think this should be part of the initial example.

draft-ietf-rtcweb-jsep.xml
juberti: Again, how would the browser fill this in? (in SDP)

Also, a unit step value is not an issue since the browser can simply send what it has. Any other step value implies the need to possibly process the image before sending. I would prefer to say that WebRTC endpoints are not required to support anything other than unit step.

juberti:
suhas, further thoughts on this?

juberti:
ping...

suhasHere:
Justin .. sorry for the delay in responding . Here are some thoughts on various aspects of image attributes and its relation to Constrains api. These may be misplaced since I am thinking loud at this point and would love to hear your thoughts

**sar attribute:**
We might safely ignore it and assume the world has moved to square pixels

**par attribute:**
RFC6236 says  " The par (width/height = x/y ratio) parameter indicates a range of allowed ratios between x and y physical size (picture aspect ratio).  This is used to limit the number of x and y image size combinations; par is given as par=[ratio_min-ratio_max]"

**Example: a=imageattr:* recv [x=[16:640],y=[16:360],q=1.0, par=1.2-1.8]**

We might represent it as different between min and max aspect ratios if provided.
var constraints = {
 aspectRatio: {min:1.2, max:1.8} // map to par
 width: {min:16, max:640}
 height:{min:16, max:360}
}

**step attribute:**
Unit Step seems fine. Do you know of any scenarios where the receiver shall not be able to decode/display the unit increments say 171 X 191 ?

** q attribute: **
q mentions the preference between various resolutions offered. 
I couldn't think of a easy way to map constraints to something like below, where we have 2 x,y ranges but assign preference to each.
_a=imageattr:99 \
      send [x=480:800],y=[320:640],q=0.8] \
      recv [x=176:208], [y=144:176],q=0.2]_

juberti:
obsoleted by #224

PR #208, opened 2016-02-05
Splitting local/remote descriptions into "current" and "pending".

taylor-b:
I left out the unqualified `localDescription`/`remoteDescription`, since these are redundant, and only exist in the WebRTC spec as a convenience and for backwards compatibility.


fluffy:
I need to read this carefully at some point but just quickly scanning the diff .. it looks workable. Thanks.

juberti:
LGTM. I didn't look to see if there were other references to the local/remote desc that need to be updated, but I assume that you already scanned carefully.

taylor-b:
I think everything should be covered now. I've gone through twice looking at every place the word "description" is used.

Though when talking about rollback, I think "return to the previous description" sounds less confusing than "return to the current description", so I left that wording alone.

fluffy:
LGTM

PR #216, opened 2016-03-03
Remove 'public' ICE policy. Fixes #203

### LINKED ISSUE #203: Reconsider "public"and "none" ICE candidate policies
From https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-12#section-4.1.1

"public" is probably something the browser should control, not the application.
"none" was previously removed, but apparently Firefox has implemented it.

ekr:


fluffy:
LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -759,13 +759,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           <list style="hanging">
-                <t hangText="all:">All candidates will be gathered and used.</t>
-                <t></t>
-                <t hangText="public:">Candidates with private IP addresses
-                [RFC1918] will be filtered out. This prevents exposure of
-                internal network details, at the cost of requiring
-                relay usage even for intranet calls, if the NAT does not allow
-                hairpinning as described in [RFC4787], section 6.</t>
+                <t hangText="all:">All candidates permitted by browser policy
+                (See <xref target="sec.ice-candidate-policy"/>) will be gathered and used.</t>
```
juberti: So we just mentioned this xref in the preceding paragraph; I don't think we need it here.

I suggest that in the preceding paragraph, we simply mention that the browser may be doing its own filtering, e.g.:

"If an ICE candidate policy is specified, it functions as described in <xref target="sec.ice-candidate-policy"/>, causing the browser to only surface the permitted candidates (including any internal browser filtering) to the application..."

juberti:
lgtm

Issue #196, opened 2015-12-18
Use RID for signaling interactions between primary and RTX/FEC flows


juberti:
cc @pthatcherg

pthatcherg:
If the header extension is used, nothing needs to be signalled to correlate source RTP streams with redundancy streams, not even the RID value itself.

juberti:
Yes. We just need to know what to put in the SDP for the createoffer/answer sections.

hardie:
Suggest contacting Christer, Magnus, or Bernard for input.

juberti:
@pthatcherg do you know what the RID SDP needs to look like?

suhasHere:
here is the ABNF we have rid : https://tools.ietf.org/html/draft-ietf-mmusic-rid-00#section-10

pthatcherg:
The result of createOffer should be:

a=rid:A send
a=rid:B send
a=rid:C send 
a=simulcast: send rid=A; B; C

When given an RtpEncodingParmeter like this:

pc.addTransceiver(videoTrack, {
  sendEncodings: [{
    rid: "A",
  }, sendEncodings: [{
    rid: "B",
  }, sendEncodings: [{
    rid: "C",
  }
 ]
});

And if there is only one encoding, we can leave off the a=simulcast line.

In other words, iterate over the encodings of the RtpSender.  If the rid is set, output an a=rid line with the rid value and "send".  If there is more than one encoding, output an a=simulcast line.  

If the browser chooses to support allowing the remote answer to constrain the rid-line via one of the parameters, then it can add that parameter to the rid line.  For example, to support constraining max FPS.

a=rid:A send max-fps
a=rid:B send max-fps
a=rid:C send max-fps
a=simulcast: send rid=A; B; C

But all of those are optional for the browser to support, as far as I know.

createAnswer is very similar to createOffer, except that it must only include rid lines where the rid value already exists in the offer, and skip those that don't.  That's a bit burdensome for the application to know what RID values were in the offer, but not overly so.  


As for receiving simulcast: I think that's out of scope for 1.0, so we don't need to output a=rid lines with "recv".

fluffy:
This looks close to done in PR

juberti:
I think this was resolved in #219.

PR #219, opened 2016-03-03
Add support for RIDs and simulcast

pthatcherg:
Here's a first attempt at https://github.com/rtcweb-wg/jsep/issues/196


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1595,7 +1595,27 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	      target="I-D.ietf-avtext-rid"></xref>.  The direction MUST be
+	      "send".  If the application has choosen a RID value, it MUST be
+	      used as the rid-identifier, otherwise a RID value MUST be chosen
+	      by the implemenatation.  Supported constraints MAY be included.
+	      </t>
+
+	      <t>If the RtpTransceiver's RtpSender is active and more than one
+	      "a=rid" line has been generated, an "a=simulcast" line, with
+	      direction "send" an d using the RID identification mechanism as
+	      defined in <xref target="I-D.ietf-mmusic-simulcsat">.  The list of
```
fluffy: you are missing </xref> here

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1720,6 +1740,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>For MediaStreamTracks that are still present, the "a=msid",
               "a=ssrc", and "a=ssrc-group" lines MUST stay the same.</t>
 
+	      <t>For RtpTransceivers that are still present, the "a=rid" lines
+	      MUST follow the rules for subsequent offers defined in <xref
+	      target="I-D.ietf-mmusic-rid"></xref>.</t>
```
juberti: Is there a section number?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1720,6 +1740,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+
+	      <t>For RtpTransceivers that are still present, the "a=simulcast"
+	      lines MUST follow the rules for subsequent offers defined in <xref
+	      target="I-D.ietf-mmusic-simulcast"></xref>.</t>
```
juberti: same here

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2467,6 +2495,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>If present, a single "a=rtcp-mux" line MUST be parsed as specified in
                 <xref target="RFC5761"/>, Section 5.1.1, and its presence or
                 absence flagged and stored.</t>
+              <t>If present, the "a=rid" lines MUST be parsed as specified in
+	      <xref target="I-D.ietf-mmusic-rid"></xref>, and their values stored.</t>
```
juberti: section numbers (or names)

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2668,6 +2701,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                   layers have been established. Otherwise, stop transmitting RTP media,
                   although RTCP should still be sent, as described in 
                   <xref target="RFC3264"/>, Section 5.1. </t>
+		<t>If the offer contained "a=rid" lines, then the "a=rid" lines,
```
juberti: This is the wrong place to handle this. The section above parses the rid lines into their constituent values; here, we mention how to apply them.

(e.g. the text here should say something like "include the RID header extension in the RTP stream , as indicated in [doc], section foo)

draft-ietf-rtcweb-jsep.xml
juberti: It might be worthwhile to summarize the guidance from the other document, as done for other attributes.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1595,7 +1595,27 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               "max-bundle", and this is not the first m= section, or the BUNDLE
               policy is set to "balanced", and this is not the first m= section
               for this media type, an "a=bundle-only" line.</t>
-            </list></t>
+
+	      <t>If the RtpTransceiver's RtpSender is active, an "a=rid" line
+	      for each encoding specified by the application in the RtpSenders's
+	      parameters.  If no encodings have been specified, then one RID
+	      line is generated.  The rid line is specified in <xref
+	      target="I-D.ietf-mmusic-rid"></xref> and indicates the RID value
```
juberti: Do we need two references in the same sentence? Perhaps we want to refer to separate sections?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1595,7 +1595,27 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	      used for sending media, as specified in <xref
+	      target="I-D.ietf-avtext-rid"></xref>.  The direction MUST be
+	      "send".  If the application has choosen a RID value, it MUST be
```
juberti: Can the app choose RID?
choosen->chosen

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1595,7 +1595,27 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	      used as the rid-identifier, otherwise a RID value MUST be chosen
+	      by the implemenatation.  Supported constraints MAY be included.
```
juberti: Not sure what this last sentence means

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1595,7 +1595,27 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               "max-bundle", and this is not the first m= section, or the BUNDLE
               policy is set to "balanced", and this is not the first m= section
               for this media type, an "a=bundle-only" line.</t>
-            </list></t>
+
+	      <t>If the RtpTransceiver's RtpSender is active, an "a=rid" line
```
juberti: Maybe omit the RtpTransceiver ref from this PR?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1595,7 +1595,27 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	      line is generated.  The rid line is specified in <xref
+	      target="I-D.ietf-mmusic-rid"></xref> and indicates the RID value
+	      used for sending media, as specified in <xref
+	      target="I-D.ietf-avtext-rid"></xref>.  The direction MUST be
+	      "send".  If the application has choosen a RID value, it MUST be
+	      used as the rid-identifier, otherwise a RID value MUST be chosen
+	      by the implemenatation.  Supported constraints MAY be included.
+	      </t>
+
+	      <t>If the RtpTransceiver's RtpSender is active and more than one
```
juberti: simulcsat->simulcast

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1595,7 +1595,27 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	      "a=rid" line has been generated, an "a=simulcast" line, with
+	      direction "send" an d using the RID identification mechanism as
```
juberti: I think you need to explain this mechanism; it's not obvious here.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1595,7 +1595,27 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	      defined in <xref target="I-D.ietf-mmusic-sdp-simulcsat">.  The list of
+	      RIDs SHOULD include all of the RID identifiers used in the "a=rid"
```
juberti: Is there any reason for this to be SHOULD vs MUST?

draft-ietf-rtcweb-jsep.xml
pthatcherg: done

draft-ietf-rtcweb-jsep.xml
pthatcherg: After further thought and discussion with Cullen, we decided it made more sense to just not change the RIDs in subsequent offers.

draft-ietf-rtcweb-jsep.xml
pthatcherg: done

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml
pthatcherg: Yes, the app can choose the RID

draft-ietf-rtcweb-jsep.xml
pthatcherg: I removed it.

draft-ietf-rtcweb-jsep.xml
pthatcherg: I'd rather just wait until the transceiver PR has landed to submit this one.  The other way around would be more work.

draft-ietf-rtcweb-jsep.xml
pthatcherg: done

draft-ietf-rtcweb-jsep.xml
pthatcherg: Changed to a MUST

draft-ietf-rtcweb-jsep.xml
pthatcherg: It turns out that I don't even need to specify.  The latest draft only has one way (the RID way).

fluffy:
Have not worked through this in detail but I don't see any problems on initial read

draft-ietf-rtcweb-jsep.xml
juberti: Need to understand what the deal is with .active and when attributes get filled in. I think we should always have a=rid filled in if there are encodings specified for the sender.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1595,7 +1595,25 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-            </list></t>
+
+	      <t>If the RtpTransceiver's RtpSender is active, an "a=rid" line
+	      for each encoding specified by the application in the RtpSenders's
+	      parameters.  If no encodings have been specified, then one RID
+	      line is generated.  The rid line is specified in <xref
+	      target="I-D.ietf-mmusic-rid"></xref>.  The direction MUST be
+	      "send".  If the application has chosen a RID value, it MUST be
+	      used as the rid-identifier, otherwise a RID value MUST be chosen
+	      by the implemenatation.
```
juberti: implementation. Does RID have to be unique?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1595,7 +1595,25 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	      </t>
+
+	      <t>If the RtpTransceiver's RtpSender is active and more than one
```
juberti: See previous discussion on active

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1595,7 +1595,25 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	      "a=rid" line has been generated, an "a=simulcast" line, with
+	      direction "send" as defined in <xref
```
juberti: comma before 'as'

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1720,6 +1738,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>For MediaStreamTracks that are still present, the "a=msid",
               "a=ssrc", and "a=ssrc-group" lines MUST stay the same.</t>
 
+	      <t>For RtpTransceivers that are still present, the
+	      "a=rid" lines MUST stay the same.</t>
+
+	      <t>For RtpTransceivers that are still present, the "a=simulcast"
+	      lines MUST follow the rules for subsequent offers defined in <xref
```
juberti: Is there a short summary of this policy that we could note here?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2467,6 +2492,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>If present, a single "a=rtcp-mux" line MUST be parsed as specified in
                 <xref target="RFC5761"/>, Section 5.1.1, and its presence or
                 absence flagged and stored.</t>
+              <t>If present, the "a=rid" lines MUST be parsed as specified in
```
juberti: Replace "If present, the "a=rid" with "Any "a=rid"..."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2467,6 +2492,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	      <xref target="I-D.ietf-mmusic-rid"></xref>, Section 10, and their values stored.</t>
+              <t>If present, the "a=simulcast" lines MUST be parsed as specified
```
juberti: See above

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2668,6 +2698,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                   layers have been established. Otherwise, stop transmitting RTP media,
                   although RTCP should still be sent, as described in 
                   <xref target="RFC3264"/>, Section 5.1. </t>
+		<t>If RID values are specified, include the RID header
```
juberti: I think we need some discussion in the 'Verification' section above ensuring that all RIDs referenced in simulcast appear as a=rid lines.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2668,6 +2698,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+		extension in the RTP streams, as indicated in <xref
+		target="I-D.ietf-mmusic-rid"></xref>, Section 4)
+		</t>
+		<t>If simulcast is specified, send simulcast as specified in
```
juberti: I think this should be combined into the "prepare to start transmitting media" section above. You're just indicating how many RTP streams should be generated.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2467,6 +2492,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>If present, a single "a=rtcp-mux" line MUST be parsed as specified in
                 <xref target="RFC5761"/>, Section 5.1.1, and its presence or
                 absence flagged and stored.</t>
+              <t>If present, the "a=rid" lines MUST be parsed as specified in
+	      <xref target="I-D.ietf-mmusic-rid"></xref>, Section 10, and their values stored.</t>
+              <t>If present, the "a=simulcast" lines MUST be parsed as specified
+              in <xref target="I-D.ietf-mmusic-sdp-simulcast"></xref>, and their
+              values stored.</t>
```
juberti: Do we want to comment that a=simulcast recv lines are ignored?

draft-ietf-rtcweb-jsep.xml
pthatcherg: A RID must be unique within a m= section.

draft-ietf-rtcweb-jsep.xml
pthatcherg: done

draft-ietf-rtcweb-jsep.xml
pthatcherg: Changed to "stay the same".

draft-ietf-rtcweb-jsep.xml
pthatcherg: Changed to "all a=rid lines"

pthatcherg: Nevermind, went with "anY

draft-ietf-rtcweb-jsep.xml
pthatcherg: done

draft-ietf-rtcweb-jsep.xml
pthatcherg: I moved it up there.

draft-ietf-rtcweb-jsep.xml
pthatcherg: A simulcast line has both a "send" and "recv" portion.  I think you mean to ignore the send portion from a remote description.  I added a comment to that effect.

draft-ietf-rtcweb-jsep.xml
pthatcherg: I think it's redundant with the simulcast draft, but I added a note to that effect in the Verification section.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2765,8 +2796,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
                   layers have been established. Otherwise, stop transmitting RTP media,
-                  although RTCP should still be sent, as described in
-                  <xref target="RFC3264"/>, Section 5.1. </t>
+                  although RTCP should still be sent, as described in 
+                  <xref target="RFC3264"/>, Section 5.1. If RID values are specified, include the RID header
+		  extension in the RTP streams, as indicated in <xref
+		  target="I-D.ietf-mmusic-rid"></xref>, Section 4).  If simulcast is negotiated,
+		  send the number of Source RTP Streams as specified in
+		  <xref target="I-D.ietf-mmusic-sdp-simulcast"></xref>, Section 6.2.2.
+		  The "send" portion of any "a=simulcast" line from a
```
juberti: I think this belongs in the parsing section.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2529,6 +2553,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                 and its presence or absence flagged and stored.</t>
 		      <t>Any "a=imageattr" attributes MUST be parsed as specified in
                 <xref target="RFC6236"/>, Section 3, and their values stored.</t>
+              <t>If present, any "a=rid" lines MUST be parsed as specified in
+	      <xref target="I-D.ietf-mmusic-rid"></xref>, Section 10, and their values stored.</t>
+              <t>If present, any "a=simulcast" lines MUST be parsed as specified
```
juberti: Are multiple simulcast lines (of the same type) allowed?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1743,6 +1761,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>For MediaStreamTracks that are still present, the "a=msid",
               "a=ssrc", and "a=ssrc-group" lines MUST stay the same.</t>
 
+	      <t>For RtpTransceivers that are still present, the
+	      "a=rid" lines MUST stay the same.</t>
+
+	      <t>For RtpTransceivers that are still present, the
```
juberti: If the encodings have changed, is this true?

draft-ietf-rtcweb-jsep.xml
juberti: Where did we end up on the .active question? If unclear, we can mark as an open issue.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1618,7 +1618,25 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
               for this media type, an "a=bundle-only" line.</t>
-            </list></t>
+
+	      <t>If the RtpTransceiver's RtpSender is active, an "a=rid" line
+	      for each encoding specified by the application in the RtpSenders's
+	      parameters.  If no encodings have been specified, then one RID
+	      line is generated.  The rid line is specified in <xref
+	      target="I-D.ietf-mmusic-rid"></xref>.  The direction MUST be
+	      "send".  If the application has chosen a RID value, it MUST be
+	      used as the rid-identifier, otherwise a RID value MUST be chosen
```
juberti: , -> ;

juberti: I assume we want to give some guidance on the size of impl-chosen RID values (e.g. 3 bytes or less, like MIDs)

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1618,7 +1618,25 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+	      by the implementation.
+	      </t>
+
+	      <t>If the RtpTransceiver's RtpSender is active and more than one
+	      "a=rid" line has been generated, one "a=simulcast" line, with
```
juberti: one -> an

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2565,6 +2594,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                       connection, if one exists.[TODO: may need revision, i.e., use of actpass</t>
                     <t>DTLS fingerprint values, where at least one fingerprint
                       MUST be present.</t>
+		    <t>All RID reference in an "a=simulcast" line must
```
juberti: This section is about mandatory features. I think we need a new section for this discussion.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1618,7 +1618,25 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
               policy is set to "balanced", and this is not the first m= section
               for this media type, an "a=bundle-only" line.</t>
-            </list></t>
+
+	      <t>If the RtpTransceiver's RtpSender is active, an "a=rid" line
+	      for each encoding specified by the application in the RtpSenders's
+	      parameters.  If no encodings have been specified, then one RID
+	      line is generated.  The rid line is specified in <xref
+	      target="I-D.ietf-mmusic-rid"></xref>.  The direction MUST be
+	      "send".  If the application has chosen a RID value, it MUST be
```
fluffy: perhaps add a bit on how app chooses rid value here or say something like api needs to provide a way to set rid value

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2765,8 +2796,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                   "sendonly", prepare to start transmitting media using the specified primary SSRC
                   and one of the selected payload types, once the underlying transport
                   layers have been established. Otherwise, stop transmitting RTP media,
-                  although RTCP should still be sent, as described in
-                  <xref target="RFC3264"/>, Section 5.1. </t>
+                  although RTCP should still be sent, as described in 
+                  <xref target="RFC3264"/>, Section 5.1. If RID values are specified, include the RID header
+		  extension in the RTP streams, as indicated in <xref
```
fluffy: I'd like to make the using the RTP Hdr Ext form of RID here a MUST so change to say something like "MUST include" instead of just include. (actually, I think you meant it had to do this but just want it to be very clear)

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2765,8 +2796,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+		  target="I-D.ietf-mmusic-rid"></xref>, Section 4).  If simulcast is negotiated,
+		  send the number of Source RTP Streams as specified in
+		  <xref target="I-D.ietf-mmusic-sdp-simulcast"></xref>, Section 6.2.2.
+		  The "send" portion of any "a=simulcast" line from a
+		  remote description MUST be ignored, since receiving
+		  of simulcast is not supported.
```
fluffy: This is not a comment you need to deal with at all but just as a random idea ... I wonder if we really need to forbid receiving simulcast in a browser. The objects would now allow that and if you want to to do a conference where one of the browsers was elected as a MCU style conference bridge and send media to and from the other browsers, .... well that might be cool. Not a use case I care about but we might want to leave the door open to something like this.

draft-ietf-rtcweb-jsep.xml
pthatcherg: done

draft-ietf-rtcweb-jsep.xml
pthatcherg: In general, yes.  But it's mostly to offer different options, not for the answer.  So we can constrain it.  I changed it to parse one and only one and ignore the rest.

draft-ietf-rtcweb-jsep.xml
pthatcherg: The API doesn't allow adding encodings, and I don't think it can remove them either.  If you don't want to send it anymore, you set .active = false, but you don't remove it.

Do we want to allow removal?

draft-ietf-rtcweb-jsep.xml
pthatcherg: done

draft-ietf-rtcweb-jsep.xml
pthatcherg: done

draft-ietf-rtcweb-jsep.xml
pthatcherg: This section is called "Semantics Verification" and has the verification of several other values.  It seems consistent.  Can you be more specific about where you would like it moved?

draft-ietf-rtcweb-jsep.xml
pthatcherg: I added a part about about generating them.

I can't find anywhere else in the doc that says something like "The PeerConnection API MUST allow X".  Do you see one I can model?

draft-ietf-rtcweb-jsep.xml
pthatcherg: There is a "MUST be performed" about above, which I think covers all of these steps.

draft-ietf-rtcweb-jsep.xml
pthatcherg: I originally wrote it like that and Justin's response was that we can leave that for a new draft/spec when we need it.

I do agree with him that the way I wrote it before, leaving that open, was more complex.

PR #221, opened 2016-03-03
Remove text advocating using non-matching PTs

### LINKED ISSUE #109: Add example to show bad interaction of not using different PTs with RTX/bundle
From Magnus's comments.

> Section 5.3.1: Note that for simplicity, the answerer MAY use
> different payload types for codecs than the offerer, as it is not
> prohibited by Section 6.1.  I react to the word simplicity here. I
> never consider usage of different PTs per direction anything
> simple. But, yes it is allowed by RFC 3264.  But I kind of protest to
> encourage the usage.
> 
> JU: Clearly this needs more explanation. The basic issue is the fact
> that using the same payload types gets really hairy when using RTX and
> when bundling. We can add an example of the problem.

juberti:
Fixes #109


fluffy:
LGTM

ekr:
LGTM.

PR #222, opened 2016-03-03
Clean up few little nits

fluffy:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1178,7 +1178,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           on when it is invoked, and differ depending on which specific
           parameters are changed:</t>
          <t><list style="symbols">
-            <t>Any changes to the STUN/TURN servers to use affect
+            <t>Any changes to the STUN/TURN servers to use will affect
```
juberti: 'will' seems unnecessary here?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1250,8 +1250,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                <xref target="I-D.nandakumar-mmusic-proto-iana-registration"/>
+                the UDP/TLS/RTP/SAVPF <xref target="RFC5764"/>,
+		TCP/DTLS/RTP/SAVPF
+                <xref
+		    target="I-D.nandakumar-mmusic-proto-iana-registration"/>,
+		"UDP/DTLS/SCTP" <xref
+		    target="I-D.ietf-mmusic-sctp-sdp"/>, 
+            and "TCP/DTLS/SCTP" <xref
+		    target="I-D.ietf-mmusic-sctp-sdp"/>
                 RTP profiles.
```
juberti: 5764 is specific to DTLS-SRTP, right?

draft-ietf-rtcweb-jsep.xml
fluffy: hmm - seems like we were missing a verb as I was thinking affect is noun but will defer to someone who speaks english

draft-ietf-rtcweb-jsep.xml
fluffy: yah it defines UDP/TLS/RTP/SAVPF but not the other 3 profiles

Issue #158, opened 2015-06-18
a=imageattr limiations and image rotation using CVO


gloinul:
E. Section 3.6:

I am missing a discussion on how video orientation affects what one states. Taking the above example with a maximum resolution of 640x360, which gives a landscape 16:9 PAR. If one rotates this, then it is not clear if one is allowed to have 360x640 or are limited to 203x360. Or should one actually state the different orientations maximum resolution and consider that this is simply different PAR, as 16:9 (Landscape) is 9:16 (Portrait) two different picture aspect ratios. 

It needs to be clear if a=imageattr is the limiations of the video being encoded and transmitted or the video being captured/displayed.

juberti:
Magnus looked into this and the proposed interpretation is fine, e.g. imageattr always refers to the unrotated video. However, this needs to be spelled out in the document.

fluffy:
We probably also need to talk about how CVO support is signalled and if the far end does not support CVO then this is a bit different.

juberti:
Right, that can happen. And in that case you would go from sending, say, 1280x720 to 720x1280. To minimize implementation complexity, I think the sender should act (regarding imageattr) as if the recipient supported CVO - that is, if the recipient says it can get 1280x720, we should also allow sending the portrait version (720x1280). Anything else just seems way too complicated.

gloinul:
If the peer do not support CVO, then I think the right thing to do is to include imageattr with 720x1280 as well as 1280x720 if the declaring endpoint can capture both and feed its encoder both directions. However, I don't see a reason to do that prior to having attempted to negotiate CVO and failed.

juberti:
Prague: Do simple thing of only including landscape resolution in imageattr.

juberti:
Not sure if anything remains to be done here.

fluffy:
@gloinul - do you think we need to make any changes? I have lost track of where we are on this one.

PR #224, opened 2016-03-04
Resolve outstanding imageattr issues

### LINKED ISSUE #157: clarify a=imageattr example
C. Section 3.6.1:

   As an example, consider a system with a HD-capable, multiformat video
   decoder, where the application has constrained the received track to
   at most 360p.  In this case, the implemention would generate this
   attribute:

   a=imageattr:* recv [x=[16:640],y=[16:360],q=1.0]

Maybe this example should discuss the minimal aspect of this. What it really is expressing in that it can handle any resolution between 16x16 pixels up to 640x360 as the longest in respective side. No limitations to any picture aspect ratio. That the examples expresses this is missing in the example text.

juberti:
Fixes #157, #158


fluffy:
LGTM

Issue #143, opened 2015-06-06
Allow CreateOffer in remote-pranswer?


ekr:
Should you be allowed to do CreateOffer when in remote-pranswer?

ekr:
@fluffy PTAL

juberti:
Context for this issue: remote-pranswer means we have a remote description. Should it be used when generating the offer?

fluffy:
No you can't create an second offer when you have not got a final answer. If you want to do that what you need to do is cancel the current offer by rolling back and now you are in a stable state and and do a new createOffer

alvestrand:
Creating the second offer should be allowed. It gives you some SDP that you can inspect, and that can be useful - might even be useful in choosing between resetting the state or waiting for a final answer.
It would be pretty dumb to send the generated offer to your communication partner, and calling SetLD() with it would result in a failure.

juberti:
Agree with alvestrand. No action needed, I think.

fluffy:
and just to add more - my recollection of the folks in the room at IETF 93 was they agreed with HTA position

ekr:
This is implicitly allowed by the spec. No action needed.

juberti:
Need to remove open issue on this.

PR #238, opened 2016-03-04
Expand imageattr text

### LINKED ISSUE #155: Usage of picture aspect ratio in a=imageattr
B. Section 3.6.1:

   Otherwise, an "a=imageattr" attribute is created with "recv"
   direction, and the resulting resolution space formed by intersecting
   the decoder limits and constraints is used to specify its minimum and
   maximum x= and y= values.  If the intersection is the null set, i.e.,
   there are no resolutions that are permitted by both the decoder and
   the mandatory constraints, this SHOULD be represented by x=0 and y=0
   values.

   The rules here express a single set of preferences, and therefore,
   the "a=imageattr" q= value is not important.  It SHOULD be set to
   1.0.

What about end-points that can produce and consume different aspect ratios on its video. Such a system is likely to have different constraints on its min and max resolution in X and Y for 1:1, 4:3, 16:9 and 1:2.35 aspect ratios. It is unclear if that is allowed, and even recommend practice. And in this case there might be a usage for the q value.

juberti:
Fixes #155, #143
Discusses multiple imageattr; use of different codecs; no aspect ratio
handling; preserve aspect ratio when scaling


draft-ietf-rtcweb-jsep.xml
```diff
@@ -629,19 +629,32 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>This specification prescribes more specific behavior.
             When a sender of a given MediaStreamTrack, which is producing video
             of a certain resolution, receives an "a=imageattr recv"
-            attribute, it MUST first check to see if the original resolution
-            meets the criteria specified in the attribute, and transmit it
-            untouched if so. Note that when considering a MediaStreamTrack that
+            attribute, it MUST check to see if the original resolution
+            meets the size criteria specified in the attribute, and adapt the
+            resolution accordingly by scaling (if needed).
+            Note that when considering a MediaStreamTrack that
```
ekr: Maybe "(if needed and allowed by policy)"

ekr:
LGTM

draft-ietf-rtcweb-jsep.xml
juberti: "(if appropriate)"?

fluffy:
LGTM

PR #240, opened 2016-03-04
Revise description of when to send candidates. Fixes #211

ekr:

PR #245, opened 2016-03-05
Tweaks to #240

juberti:
Change position of initial paragraph on handling of bundled m= sections;
simplify some text; remove discussion of ICE restart option in answers.


juberti:
@ekr FYI

PR #246, opened 2016-03-08
s/omit/emit

martinthomson:
Seems like an important distinction.

I went with "MUST NOT emit", though "MUST omit" could work too.


juberti:
wow lgtm

PR #254, opened 2016-03-16
Mentioning the end-of-candidates indication in a couple sections.

taylor-b:
Specifically, "addIceCandidate" and "Applying a Remote Description".

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1243,10 +1243,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           and/or pending remote description according to the rules defined for
           Trickle ICE. Connectivity checks will be sent to the new candidate.</t>
 
-          <t>If this method is called with a null candidate, it provides
-          an end-of-candidates indication (as defined in
-          <xref target="I-D.ietf-ice-trickle"/>) to the ICE Agent for all
-          media descriptions in the last remote description.</t>
+          <t>This method can also be used to provide an end-of-candidates
+          indication (as defined in <xref target="I-D.ietf-ice-trickle"/>)
+          to the ICE Agent for all media descriptions in the last remote
+          description.</t>
 
           <t>This call will result in a change to the state of the ICE Agent,
           and may result in a change to media state if it results in
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1243,6 +1243,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           and/or pending remote description according to the rules defined for
           Trickle ICE. Connectivity checks will be sent to the new candidate.</t>
 
+          <t>This method can also be used to provide an end-of-candidates
+          indication to the ICE Agent, as defined in
+          <xref target="I-D.ietf-ice-trickle"/>.</t>
+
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1243,6 +1243,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           and/or pending remote description according to the rules defined for
           Trickle ICE. Connectivity checks will be sent to the new candidate.</t>
 
+          <t>This method can also be used to provide an end-of-candidates
+          indication to the ICE Agent, as defined in
+          <xref target="I-D.ietf-ice-trickle"/>.</t>
```
pthatcherg: Should we be more specific about it being the null value that mean end-of-candidates?

ekr: +1

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2723,6 +2727,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>Pair any supplied ICE candidates with any gathered local candidates, as described
             in Section 5.7 of <xref target="RFC5245"/>
             and start connectivity checks with the appropriate credentials.</t>
+            <t>If an "a=end-of-candidates" attribute is present, process the
+            end-of-candidates indication as described in
+            <xref target="I-D.ietf-ice-trickle"/>.</t>
```
juberti: Would be good to mention the appropriate section name.

draft-ietf-rtcweb-jsep.xml
juberti: Might also want to mention that this applies to all m= sections.

fluffy:
I'm fine this however it ends up. I will r+ this and people can merge it once it is done. Thanks

draft-ietf-rtcweb-jsep.xml
taylor-b: Done.

draft-ietf-rtcweb-jsep.xml
taylor-b: @pthatcherg I wasn't sure if JSEP was supposed to mention specifics about the API. But I'll change it and see if anyone objects.

@ekr Redundant because "candidate" from the above paragraph can include a null candidate? I assume the meaning of a null candidate needs to be defined here though, since it's not defined in Trickle ICE.

@juberti Done.

fluffy:
Just adding a still LGTM

draft-ietf-rtcweb-jsep.xml
juberti: @ekr, I think this is worth spelling out since it wasn't always this way, at least not in the w3c spec.

@taylor-b / @pthatcherg not sure that we should specify null, since that's a bit of a w3c-ism; rather, I think the generic notion that we can supply end-of-candidates here should be sufficient. This would match the advice given in https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-13#section-3.4.1, where it simply says that an event is generated to indicate end-of-candidates.

juberti:
lgtm once my comment from today is addressed

draft-ietf-rtcweb-jsep.xml
taylor-b: @juberti That's what I thought initially. I changed it back.

PR #215, opened 2016-03-03
Update to reflect RtpTransceivers

pthatcherg:
- Add intro for what RtpTransceiver, RtpSender, and RtpReceiver are
- createOffer, createAnswer, setLocalDescription, setRemoteDescription based on transcievers, not tracks or streams.
- remove OfferToReceiveAudio and OfferToReceiveVideo
- LS groups based on transceivers, not tracks
- port 0 based on a transceiver being stopped
- setting RtpTransceiver.mid when they are associated with an m= section
- bind m= sections from remote offer to existing RtpTransceiver when appropriate.
- create transceivers in setRemoteDescription if necessary (RtpReceiver is active, RtpSender is inactive)
- rollback may set an RtpTransceiver.mid or remove an RtpTransceiver

PR #257, opened 2016-03-21
Part 2 of transceiver PR

juberti:
Second half of #215.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1521,18 +1521,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>The next step is to generate m= sections, as specified in
           <xref target="RFC4566"/> Section 5.14, for each
-          MediaStreamTrack that has been added to the PeerConnection via the
-          addStream method. (Note that this method takes a MediaStream, which
-          can contain multiple MediaStreamTracks, and therefore multiple m=
-          sections can be generated even if addStream is only called once.)
-          m=sections MUST be sorted first by the order in which the
-          MediaStreams were added to the PeerConnection, and
-          then by the alphabetical ordering of the media type for the MediaStreamTrack.
-          For example, if a MediaStream containing both an audio and a video
-          MediaStreamTrack is added to a PeerConnection, the resultant m=audio
-          section will precede the m=video section. If a second MediaStream
-          containing an audio MediaStreamTrack was added, it would follow
-          the m=video section.</t>
+          RtpTransceiver that has been added to the PeerConnection via
+          the addTrack or addTransceiver methods, excluding stopped
+          RtpTransceivers. m= sections are generated in the order that
+          their associated RtpTransceivers were added to the
+          PeerConnection.</t>
 
           <t>Each m= section, provided it is not marked as
           bundle-only, MUST generate a unique set of ICE credentials and gather
@@ -1581,11 +1574,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               Section 2.1, containing the dummy value "9 IN IP4 0.0.0.0", because
               no candidates have yet been gathered.</t>
 
-              <t>An "a=msid" line, as specified in <xref
-              target="I-D.ietf-mmusic-msid"></xref>, Section 2.</t>
-
-              <t>An "a=sendrecv" line, as specified in <xref
-              target="RFC3264"></xref>, Section 5.1.</t>
+              <t>A direction attribute for the associated
+              RtpTransceiver as described by <xref
+              target="sec.offer-dirattr"></xref>.</t>
 
               <t>For each supported codec, "a=rtpmap" and "a=fmtp" lines, as
               specified in <xref target="RFC4566"></xref>, Section 6.
@@ -1675,15 +1666,49 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               policy is set to "balanced", and this is not the first m= section
               for this media type, an "a=bundle-only" line.</t>
 
-              <t>If the RtpTransceiver's RtpSender is active, an "a=rid" line for each
-              encoding specified by the application in the RtpSenders's parameters.
-              If no encodings have been specified, then a single "a=rid" line is generated.
-              The "a=rid" line is specified in <xref target="I-D.ietf-mmusic-rid"></xref>, and its direction MUST be "send".  If the application has chosen a RID
+              <t>If the RtpTransceiver's RtpSender is active:
+                <list style="symbols">
+                  <t>An "a=msid" line, as specified in <xref target=
+                  "I-D.ietf-mmusic-msid"></xref>, Section 2.</t>
+
+                  <t>An "a=ssrc" line, as specified in <xref target=
+                  "RFC5576"></xref>, Section 4.1, indicating the SSRC to
+                  be used for sending media, along with the mandatory
+                  "cname" source attribute, as specified in Section 6.1,
+                  indicating the CNAME for the source. The CNAME MUST be
+                  generated in accordance with Section 4.9 of
+                  <xref target="I-D.ietf-rtcweb-rtp-usage"></xref>.</t>
+
+                  <t>If RTX is supported for this media type, another
+                  "a=ssrc" line with the RTX SSRC, and an "a=ssrc-group"
+                  line, as specified in <xref target="RFC5576"></xref>,
+                  section 4.2, with semantics set to "FID" and including
+                  the primary and RTX SSRCs.</t>
+
+                  <t>If FEC is supported for this media type, another
+                  "a=ssrc" line with the FEC SSRC, and an "a=ssrc-group"
+                  line with semantics set to "FEC-FR" and including the
+                  primary and FEC SSRCs, as specified in <xref target=
+                  "RFC5956"></xref>, section 4.3. For simplicity, if both
+                  RTX and FEC are supported, the FEC SSRC MUST be the
+                  same as the RTX SSRC.</t>
+                </list>
+              </t>
+
+              <t>If the RtpTransceiver's RtpSender is active, and the
+              application has specified RID values or has specified more
+              than one encoding in the RtpSenders's parameters,
+              an "a=rid" line for each encoding specified.
+              The "a=rid" line is specified in <xref target="I-D.ietf-mmusic-rid"></xref>,
+              and its direction MUST be "send".  If the application has chosen a RID
               value, it MUST be used as the rid-identifier; otherwise a RID value MUST
               be generated by the implementation.  When generating RID values, it is
               RECOMMENDED that the values be 3 bytes or less, to allow them to
               efficiently fit into the RTP header extension defined in <xref
-              target="I-D.ietf-avtext-rid"></xref>, Section 11.</t>
+              target="I-D.ietf-avtext-rid"></xref>, Section 11.
+              If no encodings have been specified or one encoding is specified
+              but without a RID value, then no "a=rid" lines are
+              generated.</t>
 
               <t>If the RtpTransceiver's RtpSender is active and more than one
               "a=rid" line has been generated, an "a=simulcast" line, with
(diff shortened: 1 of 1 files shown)


juberti:
cc @pthatcherg; copying remaining comments from #215 here.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1521,18 +1521,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          MediaStreams were added to the PeerConnection, and
-          then by the alphabetical ordering of the media type for the MediaStreamTrack.
-          For example, if a MediaStream containing both an audio and a video
-          MediaStreamTrack is added to a PeerConnection, the resultant m=audio
-          section will precede the m=video section. If a second MediaStream
-          containing an audio MediaStreamTrack was added, it would follow
-          the m=video section.</t>
+          RtpTransceiver that has been added to the PeerConnection via
+          the addTrack or addTransceiver methods, excluding stopped
+          RtpTransceivers. m= sections are generated in the order that
```
juberti: Need to keep stopped RtpTransceivers, in the case where we keep the number of m= lines from shrinking.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1675,15 +1666,50 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
               for this media type, an "a=bundle-only" line.</t>
 
-              <t>If the RtpTransceiver's RtpSender is active, an "a=rid" line for each
-              encoding specified by the application in the RtpSenders's parameters.
-              If no encodings have been specified, then a single "a=rid" line is generated.
-              The "a=rid" line is specified in <xref target="I-D.ietf-mmusic-rid"></xref>, and its direction MUST be "send".  If the application has chosen a RID
+              <t>If the RtpSender of the RtpTransceiver associated with this
+              m=section is active:
+                <list style="symbols">
+                  <t>An "a=msid" line, as specified in <xref target=
```
juberti: probably need to keep a=msid from changing. When is it legal to change msid, if ever?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1718,11 +1745,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
 
-          <t>The next step is to generate session-level lip sync groups as
-          defined in <xref target="RFC5888"/>, Section 7.  For each MediaStream
-          with more than one MediaStreamTrack, a group of type "LS" MUST be
-          added that contains the mid values for each MediaStreamTrack in that
-          MediaStream. </t>
+          <t>The next step is to generate session-level lip sync
+          groups as defined in <xref target="RFC5888"/>, Section
+          7. For each MediaStream referenced by more than one
+          RtpTransceiver, a group of type "LS" MUST be added that
```
juberti: But how can a mediastream be referenced by a transceiver, other than by addtrack? Seems like it we need to specify how the MediaStream is referenced, especially if it is implicit.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1777,6 +1804,31 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
 
+              <t>If any RtpTransceiver has been added, and there
+              exists an m= section with a zero port in the current
+              local description or the current remote description,
+              that m= section MUST be recycled by generating an m=
+              section for the added RtpTransceiver as if the m=
+              section were being added to the session description, placed at
+              the same index as the m= section with a zero port.</t>
+
+              <t>If an RtpTransceiver is stopped and is not associated
```
juberti: unless we need to add the m= section per SDP rules.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1827,23 +1880,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-              <t>If any MediaStreamTracks have been added, and there exist
-              m= sections of the appropriate media type with no associated
-              MediaStreamTracks (i.e. as described in the preceding paragraph),
-              those m= sections MUST be recycled by adding the new
-              MediaStreamTrack to the m= section. This is done by adding the
-              necessary "a=msid", "a=ssrc", and "a=ssrc-group" lines to the
-              recycled m= section, and removing the "a=recvonly" attribute.
-              </t>
-            </list></t>
+              <t>The directional attribute MAY change, depending on
```
juberti: It's not clear what this text is trying to say. Does the guidance here supersede any of the guidance above? If not, we can just remove it.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1818,8 +1870,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               "a=candidate" and "a=end-of-candidates" MUST be omitted.
               </t>
 
-              <t>For MediaStreamTracks that are still present, the "a=msid",
-              "a=ssrc", and "a=ssrc-group" lines MUST stay the same.</t>
+              <t>For RtpTransceivers with an RtpSender that is still
+              active, the "a=msid", "a=ssrc", and "a=ssrc-group" lines
```
juberti: I don't think these values get to change just because the transceiver is no longer active (e.g. temporarily recvonly).

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1827,23 +1880,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              section for the added RtpTransceiver as if the m=
+              section were being added to session description, except
+              that instead of adding it, the generated m= section
+              replaces the m= section with a zero port.
+          </t>
+            </list>
+      </t>
 
           <t>If the initial offer was applied using setLocalDescription, and
           an answer from the remote side has been applied using
```
juberti: I think the section below needs to be updated for transceivers.

juberti:
@pthatcherg see what you can do with these comments. If we can't address them all, we can simply mark the remainder with OPEN ISSUE markers and GH bugs.

draft-ietf-rtcweb-jsep.xml
pthatcherg: Yes, in #215, I change it, but failed to push in time, to say "excluding stopped RtpTranscievers that are not associated with any m= section (either due to m= section recycling or having been stopped before being associated with an m= section)".  I'll change it to that.

draft-ietf-rtcweb-jsep.xml
pthatcherg: I don't think you can change the MSID unless you recycle the MID, but this section is for initial offers.  Not changing the MSID is down below.

draft-ietf-rtcweb-jsep.xml
pthatcherg: Both addTrack and addTransceiver take a list of MediaStreams, and that's how they are referenced.   Are you simply suggesting that we say "MediaStream referenced by more than one RtpTransceiver (via the addTrack and addTransceiver methods)"?

draft-ietf-rtcweb-jsep.xml
pthatcherg: If the RtpTransceiver is not associated with an m=section, that means the the m= section has been recycled, which means there is a new RtpTransceiver taking it's place, and the m=section will be generated.  In other words, it's impossible to end up needing to generate an m=section per SDP rules without having an RtpTransceiver backing it.

draft-ietf-rtcweb-jsep.xml
pthatcherg: Agreed, I'll change this to say "MUST not change", regardless of the state.

draft-ietf-rtcweb-jsep.xml
pthatcherg: No, it doesn't, and yes, we can remove it.

draft-ietf-rtcweb-jsep.xml
pthatcherg: Reading it, I think it can just be deleted, since it's taken care of elsewhere.  In other words, there are no more exceptions like these.

pthatcherg:
I resolved everything, I think, and merge those changes into this PR (I wasn't sure how to push directly from the command line).

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1521,18 +1521,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          MediaStreamTrack is added to a PeerConnection, the resultant m=audio
-          section will precede the m=video section. If a second MediaStream
-          containing an audio MediaStreamTrack was added, it would follow
-          the m=video section.</t>
+          RtpTransceiver that has been added to the PeerConnection via
+          the addTrack or addTransceiver methods, excluding stopped
+          RtpTransceivers that are not associated with an m= section
+          (either due to an m= section being recycled or an
+          RtpTransceiver having been stopped before being associated
+          with an m= section), m= sections are generated in the order
```
juberti: Seems like this should be split into two sentences somewhere.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1718,11 +1748,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
 
-          <t>The next step is to generate session-level lip sync groups as
-          defined in <xref target="RFC5888"/>, Section 7.  For each MediaStream
-          with more than one MediaStreamTrack, a group of type "LS" MUST be
-          added that contains the mid values for each MediaStreamTrack in that
-          MediaStream. </t>
+          <t>The next step is to generate session-level lip sync
+          groups as defined in <xref target="RFC5888"/>, Section
+          7. For each MediaStream referenced by more than one
+          RtpTransceiver (by passing those MediaStreams as arugments
```
juberti: spelling: arguments

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2315,8 +2303,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               "a=candidate" and "a=end-of-candidates" MUST be omitted.
               </t>
 
-              <t>For MediaStreamTracks that are still present, the "a=msid",
-              "a=ssrc", and "a=ssrc-group" lines MUST stay the same.</t>
+              <t>For RtpTransceivers that are not stopped, the "a=msid",
```
juberti: why do you mention the remote description here?

juberti:
Overall LGTM. A few comments, plus the build is broken.

draft-ietf-rtcweb-jsep.xml
pthatcherg: I don't.  I just removed it.

draft-ietf-rtcweb-jsep.xml
pthatcherg: I changed it to be more clear, and more accurate.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1520,15 +1520,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          RtpTransceiver that has been added to the PeerConnection via
-          the addTrack or addTransceiver methods, excluding stopped
-          RtpTransceivers that are not associated with an m= section
-          (either due to an m= section being recycled or an
-          RtpTransceiver having been stopped before being associated
-          with an m= section), m= sections are generated in the order
+          <xref target="RFC4566"/> Section 5.14.  An m= section is
+          generated for each RtpTransceiver that has been added to the
+          PeerConnection via the addTrack, addTransceiver, and
+          setRemoteDescription methods.  This is done in the order
```
pthatcherg: I realized that since the 'subsequent offer' text says follow the same steps here, we need to reference setRemoteDescription here, which can also create a transceiver, even if it's a little awkward to mention that for the initial offer.

juberti: maybe just point back at this and reference setRD in the subsequent offer section?

PR #262, opened 2016-03-25
Cj fix1

fluffy:
This is all strictly editorial and I think if either of you give it a quick scan, you can just merge it


juberti:
LGTM - will merge

PR #263, opened 2016-03-25
Fixup how LS groups work.

### LINKED ISSUE #162: LipSync groups violating RFC5888

fluffy:
Fixes #162 

Replaces https://github.com/rtcweb-wg/jsep/pull/167


fluffy:
@juberti this just flattens all the changes in last PR to the final thin and puts it onto new base with no merge issue. Still needs changes to align with new sender / receiver objects

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2621,19 +2621,28 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                    target="RFC5888" />, it is allowed when implementing
-                    this specification. [[OPEN ISSUE: This is still
-                    under discussion. See:
-                    https://github.com/rtcweb-wg/jsep/issues/162.]]</t>
+                    <t>The next step is to generate session-level lip
+                    sync groups as defined in <xref target="RFC5888"/>,
+                    Section 7. All Tracks in the same MediaStream need
+                    to be placed in the same "LS" group as long as
+                    theses tracks were in the offered to by synchronized
+                    in the offer.  RFC5888 only allows the groups in the
```
alvestrand: I can't parse the "as long as" sentence. The sentence seems to make sense if it stops with "in the same "LS" group". If the rest of the sentence modifies it, it's not clear what it says.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2621,19 +2621,28 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                    answer to be a subset of what was in the offer. If
+                    there are tracks that the same MediaStream that can
+                    not be added to the "LS" group, once the state moves
+                    back to "stable" the browser MAY initiate another
+                    offer/answer negotiation put all the tracks in the
+                    MediaStream into the "LS" group or they can be left
+                    in a state that though answer wishes them to be
+                    synchronized, the answer can not signal them as
+                    synchronized because the offer did put them in the
+                    same LS group. The working group consider varios
```
alvestrand: English nit: "The working group considered various solutions to this rare corner condition". In general, I don't think WG considerations belong in the normative part of the document, so I'd be fine with yanking the whole sentence.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2621,19 +2621,28 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                    theses tracks were in the offered to by synchronized
+                    in the offer.  RFC5888 only allows the groups in the
+                    answer to be a subset of what was in the offer. If
+                    there are tracks that the same MediaStream that can
+                    not be added to the "LS" group, once the state moves
+                    back to "stable" the browser MAY initiate another
+                    offer/answer negotiation put all the tracks in the
+                    MediaStream into the "LS" group or they can be left
+                    in a state that though answer wishes them to be
+                    synchronized, the answer can not signal them as
```
alvestrand: This paragraph is hard to parse. The paragraph seems to say the same thing twice in two different ways, but the language is different enough that I can't be sure there's not a subtle difference.

Suggested replacement for the whole paragraph starting at "If there are tracks... if it is critical for their use case": "If there are tracks that belong to a MediaStream that cannot be added to the "LS" group defined by the offerer, the answerer MAY add them without synchronization. It MAY also skip adding the tracks to the answer, and instead create a new offer after sending the answer."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2621,19 +2621,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                    not in that LS group in the associated
-                    offer. Although this is not allowed by <xref
-                    target="RFC5888" />, it is allowed when implementing
-                    this specification. [[OPEN ISSUE: This is still
-                    under discussion. See:
-                    https://github.com/rtcweb-wg/jsep/issues/162.]]</t>
+                    <t>The next step is to generate session-level lip
+                    sync groups as defined in <xref target="RFC5888"/>,
+                    Section 7. For each group of type "LS" present in the offer, determine which of the RtpTransceivers referenced
+                    by the supplied mid values reference a common MediaStream.
```
ekr: a common MediaStream in the answer?

juberti:
PTAL

fluffy:
LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2621,19 +2621,20 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                    offer. Although this is not allowed by <xref
-                    target="RFC5888" />, it is allowed when implementing
-                    this specification. [[OPEN ISSUE: This is still
-                    under discussion. See:
-                    https://github.com/rtcweb-wg/jsep/issues/162.]]</t>
+                    <t>The next step is to generate session-level lip
+                    sync groups as defined in <xref target="RFC5888"/>,
+                    Section 7. For each group of type "LS" present in the offer,
+                    determine which of the local RtpTransceivers identified by
+                    the group's mid values reference a common local MediaStream
```
ekr: Nit: perhaps "that group's" mid values?

is which singular here so we need "references"? not sure.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2621,19 +2621,20 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                    (as specified in the addTrack and
+                    addTransceiver methods). If at least two such
+                    RtpTransceivers exist, a group of type "LS" with the mid
+                    values of these RtpTransceivers MUST be added. Otherwise,
```
ekr: Nit: "those"?

ekr:
LGTM

draft-ietf-rtcweb-jsep.xml
juberti: Since 'these' references something nearby, and the transceivers were mentioned in the initial clause, I'm inclined to keep 'these'.

ekr: Fine by me. The question mark indicated I was unsure.

draft-ietf-rtcweb-jsep.xml
juberti: changed to 'that'; agree that's slightly more specific
which is plural here, I believe

PR #269, opened 2016-04-15
Changed restriction on subsequent offers to refer to most recent answ…

### LINKED ISSUE #266: When should it be possible to add previously discarded codecs in a subsequent offer?
Section 5.2.2, "Subsequent Offers", says:

>    o  The m= line and corresponding "a=rtpmap" and "a=fmtp" lines MUST
>       only include codecs present in the remote description.

However, consider a situation where Alice offers codecs `foo` and `bar`, and Bob answers with only codec `foo`, thereby rejecting `bar`. If Bob were to now create a subsequent offer, his remote description still contains both `foo` and `bar`, so he would technically be allowed to offer both codecs, even though `bar` was previously rejected.

Is this intentional? It seems to go against the spirit of the rule. Perhaps the rule is meant to be "only include codecs present in the current local and remote descriptions".

For context, see: https://www.ietf.org/mail-archive/web/rtcweb/current/msg15697.html

ekr:
…er. Fixes #266

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2451,21 +2451,20 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
                         <t>The m= line and corresponding "a=rtpmap" and
                         "a=fmtp" lines MUST only include codecs present
-                        in the remote description.</t>
+                        in the most recent answer.</t>
 
                         <t>The RTP header extensions MUST only include
-                        those that are present in the remote
-                        description.</t>
+                        those that are present in the most recent answer.</t>
 
                         <t>The RTCP feedback extensions MUST only
-                        include those that are present in the remote
-                        description.</t>
+                        include those that are present in the most recent
+                        answer.</t>
 
                         <t>The "a=rtcp-mux" line MUST only be added if
-                        present in the remote description.</t>
+                        present in the most recent answer.</t>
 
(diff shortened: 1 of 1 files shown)


juberti:
lgtm

taylor-b:
Just one observation: If a JSEP implementation offers codec A to an SFU, and the SFU responds with codecs A and B in the answer (which it's allowed to do), with the "most recent answer" wording, that means the JSEP implementation is technically allowed to offer codecs A and B in a subsequent offer. Is that the intention?

I'm fine with it, I'm just making sure. This would then be the only way to seamlessly switch to a new codec. For example, after a hardware encoder becomes available.

juberti:
oh, hmm, I see that is legal in 3264. That wasn't the intention, but let me think about it.

fluffy:
I think the case @taylor-b points out sounds fine.

fluffy:
LGTM but if we end up changing this a bunch I would like to look at it again

juberti:
The rationale for keeping to the negotiated set was that it becomes ambiguous how to prioritize the codecs in a reoffer. e.g. if you offer ABC and the answer is CB (while supporting DCB), if we keep to the negotiated set then clearly both sides by default should re-offer with CB. Otherwise, it's not clear where A or D should go. 

In this case, with an A offer and an AB answer, the callee's reoffer would have to be A, and the caller's reoffer could be either A or AB (if B somehow was now available). That is non-ambiguous, and seems fine.

Generally, this is a corner case, and as such taking the simplest path seems best.

PR #271, opened 2016-04-15
Reject on codec mismatch. Fixes #247

### LINKED ISSUE #247: Document what should happen when there are no matching codecs in answer
Split from https://bugs.chromium.org/p/webrtc/issues/detail?id=5605#c3
Something for the "Setting a remote description section"

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2659,6 +2659,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
 
+                    <t>For each offered m= section, if no supported codec
+                    is present in the offer, the corresponding
+                    m= section in the answer MUST be marked as rejected
+                    by setting the port in the m= line to zero, as
+                    indicated in <xref target="RFC3264"></xref>, Section
+                    6., and further processing for this m= section can
+                    be skipped.</t>
+
                     <t>Provided that is not the case, each m= section in
```
juberti: Probably need to tweak this text now that we have two ways to reject, e.g. replace "Provided that is not the case" with "Otherwise"

juberti:
LG modulo comment

fluffy:
LGTM

Issue #239, opened 2016-03-04
When is the SDP sess-version in o= line updated


fluffy:
When the SDP changes, including new trickle ICE candates, the sess-version needs to go up. Need to decide how to do this and add text to support. 

See http://tools.ietf.org/html/rfc4566#section-5.1

juberti:
Discussion at IETF 95 indicated that o= is basically used to indicate whether SDP has changed. This would indicate that o= would need to increase any time the SDP changed, including addition of candidates.

Question around rollback - if SDP with version 12 is set, then SDP with 13 is offered, and then rolled back, what does the SDP version become? Seems like it should stay at 12, but the next offer should be 14, not 13.

Simplest policy: createOffer returns SDP with o= version incremented by 1 from the previous setLocal SDP, even if that SDP was rolled back.

cdh4u:
In Justin's example, I don't think the next version would need to be 14, as 13 was never accepted. On the other hand, SDP version numbers are cheap, so...

ekr:
Consensus in Berlin

  V is the version in the o=line in last call to setLocal
  o=V+1 on CreateOffer() if if anything changes

cdh4u:
I think that consensus shall be verified with MMUSIC. I want to make sure that we don't send two different offers with the same version number, and the answerer rejects them both because it thinks they are identical.

juberti:
For posterity, see some additional context and history in https://github.com/rtcweb-wg/jsep/pull/277

PR #274, opened 2016-04-29
clarify what happens with version in SDP o= lines during roll back

fluffy:
Fixes https://github.com/rtcweb-wg/jsep/issues/239

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2308,11 +2308,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     subject to the following restriction: 
                     <list style="symbols">
 
-                        <t>The fields of the "o=" line MUST stay the
-                        same except for the &lt;session-version&gt;
-                        field, which MUST increment if the session
-                        description changes in any way, including the
-                        addition of ICE candidates.</t>
+                        <t>The fields of the "o=" line MUST stay the same except
+                        for the &lt;session-version&gt; field, which MUST
+                        increment to be greater than any other SDP offers in
+                        this session if the session description changes in any
+                        way, including the addition of ICE candidates. This
+                        means that if an Offer with version 12 was generated,
+                        then an Offer with version with 13 was generated, then
+                        the SDP was rolled back to the Offer with version 12,
+                        the next offer would have a version 14 or larger. </t>
                     </list></t>
 
                     <t>If the initial offer was applied using


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2308,11 +2308,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
                     <list style="symbols">
 
-                        <t>The fields of the "o=" line MUST stay the
-                        same except for the &lt;session-version&gt;
-                        field, which MUST increment if the session
-                        description changes in any way, including the
-                        addition of ICE candidates.</t>
+                        <t>The fields of the "o=" line MUST stay the same except
+                        for the &lt;session-version&gt; field, which MUST
+                        increment to be greater than any other SDP offers in
```
ekr: s/increment to/

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2308,11 +2308,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                        this session if the session description changes in any
```
ekr: s/changes/has changed/

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2308,11 +2308,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                        way, including the addition of ICE candidates. This
+                        means that if an Offer with version 12 was generated,
```
ekr: We seem to be using lower-case offer.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2308,11 +2308,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                        then an Offer with version with 13 was generated, then
+                        the SDP was rolled back to the Offer with version 12,
+                        the next offer would have a version 14 or larger. </t>
```
ekr: Can you rewrite this with more reference to the APIs? I.e., "If createOffer() returns a ..."

juberti:
The idea here seems right, but I don't think this is the right place in the overall text for it. Specifically, we need to point out that this affects even getLocalDescription, so we probably need to cover this in section 3.

fluffy:
Replaced by https://github.com/rtcweb-wg/jsep/pull/277

PR #276, opened 2016-05-06
Write up initial version of addTrack

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1140,7 +1140,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                 </section>
 
                 <section title="addTrack" anchor="sec.addTrack">
-                  <t>[TODO]</t>
+                  <t>
+                    The addTrack method adds a MediaStreamTrack to the PeerConnection.
```
juberti: Need to reference the MediaStreams argumetn here, i.e. regarding grouping/lip-sync, as we will refer to the MediaStream grouping in that section.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1140,7 +1140,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                    addTrack attempts to minimize the number of transceivers as follows:
+                    The track will be attached to the first compatible transceiver
+                    (of the same media type) which has never had a direction of
+                    "sendonly" or "sendrecv". If no such transceiver exists,
```
juberti: This seems mostly right, but I forget how m= section reuse works. i.e., are stopped transceivers GCed, such that the 'no such transceiver' clause applies? If so, where is that documented?
@pthatcherg probably knows

fluffy: So do we want to mention here that you try and match things in the same Lip Sync group if possible.

ekr: Justin, here is the text I am relying on:

"If an RTCRtpSender exists in connection's set of senders that has never been used to send (the corresponding media description has always had a direction of recvonly or inactive), then set that sender's track to track and return the sender.

Doing so will cause future calls to createOffer and createAnswer to mark the corresponding media description as sendrecv or sendonly, as defined in [JSEP]."

juberti: I'm not sure this is right - wouldn't this lead to reuse of a recvonly m= section, which we decided not to pursue?

ekr: @pthatcherg we need your input here.

pthatcherg: Transceivers that are stopped are never reused (stopping is permanent), but the m= section slot can be filled with a new transceiver.   I think the text needs to specify:

1.  The media type is the same (already done)
2.  The direction has never been "sendonly" or "sendrecv" (already done)
3.  The transceiver is not stopped (not done yet)

I think we need to add 3 with something like "The track will be attached to the first compatible transceiver (of the same media type and not stopped) ...".

pthatcherg: This *would* have the effect of upgrading a recvonly line to sendrecv if addTrack is called after the initial offer/answer exchange.  

It was a while ago, but I recall that we toyed with the idea of having something like "only reuse the transceiver if it was created by setRemoteDescription(offer)", but there were some edge cases we worked through and decided that this was the most simple approach to take (which we then wrote in the W3C draft).  I don't recall what those were, but I guess we could go dig up old notes if we think it's worth it.

juberti: I'd be interested in those notes. The setRD proposal seems more correct than 'never send*'

juberti:
I'd suggest that we try to move forward with this and leave the exact ruleset as an open issue pending Peter's investigation.

ekr:
OK, I updated the PR to have the MST text. I think we had Cullen's r+ so ready to merge?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1140,7 +1140,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                  <t>[TODO]</t>
+                  <t>
+                    The addTrack method adds a MediaStreamTrack to the PeerConnection,
+                    using the MediaStream argument to associate the track with other
+                    tracks in the same MediaStream, so that they can be added to the
+                    same "LS" group when creating an offer or answer.
+                    addTrack attempts to minimize the number of transceivers as follows:
+                    The track will be attached to the first compatible transceiver
+                    (of the same media type) which has never had a direction of
+                    "sendonly" or "sendrecv". If no such transceiver exists,
```
juberti: I think we need to note as an open issue that the reuse of 'recvonly'/'inactive' m= lines may not be correct.

juberti:
Merging this and will file an issue on the algorithm.

PR #277, opened 2016-05-06
Document o= line behavior. Fixes #239.

### LINKED ISSUE #239: When is the SDP sess-version in o= line updated

ekr:


fluffy:
LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -454,6 +454,27 @@ setRemote(ANSWER) |                                   |
                 application will not need to read or change them.</t>
 
             </section>
+            <section title="Session Description Versions"
+                     anchor="sec.session-description-vers">
+              <t>
+                The SessionDescription contains a "session-version"
+                field which is required by <xref target="RFC4566"/> to increase with
```
fluffy: I think 4566 only requires it to change not increment but I think we should say it needs to increment because that makes things easier.

fluffy:
Looks good but one comment inline ...

draft-ietf-rtcweb-jsep.xml
ekr: 4566 requires it to increase
   <sess-version> is a version number for this session description.  Its
      usage is up to the creating tool, so long as <sess-version> is
      increased when a modification is made to the session data.  Again,
      it is RECOMMENDED that an NTP format timestamp is used.

fluffy:
Justin is going to write up some stuff on this one ...

juberti:
My main concern with this is that you end up with situations where an immediate createOffer -> setLD -> getLD sequence can give you make an o= that doesn't match what you thought you set (or sent over the wire).

I think the simplest thing here would be to make createOffer always increment o=, but not increment the o= in a SDP returned from getLD.

juberti:
Cullen raised a concern about clients that need to do pseudo-trickle by sending full offers every time a new candidate is gathered. This works fine even if we only increment o= on a new createOffer call, as long as the client calls createOffer for each new offer that is sent out (instead of just polling getLD).

juberti:
An example of my concern with changing o= on every interaction:
1) call createOffer and get o=0
2) call setLD (gathering starts), send offer with o=0
3) call getLD (o=0 initially, but could change to o=1, 2, 3 as candidates are gathered)
4) call createOffer and get o=4 (for example), send it off
5) call setLD, but before this occurs, a candidate was gathered, making the 'current' o=4
6) Now the meaning of o=4 is ambiguous; the offer that you just created and sent has o=4, but the 'old' offer with a new candidate is also o=4.
7) So, if you discard the offer, but instead call getLD and send that to the other side, the remote side will be confused because both old and new SDPs have o=4.
8) Incrementing to o=5 in step 5) doesn't seem much better, since now we're incrementing the current o= based on an outstanding offer, which seems strange.

The problem exists because we are trying to to increment on both createOffer and candidate gathering, and these things are somewhat different. 

I maintain that the simplest, most reliable policy is to maintain a local o= counter (call it V), initialized to 0, and createOffer/Answer always return o= V+1. When calling setLocal, V is set to the value from the o= line (unless it is a rollback, in which case it does not change). Note that the only reason V is needed as a separate counter from o= is to handle the rollback case, since o= still needs to increment for the next offer after a rollback. getLocal always returns the value from setLD, as one would expect. The o= line (and V) do not increment on candidate gathering.

ekr:
I find Justin's argument fairly persuasive here.

As far as I can tell, the only situation in which Justin's algorithm produces a different result from that dictated in RFC 4566 is when candidates are added (asynchronously) and you use getLocal. I remain unconvinced that any implementation which is non-trickle-aware is going to get reasonable results by repeated calls to getLocal() because non-trickle ICE is not set up to handle the case of additional candidates. And a smart implementation can get the right behavior as Justin indicates.

juberti:
I suppose we also need to decide what to do if someone calls setLD/setRD with a non-incremented vnum (or possibly, a smaller one). My inclination would be to simply accept it, since we don't care.

ekr:
@juberti @fluffy I think the way to proceed here is to make this change and call it out to the list. @fluffy, objections?

fluffy:
I can live with Justin's proposal but would like a bit of text somewhere that warns that the JD developer can get SDP that is different (due to candidates) but has the same session version.

ekr:
I'm not going to have time to do this

fluffy:
Have a look at https://github.com/rtcweb-wg/jsep/pull/309

juberti:
obsoleted by #309

PR #279, opened 2016-05-11
Reject a media description if it's not bundled and policy is max-bundle.

taylor-b:
Also restructured the "rejecting media descriptions" sections since
there are now four things that can cause it to be rejected.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1102,9 +1102,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         bundle-only. This policy aims to minimize
                         candidate gathering and maximize multiplexing,
                         at the cost of less compatibility with legacy
-                        endpoints. When acting as answerer, the browser
-                        will reject any m= section that is not in a
-                        BUNDLE group.</t>
+                        endpoints. When acting as answerer, the
+                        implementation will reject any m= section that
+                        is not in a BUNDLE group.</t>
 
                     </list></t>
 
@@ -1128,9 +1128,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         <t hangText="require:">The browser will only
                         gather RTP candidates. This halves the number
                         of candidates that the offerer needs to gather.
-                        When acting as answerer, the browser will
-                        reject any m= section that does not provide an
-                        "a=rtcp-mux" attribute.</t>
+                        When acting as answerer, the implementation
+                        will reject any m= section that does not
+                        provide an "a=rtcp-mux" attribute.</t>
 
                     </list></t>
 
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1104,7 +1104,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         at the cost of less compatibility with legacy
                         endpoints. When acting as answerer, the
                         implementation will reject any m= section that
-                        is not in a BUNDLE group.</t>
+                        is not in a bundle group.</t>
 
                     </list></t>
 
@@ -1130,7 +1130,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         of candidates that the offerer needs to gather.
                         When acting as answerer, the implementation
                         will reject any m= section that does not
-                        provide an "a=rtcp-mux" attribute.</t>
+                        contain an "a=rtcp-mux" attribute.</t>
 
                     </list></t>
 
@@ -2683,10 +2683,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                       <t>No supported codec is present in the offer.</t>
 
                       <t>The bundle policy is "max-bundle" and the m=
-                      section is not in a BUNDLE group.</t>
+                      section is not in a bundle group.</t>
 
                       <t>The RTP/RTCP multiplexing policy is "require"
-                      and the m= section doesn't provide an
+                      and the m= section doesn't contain an
                       "a=rtcp-mux" attribute.</t>
                     </list></t>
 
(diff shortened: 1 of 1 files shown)


pthatcherg:
I think this does a good job clarifying these edge cases and is the behavior we want.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1102,7 +1102,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         bundle-only. This policy aims to minimize
                         candidate gathering and maximize multiplexing,
                         at the cost of less compatibility with legacy
-                        endpoints.</t>
+                        endpoints. When acting as answerer, the browser
```
juberti: typically we use 'endpoint' or 'implementation', rather than 'browser'

taylor-b: I just copied the terminology used in the RTCP mux section. I'll change that as well.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2667,21 +2669,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                    indicated in <xref target="RFC3264"></xref>, Section
-                    6., and further processing for this m= section can
-                    be skipped.</t>
+                      <t>No supported codec is present in the offer.</t>
+
+                      <t>The bundle policy is "max-bundle" and the m=
+                      section is not in a BUNDLE group.</t>
+
+                      <t>The RTP/RTCP multiplexing policy is "require"
+                      and the m= section doesn't provide an
```
juberti: provide -> contain. Note that this text may need to be generalized if we go down the a=rtcp-mux-exclusive path.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2667,21 +2669,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                    is present in the offer, the corresponding
-                    m= section in the answer MUST be marked as rejected
-                    by setting the port in the m= line to zero, as
-                    indicated in <xref target="RFC3264"></xref>, Section
-                    6., and further processing for this m= section can
-                    be skipped.</t>
+                      <t>No supported codec is present in the offer.</t>
+
+                      <t>The bundle policy is "max-bundle" and the m=
+                      section is not in a BUNDLE group.</t>
```
juberti: BUNDLE -> bundle, since it's not an acronym

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1102,7 +1102,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         bundle-only. This policy aims to minimize
                         candidate gathering and maximize multiplexing,
                         at the cost of less compatibility with legacy
-                        endpoints.</t>
+                        endpoints. When acting as answerer, the
+                        implementation will reject any m= section that
+                        is not in a BUNDLE group.</t>
```
juberti: BUNDLE -> bundle

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1126,9 +1128,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         <t hangText="require:">The browser will only
                         gather RTP candidates. This halves the number
                         of candidates that the offerer needs to gather.
-                        When acting as answerer, the browser will
-                        reject any m= section that does not provide an
-                        "a=rtcp-mux" attribute.</t>
+                        When acting as answerer, the implementation
+                        will reject any m= section that does not
+                        provide an "a=rtcp-mux" attribute.</t>
```
juberti: provide -> contain

draft-ietf-rtcweb-jsep.xml
taylor-b: For now I'm just restructuring the existing text here to make it more readable. I'll change "provide" to "contain" but we should probably do the generalizing it in a separate PR.

juberti:
lgtm

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1126,9 +1128,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         <t hangText="require:">The browser will only
                         gather RTP candidates. This halves the number
                         of candidates that the offerer needs to gather.
-                        When acting as answerer, the browser will
-                        reject any m= section that does not provide an
-                        "a=rtcp-mux" attribute.</t>
+                        When acting as answerer, the implementation
+                        will reject any m= section that does not
+                        contain an "a=rtcp-mux" attribute.</t>
```
fluffy: prefer that it not reject it but instead just treat it as if it does "a=rtcp-mux"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2667,21 +2669,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                    6., and further processing for this m= section can
-                    be skipped.</t>
+                      <t>No supported codec is present in the offer.</t>
+
+                      <t>The bundle policy is "max-bundle" and the m=
+                      section is not in a bundle group.</t>
+
+                      <t>The RTP/RTCP multiplexing policy is "require"
+                      and the m= section doesn't contain an
+                      "a=rtcp-mux" attribute.</t>
```
fluffy: same as above comment

fluffy:
LGTM

taylor-b:
I just realized this may not be complete. Even if max-bundle is used, shouldn't *one* media description be accepted? This would ensure the same result regardless of whether the max-bundle endpoint is the offerer or answerer. It would also prevent a remote offer from being rejected if it didn't include a "BUNDLE" group simply because there was only one media description.

taylor-b:
Deficiencies should be addressed by this PR (apologies for missing them on the first pass): https://github.com/rtcweb-wg/jsep/pull/282

PR #282, opened 2016-05-17
Clarify when an m= section is rejected due to the bundle policy.

taylor-b:
If a non-bundle offer is received, and "max-bundle" is used, the first
m= section should be accepted and all others rejected. This will ensure
the same result whether or not the "max-bundle" endpoint is the offerer
or answerer (ignoring some unlikely corner cases), and will prevent a
lone media description from being rejected if the remote offer didn't
contain a bundle group simply because it wasn't necessary.

Also added equivalent rules for the "balanced" bundle policy.


ekr:
This LGTM

juberti:
This isn't exactly what I had in mind for 'balanced', but this might not be an issue in practice. If the answerer needs to handle multiple m= lines that are unbundled, it can set max-compat.

juberti:
lgtm

PR #286, opened 2016-06-03
change TIAS formula

fippo:
b=TIAS is specified in bits per second, b=AS in kbps. Hence the formula needs to multiply by 1000 to change units. Thanks @nils-ohlmeier for catching this in code review.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3806,12 +3806,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
                         <list style="format">
 
-                            <t>TIAS = AS * 0.95 - 50 * 40 * 8</t>
+                            <t>TIAS = AS * 1000 * 0.95 - 50 * 40 * 8</t>
 
                         </list>
 
                         The 50 is based on 50 packets per second, the 40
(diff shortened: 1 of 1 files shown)


nils-ohlmeier:
Even though I noticed it in the code review I had not noticed it as an issue in the spec. Good catch!

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3806,12 +3806,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
 
-                            <t>TIAS = AS * 0.95 - 50 * 40 * 8</t>
+                            <t>TIAS = AS * 1000 * 0.95 - 50 * 40 * 8</t>
 
                         </list>
 
                         The 50 is based on 50 packets per second, the 40
-                        is based on an estimate of total header size,
+                        is based on an estimate of total header size, the
+                        1000 is to change the unit from kbps to bps
```
juberti: might want to say 
 > from kbps to bps (as required by TIAS)

juberti:
nice catch - lgtm with minor comment

fluffy:
LGTM

ekr:
LGTM to me as well. I updated the text in #287

Issue #285, opened 2016-06-02
Need JSEP anchor for receipt of multiple encodings


burnburn:
In WebRTC section 5.1, addTransceiver(), it says

'When setRemoteDescription is called with a corresponding remote description that is able to receive multiple RTP encodings as defined in \[JSEP\], the RTCRtpSender may send multiple RTP encodings and the parameters retrieved via the transceiver’s sender.getParameters() will reflect the encodings negotiated.'

It is not obvious what to use as the JSEP reference because there does not seem to be a single place that talks about receipt of multiple encodings.

alvestrand:
The specification of signalling that you can recevie multiple encodings seems to belong in either the RID spec or the JSEP spec.
A decision was made some time back that WebRTC entities are only required to support receiving one incoming stream, but all the specs need to be clear on what that means in practice.

fluffy:
I think it makes sense for us to add some text in JSEP for this and create a link

juberti:
Right now we're deciding between making the pointer go to the simulcast document, or adding a new section in JSEP section 3 that would give an overview of simulcast.

PR #289, opened 2016-06-03
Add simulcast section

fluffy:
Fixes https://github.com/rtcweb-wg/jsep/issues/285


draft-ietf-rtcweb-jsep.xml
```diff
@@ -993,7 +993,24 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                same media source, the techniques from <xref
+                target="I-D.ietf-mmusic-sdp-simulcast"/> and <xref
+                target="I-D.ietf-mmusic-rid" /> are used. An offerer wanting to
+                use simulcast MUST include the "a=simulcast" attribute in the
+                offer as defined in Section 6.2.2 of <xref
+                target="I-D.ietf-mmusic-sdp-simulcast"/>. Devices implementing
+                JSEP are not required to be able to send or receive multiple
+                media streams but they MUST be able to reject undesired streams
+                as described in the <xref
+                target="I-D.ietf-mmusic-sdp-simulcast"/>. </t>
```
ekr: Isn't this redundant because that's just ordinary SDP behavior?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -993,7 +993,30 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+		        to understand that these are simulcasts of the same source
+		        media. Clients that don't support simulcast will end up
+		        selecting only of the encodings and functionality in a non
+		        simulcast way. </t>
+	      
+                <t> When using JSEP to send or receive multiple encodings of the
+                same media source, the techniques from <xref
+                target="I-D.ietf-mmusic-sdp-simulcast"/> and <xref
+                target="I-D.ietf-mmusic-rid" /> are used. An offerer wanting to
+                use simulcast MUST include the "a=simulcast" attribute in the
```
alvestrand: is this "wanting to send", "wanting to receive", "both", or is that controlled by further parameters?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -993,7 +993,30 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              </section>
+
+              <section  title="Simulcast" anchor="sec.simulcast">
+
+		        <t> JSEP supports simulcast using multiple encodings of the
+		        source media within the same m-line section of the SDP with
+		        additional SDP marking to allow clients that support simulcast
+		        to understand that these are simulcasts of the same source
+		        media. Clients that don't support simulcast will end up
+		        selecting only of the encodings and functionality in a non
```
nils-ohlmeier: "... will end up selecting only **one** of the encodings..." ?

fluffy:
LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -877,6 +877,45 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                only send a single encoding per m= section. Future versions of
+                this specification may add additional APIs to control this
+                behavior.</t>
+
+                <t> When using JSEP to transmit multiple encodings from a
+                RTPSender, the techniques from <xref
+                target="I-D.ietf-mmusic-sdp-simulcast"/> and <xref
+                target="I-D.ietf-mmusic-rid" /> are used. Specifically, when
+                multiple encodings have been configured for a RTPSender,
+                the m= section for the RTPSender will include the "a=simulcast"
```
fluffy: Do we need to say this part is just for an offer ?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -877,6 +877,45 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            <section  title="Simulcast" anchor="sec.simulcast">
+
+                <t>JSEP supports simulcast of a MediaStreamTrack, where
+                multiple encodings of the source media can be transmitted within
+                the context of a single m= section. Receivers that understand
+                simulcast and are willing to receive it will include SDP markings
+                to indicate this support, and JSEP endpoints will use these
+                markings to determine whether simulcast is permitted for a given
+                RTPSender. If simulcast support is not negotiated, the RTPSender
+                will only be permitted to use a single encoding.</t>
```
ekr: Perhaps:
"JSEP supports simulcast of a MediaStreamTrack, where
multiple encodings of the source media can be transmitted within
the context of a single m= section. The current JSEP API is designed
to allow applications to send simulcasted media but only to receive
a single encoding. This allows for applications where browser clients
send multiple encodings to a conferencing server which then selects
a single encoding to send to each client.

Applications request support for simulcast by configuring multiple
encodings on an RTPSender, which causes the m= section to be
marked as containing multiple encodings, as described below. Receivers that understand
simulcast and are willing to receive it will include SDP markings
to indicate this support, and JSEP endpoints will use these
markings to determine whether simulcast is permitted for a given
RTPSender. If the JSEP endpoint is the answerer, then the RTPSender
will only be permitted to use encodings consistent with the offer.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -877,6 +877,45 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+
+                <t>Note that the exact simulcast parameters are up to the sending
+                application. While the aforementioned SDP markings are provided
+                to ensure the remote side can receive and demux simulcast,
+                the specific resolutions and bitrates to be used for simulcast
+                are purely a send-side decision in JSEP.</t>
+
+                <t>JSEP currently does not provide a mechanism to configure receipt
+                of simulcast. This means that if simulcast is offered by the
```
ekr: "to allow browser endpoints to configure receipt of simulcast"

juberti:
Updated to incorporate feedback from comments. Some more detail needed in the last para, but the narrative should be complete.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -877,6 +877,63 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                <t>Applications request support for simulcast by configuring
+                multiple encodings on an RTPSender, which, upon generation of
+                an offer or answer, are indicated in SDP markings on the
+                corresponding m= section, as described below.
+                Receivers that understand simulcast and are willing to receive it
+                will also include SDP markings to indicate their support, and JSEP
+                endpoints will use these markings to determine whether simulcast
+                is permitted for a given RTPSender. If simulcast support is not
+                negotiated, the RTPSender will only be permitted to use a single
+                encoding.</t>
```
ekr: "Permitted" seems ambiguous here. Is the result that:

- You can add as many encodings as you want, but it will only use one.
- You won't even be able to add more encodings (say after the first O/A exchange)

juberti:
PTAL

ekr:
LGTM

PR #291, opened 2016-06-28
The sess-id field of the o= line can't be 64 bits

martinthomson:
RFC 3264 says this:

    The numeric value of the session id and version in the o line MUST
    be representable with a 64 bit signed integer.

Since the negative values that can be represented in a 64-bit signed
integer can't be represented by the sess-id grammar (`1*DIGIT`),
the only reasonable conclusion is that this has to be a 63-bit
number.

FWIW, that's what Firefox does, though it will accept 64-bit values
(apparently longer values clamp to ULLONG_MAX).

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1935,7 +1935,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         field SHOULD be "-". The value of the
                         &lt;sess-id&gt; field SHOULD be a
                         cryptographically random number. To ensure
-                        uniqueness, this number SHOULD be at least 64
+                        uniqueness, this number SHOULD be 63
                         bits long. The value of the &lt;sess-version&gt;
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1935,7 +1935,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         field SHOULD be "-". The value of the
                         &lt;sess-id&gt; field SHOULD be a
                         cryptographically random number. To ensure
-                        uniqueness, this number SHOULD be at least 64
+                        uniqueness, this number SHOULD be 63
```
ekr: I would write this as "this number should be a 64-bit integer with the bottom 63 bits randomly generated and the top bit set to zero"

fluffy:
A non negative singed integer with at least 63 bits of entropy ?   I happy with whatever you and EKR decide here and will r+ on the basis I will like what happens

PR #304, opened 2016-08-23
remove empty set of things changed by options and by SDP munging

### LINKED ISSUE #302: Placeholder open issue in 6. should be removed
Section 6 has a placeholder open issue:

o  [[OPEN ISSUE: This is a placeholder for other modifications, which
      we may continue adding as use cases appear.]]

this should be removed before WGLC

fluffy:
Fixes #302

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3109,11 +3109,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                 capabilities in the offer it sends to the far side
                 (post-setLocalDescription) or the
                 offer that it installs from the far side (pre-setRemoteDescription),
-                as long as it retmains a
+                as long as it remains a
                 valid SDP offer and specifies a subset of what was in the
                 original offer. This is safe because the answer is not
                 permitted to expand capabilities, and therefore will just
                 respond to what is present in the offer.</t>
+
+                <t>The application SHOULD NOT modify the SDP in the answer it
+                transmits, as the answer contains the negotiated capabilities,
+                and this can cause the two sides to have different
+                ideas about what exactly was negotiated.</t>
 
                 <t>As always, the application is solely responsible for
                 what it sends to the other party, and all incoming SDP will
@@ -3629,7 +3634,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
                 <t>First, the parsed parameters are checked to ensure
                 that they have not been altered after their generation in
-                createOffer, as discussed in
+                createOffer/createAnswer, as discussed in
                 <xref target="sec.modifying-sdp" />; otherwise,
                 processing MUST stop and an error MUST be returned.</t>
 
@@ -3876,7 +3881,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         <t>[TODO: handling of CN, telephone-event,
                         "red"]</t>
 
-                        <t>If the media section if of type audio:
+                        <t>If the media section is of type audio:
                         <list style="symbols">
 
                             <t>For any specified "ptime" value,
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -4018,24 +4018,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-            the endpoint can encode):</t>
-
-            <t>
-                <list style="symbols">
-
-                    <t>[[OPEN ISSUE: This is a placeholder for other
-                    modifications, which we may continue adding as use
-                    cases appear.]]</t>
+                  <t>Remove or reorder codecs (m=)</t>
+
```
ekr: I think this mostly does concur with what we agreed in the meeting

juberti: My understanding was that we really decided we wanted to remove all such munging (see https://docs.google.com/presentation/d/1pP4LDQQSyVDUTa20R3Zgmb9PXZJPPcXiSzFOYw64dKc/edit#slide=id.g1274a0be3c_0_258), in favor of real APIs.

Having this one exception seems to be of questionable value. I would recommend that we simply say that the description MUST NOT be modified between createOffer and setLocal; then this whole section can be simplified.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1262,7 +1262,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     reflect the current state of the system. Because
                     this method may need to inspect the system state to
                     determine the currently available resources, it may
-                    need to be implemented as an async operation.</t>
+                    need to be implemented as an async operation.
+		    The SDP returned from createAnswer MUST NOT be changed
+		    before passing it to setLocalDescription.</t>
```
ekr: Remove this and make the section below say "createOffer/createAnswer"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3963,92 +3965,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         <section title="Configurable SDP Parameters"
```
ekr: I would rename this section "Modifying SDP"

juberti:
See outstanding comment from @ekr

juberti:
Moved this section from its own Section 6 to a new section 5.4, and fixed up the xrefs accordingly.

PTAL

fluffy:
@ekr - have a quick review of this. I think you will be fine with justin's changes so just fix what you want and merge

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3096,6 +3097,33 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                NOT be changed before passing it to
+                setLocalDescription. If precise control over the SDP
+                is needed, the aformentioned createOffer/createAnswer options
+                or RTPSender APIs MUST be used.</t>
+
+                <t>Note that the application MAY modify the SDP to reduce the
+                capabilities in the offer it sends to the far side
+                (post-setLocalDescription) or the
+                offer that it installs from the far side (pre-setRemoteDescription),
+                as long as it retmains a
```
ekr: retains

ekr: I notice you just say offer here, but does this not apply to the answer as well?

I assume the argument here is that you can't modify the answer because the offerer uses that to determine what capabilities it must keep alive? If so, perhaps we should say that explicitly. Something like
"Applications MUST NOT modify the answer because it contains the answerer's representation to the offerer of what it intends to do."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3600,9 +3628,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                 level to apply a local description.</t>
 
                 <t>First, the parsed parameters are checked to ensure
-                that any modifications performed fall within those
-                explicitly permitted by <xref
-                target="sec.configurable-sdp-paramete" />; otherwise,
+                that they have not been altered after their generation in
+                createOffer, as discussed in
```
ekr: Why just createOffer? This is the whole sLD section.

juberti:
The text you mention was mostly previous text, but I agreed with what you suggested, and added a new paragraph accordingly.

PR #308, opened 2016-09-02
Remove spurious text about SetRD. Fixes #301

### LINKED ISSUE #301: Open Issue in 5.2.1
The current draft suggests moving the discussion of setRemoteDescription:

[[OPEN ISSUE: move
   discussion of setRemoteDescription to the subsequent-offer section.]]

ekr:


fluffy:
LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1984,13 +1984,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     specified in <xref target="RFC4566" /> Section
                     5.14. An m= section is generated for each
                     RtpTransceiver that has been added to the
-                    PeerConnection via the addTrack, addTransceiver, and
-                    setRemoteDescription methods. [[OPEN ISSUE: move
-                    discussion of setRemoteDescription to the
-                    subsequent-offer section.]] This is done in the
+                    PeerConnection via the addTrack or addTransceiver
```
juberti: The issue here is that it's unclear what to do with transceivers that have been added via setRemoteDescription.

To address this, we either need to remove 'via the addTrack or addTransceiver methods', or add some text around line 2565 indicating that m= sections should also be generated for transceivers that have been added via setRD.

I think the former is sufficient (and obviously simpler).

ekr:
WFM. @juberti PTAL

PR #309, opened 2016-09-02
Take 2 to fix SDP version stuff

### LINKED ISSUE #239: When is the SDP sess-version in o= line updated

fluffy:
Fixes #239

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2345,7 +2345,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
                     <t>The fields of the "o=" line MUST stay the
                     same except for the &lt;session-version&gt;
-                    field, which MUST increment on each call to
+                    field, which MUST increment by one on each call to
                     createOffer if the offer might differ
                     from the output of the previous call to createOffer;
                     implementations MAY opt to increment
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2344,15 +2344,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
                         <t>The fields of the "o=" line MUST stay the
                         same except for the &lt;session-version&gt;
-                        field, which MUST increment for each call to
-                        the createOffer() API. The only actions which
-                        causes the &lt;session-version&gt; to
-                        increment is a call to the createOffer()
-                        API. If the Javascript has provided an offer
+                        field,
+			which MUST be incremented on each call to the
+			createOffer method which would result in a
+			different offer from the current local
+			description. Implementations MAY opt to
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2344,9 +2344,19 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                        increment is a call to the createOffer()
+                        API. If the Javascript has provided an offer
+                        to a remote side that might be confused by
+                        receiving different SDP with the same
+                        &lt;session-version&gt;, it should ensure that
+                        the createOffer() API is used to create the
+                        SDP such that all &lt;session-version&gt; of
+                        the SDP that the remote side will see have a
+                        unique &lt;session-version&gt;.</t>
+
```
ekr: Regrettably, this is not what we agreed:

"V is the version in the o=line in last call to setLocal
o=V+1 on CreateOffer() if if anything changes 
"

ekr:
From meeting minutes
"SDP o= line increment [#239]
While the proposed solution on the slide was accepted, it was noted that the proposal on the slide goes against what was agreed at last meeting. Emil Ivov also commented that Trickle ICE for SIP does exactly what is described on the slide. Jonathan Lennox said that we should make sure this is compatible with whatever is done in trickle ICE SIP stacks."

ekr:
This algorithm becomes correct once we decide to not allow o= modification. LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2344,9 +2344,19 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
                         <t>The fields of the "o=" line MUST stay the
                         same except for the &lt;session-version&gt;
-                        field, which MUST increment if the session
-                        description changes in any way, including the
-                        addition of ICE candidates.</t>
+                        field, which MUST increment for each call to
+                        the createOffer() API. The only actions which
```
ekr: I think actually:
Which MUST be incremented on each call to the createOffer method which would result in a different offer from the current local description. Implementations MAY
opt to increment the "o=" line on every call to createOffer.

juberti:
Discussion in-room:
1) add example to make it clear that there is a separate counter (other than current o=) which is what increments.
2) change text about &lt;session-version&gt; to be more of a "note that" for an edge case, as opposed to normative text about a mainline case.

juberti:
Updated this branch. PTAL

fluffy:
LGTM

ekr:
2349: "might differ" is creepy. "differs" perhaps.

2359: "N + 2"  -> must be greater than N + 1

This is basically OK modulo the above

PR #312, opened 2016-09-02
ICE rollback/restart text. Fixes #250

### LINKED ISSUE #250: Need to explain how to roll back an ICE restart.
The "Applying A Local Description" section says that when applying a local description with changed ufrag/password, you perform an ICE restart as defined in RFC 5245 Section 9.1.1.1.

RFC 5245 Section 9.1.1.1 says when doing an ICE restart, you flush all previous ICE processing states. However, what happens when you roll back the description that triggered the ICE restart? One would expect the previous ICE processing states to be restored.

However, it's not as simple as just caching the previous check lists and restoring them, because events that affect the old check lists could have occurred in between setting the local description and performing the rollback. For example, receiving a response to a connectivity check. So really, we need to redefine the ICE restart rules to say - instead of just flushing the check list states - keep the old check lists around and continue updating them, and make _new_ check lists for the new candidates. The old check lists can be thrown away when the signaling state is stable (when an answer has been applied).

We should also clarify what happens if you receive a trickled candidate or "end-of-candidates" while in this intermediate stage where two checklists exist for each media section. I think it should only apply to the old checklist, because if you receive a candidate before receiving an answer, you can infer that the remote endpoint has not yet set an answer beginning an ICE restart on its side.

ekr:


ekr:
@juberti: removed the dangerous text.

fluffy:
Stil LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3732,11 +3736,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                 MUST be returned.
                 <list style="symbols">
 
-                    <t>If the description is of type "offer", and the
-                    ICE ufrag or password changed from the previous
-                    remote description, as described in Section 9.1.1.1
-                    of <xref target="RFC5245" />, mark that an ICE
-                    restart is needed.</t>
+                    <t>If the ICE ufrag or password changed from the previous
+                    remote description, then an ICE restart is needed,
```
juberti: The answer processing here seems kind of confusing. If the offer changed the username/password, an ICE restart is already in the works (new gathering etc), so saying that one is needed doesn't really make sense.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3732,11 +3736,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                    as described in Section 9.1.1.1
+                    of <xref target="RFC5245" />
+                    If the description is of type "offer", mark that an ICE
+                    restart is needed. If the description is of type
+                    "answer" and the current local description is also
+                    an ICE restart, then signal the ICE agent to begin
+                    checks as described in Section 9.3.1.1 of <xref target="RFC5245" />.
+                    It is an error for an answer to change the ufrag
```
juberti: Agreed. I assume it is also an error for the answer to not change the ufrag and password if the offer did. 

Also, should this use MUST etc instead of "It is an error"?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3619,10 +3619,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
                     <t>Or, if the ICE ufrag and password values have
                     changed, trigger the ICE Agent to start an ICE
-                    restart and begin gathering new candidates for the
-                    media section, as defined in <xref target="RFC5245"
-                    />, Section 9.1.1.1, unless it has been marked as
-                    bundle-only.</t>
+                    restart and, unless it has been marked as
+                    bundle-only, begin gathering new candidates for the
+                    media section as described in <xref target="RFC5245"
+                    />, Section 9.1.1.1, and if this description is an
```
juberti: Run-on sentence. Suggest splitting the answer processing out into its own sentence.

draft-ietf-rtcweb-jsep.xml
ekr: That was actually previous text.

draft-ietf-rtcweb-jsep.xml
ekr: Done

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3619,10 +3619,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
                     <t>Or, if the ICE ufrag and password values have
                     changed, trigger the ICE Agent to start an ICE
-                    restart and begin gathering new candidates for the
-                    media section, as defined in <xref target="RFC5245"
-                    />, Section 9.1.1.1, unless it has been marked as
-                    bundle-only.</t>
+                    restart and, unless it has been marked as
```
juberti: My original comment was unclear - I was trying to suggest that we probably want to skip any ICE restart processing (either offer or answer) if this section is bundle-only.

e.g. "Or, if the ICE ufrag and password values have changed, and it has not been marked as bundle-only, begin gathering..."

May also want to say "also start checks on that media section" to make it clear that starting checks is in addition to gathering.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3618,12 +3618,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     been marked as bundle-only.</t>
 
                     <t>Or, if the ICE ufrag and password values have
-                    changed, trigger the ICE Agent to start an ICE
-                    restart and, unless it has been marked as
-                    bundle-only, begin gathering new candidates for the
+                    changed, and it has not been marked as bundle-only,
+                    trigger the ICE Agent to start an ICE,
```
juberti: left out 'restart, and'

PR #314, opened 2016-09-02
Remove mapping table for direction attributes. Fixes #292

### LINKED ISSUE #292: Revise S 5.2.4 to refer to transceiver.direction, instead of sender/receiver.active
See https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-14#section-5.2.4

ekr:


juberti:
Can you separate this from the ICE changes that are also in this PR?

ekr:
Oops, yeah

ekr:
@juberti PTAL

juberti:
LGTM

PR #318, opened 2016-09-02
Checkpoint: start of splitting out bundled versus non-bundle a= lines

ekr:
@juberti does this kind of restructuring look good?


ekr:
Replaced

PR #319, opened 2016-09-02
Adding text for mux-only 

### LINKED ISSUE #297: Say something about a=rtcp-mux-only
JSEP should probably point at the draft and include the attribute on every (applicable) media section.

fluffy:
Fixes #297

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2897,7 +2897,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
 			 <t>If present in the offer, an
 			 "a=rtcp-mux-only" line, as specified in
-			 target="I-D.ietf-mmusic-mux-exclusive" />,
+			 <xref target="I-D.ietf-mmusic-mux-exclusive" />,
 			 Section 4.3.</t>
 
                         <t>If present in the offer, an "a=rtcp-rsize"
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2586,10 +2586,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         <t>The "a=rtcp-mux" line MUST only be added if
                         present in the most recent answer.</t>
 
-                        <t>An "a=rtcp-mux-only" line, as specified in <xref
-                        target="I-D.ietf-mmusic-mux-exclusive" /> Section 4, if
-                        and only if it was present in the most recent
-                        answer.</t>
+                        <t>The "a=rtcp-mux-only" line MUST only be added if
+                        present in the most recent answer.</t>
 
                         <t>The "a=rtcp-rsize" line MUST only be added
                         if present in the most recent answer.</t>
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2210,10 +2210,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         <t>An "a=rtcp-mux" line, as specified in <xref
                         target="RFC5761"></xref>, Section 5.1.1.</t>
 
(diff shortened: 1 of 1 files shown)


juberti:
Line 2589: "I don't think this is quite right. I imagine this should probably echo the above, e.g. MUST only be added if present in the most recent answer."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2581,6 +2586,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         <t>The "a=rtcp-mux" line MUST only be added if
                         present in the most recent answer.</t>
 
+			<t>An "a=rtcp-mux-only" line, as specified in
```
juberti: I suppose there are two cases here:
1) other side knows about mux-only: then the answer should always have it, and the 'if in most recent answer' text is fine.
2) other side doesn't know about mux-only: then there's no point in including it, and the text is also fine.

If the remote side doesn't support rtcp-mux, of course, things will blow up, but that's already been accounted for elsewhere.

fluffy: Hmm, the more I think about this, I think we should say iff the policy is require, and the other side does not have mux-only in it's SDP (both answer and offer), then we generate an error of some type that rejects the SDP and incompatible. I'd much rather have failure with an explicit reason than just the circuit breakers fail down the road or something

juberti: I don't think we want to fail things that use rtcp-mux but not rtcp-mux-only. That's worse than the current situation.

I agree we should fail if the policy is set to require but the other side doesn't say rtcp-mux; this is covered in 4.1.1, but could be made more clear in Section 5 discussion of handling a remote description.

fluffy: Read section 4.1.1. and I think it is clear enough.

fluffy:
So re-review this one and see if you are OK with it as is.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2210,6 +2210,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         <t>An "a=rtcp-mux" line, as specified in <xref
                         target="RFC5761"></xref>, Section 5.1.1.</t>
 
+                        <t>An "a=rtcp-mux-only" line, as specified in
+                        <xref target="I-D.ietf-mmusic-mux-exclusive"
+                        /> Section 4, if and only if, the RTP/RTCP
```
ekr: I don't think you want a comma after "only if"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2581,6 +2586,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         <t>The "a=rtcp-mux" line MUST only be added if
                         present in the most recent answer.</t>
 
+			            <t>An "a=rtcp-mux-only" line, as specified in <xref
+			            target="I-D.ietf-mmusic-mux-exclusive" /> Section 4, if
+			            and only if, it was present in the most recent answer
```
ekr: see above re "if and only if"

ekr:
This seems fine.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2581,6 +2586,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         <t>The "a=rtcp-mux" line MUST only be added if
                         present in the most recent answer.</t>
 
+                        <t>An "a=rtcp-mux-only" line, as specified in <xref
+                        target="I-D.ietf-mmusic-mux-exclusive" /> Section 4, if
+                        and only if it was present in the most recent answer
+                        and the RTP/RTCP multiplexing policy is "require".</t>
```
juberti: I think line 2592 is probably superfluous here; the only way one would have this in the most recent answer is if this condition were already true.

PR #324, opened 2016-09-09
Clarify that you can't have a candidate that doesn't correspond to a …

### LINKED ISSUE #251: Need to clarify behavior of addIceCandidate around an ICE restart.
Consider the following sequence of actions:

setRemoteDescription(offer_with_ice_restart);
addIceCandidate(candidate);
setLocalDescription(answer_with_ice_restart);

As issue #250 describes, when the remote peer has set an offer with an ICE restart, but has not yet set an answer, it's doing ICE processing for both the new and old "generations" simultaneously (assuming we go that route). So, is this a candidate for the new or old generation? If "old", then it can be applied immediately. Otherwise, it should be queued and only applied after setting the local description, since the local description is what triggers the ICE restart.

Of course, there's a separate issue, which is that there's no way of telling which "generation" the candidate applies to. One proposed way of fixing this is to trickle a ufrag/password along with the candidate.

Another way would be to queue the candidates from the "old generation" before signaling them, and only signal them after rolling back the local description. If instead of rolling back the local description, an answer is applied, these queued old candidates would be discarded. This way, once a candidate is signaled, it unambiguously applies to only the newest generation.

Note that the "candidate" could also be an end-of-candidates indication, and the same reasoning applies.

ekr:
…sane peer. Fixes #251.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1764,14 +1764,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     successfully, will be added to the current and/or
                     pending remote description according to the rules
                     defined for Trickle ICE.
-                    The ufrag is used to determine which
-                    remote description the candidate corresponds to.
-                    If the ufrag does not match either remote description
-                    an error is generated.
                     If the MID, m-line index, or candidate string provided in the ICE candidate is
                     invalid, an error is generated.
+                    If present, ufrag is used to determine which
+                    generation of remote description the candidate corresponds to;
+                    this resolves ambiguities when trickle ICE is in progress
+                    during an ICE restart. If the ufrag is present and
+                    does not match either remote description an error is generated.
+                    If ufrag is absent, the candidate MUST be assumed to correspond to
+                    the most recently applied remote description.
                     Connectivity checks will
-                    be sent to the new candidate.</t>
+                    be sent to the new candidate.
+                    </t>
 
                     <t>This method can also be used to provide an
                     end-of-candidates indication (as defined in <xref
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1767,11 +1767,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     If the MID, m-line index, or candidate string provided in the ICE candidate is
                     invalid, an error is generated.
                     If present, ufrag is used to determine which
-                    generation of ICE candidates the candidate corresponds to;
+                    generation of ICE candidates the candidate belongs to;
                     this resolves ambiguities when trickle ICE is in progress
                     during an ICE restart. If the ufrag is present and
-                    does not match any remote description an error is generated.
-                    If ufrag is absent, the candidate MUST be assumed to correspond to
+                    does not match any remote description, an error is generated.
+                    If the ufrag is absent, the candidate MUST be assumed to belong to
                     the most recently applied remote description.
                     Connectivity checks will
                     be sent to the new candidate.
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1764,6 +1764,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     successfully, will be added to the current and/or
                     pending remote description according to the rules
                     defined for Trickle ICE.
+                    The ufrag is used to determine which
```
juberti: A few things:
- since the ufrag is often per-mid, I think the ufrag check needs to come after the MID check on line 1771. 
- Not all candidates will have ufrags, so need to mention that ufrag is only checked if present
- It's not obvious when the ufrag is needed to resolve ambiguity, so a brief discussion of this case would be helpful

ekr:
@juberti PTAL.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1766,8 +1766,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     defined for Trickle ICE.
                     If the MID, m-line index, or candidate string provided in the ICE candidate is
                     invalid, an error is generated.
+                    If present, ufrag is used to determine which
+                    generation of remote description the candidate corresponds to;
```
taylor-b: "Generation of remote description" or "generation of ICE candidates"? Multiple remote descriptions could have the same ufrag/password.

Also, a small side question: is an ICE "generation" defined anywhere? I've been using the term informally for a while, but I can't find an official definition of it; only one mention in Trickle ICE.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1766,8 +1766,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                    this resolves ambiguities when trickle ICE is in progress
+                    during an ICE restart. If the ufrag is present and
+                    does not match either remote description an error is generated.
```
taylor-b: nit: either -> any

draft-ietf-rtcweb-jsep.xml
ekr: No, I don't think so.

Sure, "generation of ICE candidates" seems fine.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1766,8 +1766,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     defined for Trickle ICE.
                     If the MID, m-line index, or candidate string provided in the ICE candidate is
                     invalid, an error is generated.
+                    If present, ufrag is used to determine which
+                    generation of ICE candidates the candidate corresponds to;
```
juberti: corresponds to -> belongs to

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1766,8 +1766,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                    this resolves ambiguities when trickle ICE is in progress
+                    during an ICE restart. If the ufrag is present and
+                    does not match any remote description an error is generated.
```
juberti: comma before 'an'

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1766,8 +1766,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                    If ufrag is absent, the candidate MUST be assumed to correspond to
```
juberti: If -> If the
correspond -> belong

Issue #323, opened 2016-09-04
Section 5.8: Document handling of invalid PTs in answer


juberti:
https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-15#section-5.8

3. If any indicated payload type is unknown, it MUST be ignored. [TODO: should fail on answers]

(q: does this logic need to also include processing for other lines - extmap, rtcp-fb, etc - that weren't present in the offer?)

juberti:
Fixed by #326

Issue #322, opened 2016-09-04
Section 5.8: Define PT->media mapping


juberti:
https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-15#section-5.8

2) * For each specified payload type that is also supported by the
local implementation, establish a mapping between the payload
type ID and the actual media format. [TODO - Justin to add
more to explain mapping.]

juberti:
Relevant to #223 as well

juberti:
Fixed by #326

PR #326, opened 2016-09-16
Discuss header extensions in more detail

### LINKED ISSUE #321: Section 5.8: handle header extensions
https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-15#section-5.8

1) If the media section proto value indicates use of RTP:
* [TODO: header extensions]

juberti:
- Talk about what to do with header extensions in a remote description
- Expand discussion of how to handle payload types in a remote
description
- Fail things we don’t know about in an answer

Fixes #321, #322, #323

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3788,8 +3788,6 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     RTP:
                     <list style="symbols">
 
-                        <t>[TODO: header extensions]</t>
-
                         <t>If the m= section is being recycled (see
                         <xref target="sec.subsequent-offers"></xref>),
                         dissociate the currently associated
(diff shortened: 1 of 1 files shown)


fluffy:
LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3826,23 +3824,38 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                        type is unknown, it MUST be ignored.  [TODO:
-                        should fail on answers]</t>
+                        a mapping between the specified payload type ID and the
+                        actual media format, as described in
+                        <xref target="RFC3264" />, Section 6.1.
+                        Specifically, this means that the
+                        implementation determines the payload type ID to
+                        be used in outgoing RTP packets when sending each
+                        specified payload type. If any indicated payload
+                        type is unknown, it MUST be ignored.</t>
```
ekr: I'm not sure I understand where you are going with this. It seems like there are two salient properties:

1. The sender of the SDP announces which PTs it wants to use and the associated formats.
2. The answerer also announces which formats it is actually willing to use.

So you need to create:
1. A mapping from the formats you want to use -> PTs you announced for outgoing
2. A mapping from the PTs the other side announced -> the formats they were announced with.

Am I confused?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3826,23 +3824,38 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                        <t>For each specified RTP header extension that is also
+                        supported by the local implementation, establish a
+                        mapping between the extension ID and
+                        URI, as described in <xref
+                        target="RFC5285" />, Section 5. Specifically, this means
+                        that the implementation determines the extension ID to
+                        be used in outgoing RTP packets when sending each
+                        specified header extension, and enables that extension
+                        for the outgoing streams associated with this m= section.
+                        If any indicated RTP header
```
ekr: I think it's "implementation determines" that is tripping me up here and above. It's not "implementation chosen". Rather, it is determined by creating a mapping between the two and then using that mapping, right?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3930,6 +3943,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     <t>If the media section proto value indicates use
                     of RTP:
                     <list style="symbols">
+                        <t>If the media section references any payload types,
+                        RTP header extensions, or RTCP feedback mechanisms that
+                        are not known to the local implementation, this
+                        indicates a negotiation malfunction and MUST
+                        result in an error.</t>
```
ekr: Not just not known, not offered, right?

juberti: Agreed.

juberti: Test new conversation

juberti: ah ok

draft-ietf-rtcweb-jsep.xml
juberti: This section is specifically about handling a remote description, so only mapping #2 is in scope here, right?

draft-ietf-rtcweb-jsep.xml
juberti: How about 'the implementation, for each specified header extension, stores the specified extension ID, to be used when sending the header extension  in RTP packets'

juberti: Note that I think the text around 'enabling' needs to move down to the 'handling an answer' section, since these things are only really enabled if the answer says so.

juberti:
@ekr, PTAL at my responses

draft-ietf-rtcweb-jsep.xml
fluffy: How would it get a unknown payload type. This is the answer returned a "unknown" codec type in the "a=rtpmap" line ?

juberti: @fluffy, yes

Issue #325, opened 2016-09-15
Should JSEP talk about generating a mid if missing?


adam-be:
In the WebRTC 1.0 spec we have text talking about generating a mid if the remote description is missing one. That text has a faulty link to a JSEP (as described in issue w3c/webrtc-pc/issues/698). I can't seem to find any text in JSEP about this. Should there be such text? I did find text that said that an answer only should have a mid if the offer did.

juberti:
OK, I agree we need some additional text in 3.5.2.1 to indicate that if a MID is not negotiated in SDP, a MID is locally generated (but not included in SDP).

juberti:
Fixed by #327

adam-be:
Thank you!

PR #327, opened 2016-09-16
Discuss mid-less scenarios

juberti:


juberti:
Fixes #325

fluffy:
LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -631,10 +631,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         target="RFC5888"></xref>, Section 4, to identify
                         the m= line. JSEP implementations creating an
                         ICE Candidate object MUST populate both of these
-                        fields. Implementations receiving an ICE
+                        fields, using the MID of the associated RtpTransceiver
+                        object (which may be locally generated when interacting
+                        with a remote non-JSEP endpoint
+                        that does not support the MID attribute).
+                        Implementations receiving an ICE
```
ekr: This seems almost right. Maybe ("which may be locally generated by the answerer") and then link to the section below where you explain that.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -631,10 +631,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
                         Candidate object MUST use the MID if present, or
-                        the m= line index, if not (as it could have come
-                        from a non-JSEP endpoint).</t>
+                        the m= line index, if not (as in the aforementioned
+                        case).
```
ekr: I think I would spell this out "(e.g., when interacting with a non-JSEP remote endpoint)"

draft-ietf-rtcweb-jsep.xml
juberti: done

draft-ietf-rtcweb-jsep.xml
juberti: done

juberti:
PTAL

PR #328, opened 2016-09-16
Clarify max-bundle. Fixes #307

### LINKED ISSUE #307: Clarify max-bundle algorithm
From Berlin:
"A clarification to that rule was found when an offer is received with some m= lines bundled and some not; you accept all m= lines that are possible to bundle with the first m= line.
"

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1165,10 +1165,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         bundle-only. This policy aims to minimize
                         candidate gathering and maximize multiplexing,
                         at the cost of less compatibility with legacy
-                        endpoints. When acting as answerer, if there
-                        if no bundle group in the offer, the
-                        implementation will reject all but the first
-                        m= section.</t>
+                        endpoints. When acting as answerer, the implementation
+                        will accept only the compatible m= sections in the first bundle
```
juberti: 'compatible' is a new concept here; you don't discuss whether the first section is compatible in the non-bundle case.

I guess you could handle this in 3 different ways:
1. "only the first m= section, if compatible"
2. flip the text to discuss rejection, e.g. the impl will reject all bundle groups except the first, or if no bundle groups, all but the first m= section
3. change the text to say 'consider' or some such, e.g. the impl will consider only the m= sections in the first bundle group, or... only the first m= section.

I think I lean toward 2).

ekr: Let's do #2.

ekr:
@juberti @fluffy PTAL

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1165,10 +1165,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         bundle-only. This policy aims to minimize
                         candidate gathering and maximize multiplexing,
                         at the cost of less compatibility with legacy
-                        endpoints. When acting as answerer, the implementation
-                        will accept only the compatible m= sections in the first bundle
-                        group, or, if the offer did not bundle any m=
-                        sections, only the first m= section.</t>
+                        endpoints. When acting as answerer, if there
+                        if no bundle group in the offer, the
```
taylor-b: if -> is

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1165,10 +1165,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                        implementation will reject all but the first
+                        m= section and any other m= sections in the same
```
taylor-b: This mention of a bundle group here seems confusing (since earlier in the sentence we have "if there is no bundle group in the offer"). What you had earlier was easier to follow in my opinion: "If there's a bundle group, do this, if there's no bundle group, do this".

juberti: agree

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1165,10 +1165,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                         bundle-only. This policy aims to minimize
                         candidate gathering and maximize multiplexing,
                         at the cost of less compatibility with legacy
-                        endpoints. When acting as answerer, if there
-                        if no bundle group in the offer, the
-                        implementation will reject all but the first
-                        m= section.</t>
+                        endpoints. When acting as answerer, the implementation
+                        will reject all but the first m= section and those
```
juberti: how about

"will reject any m= sections other than the first m= section, unless they are in the same bundle group as that m= section."

or 

"will reject any m= sections other than the first m= section, unless they are bundled into that m= section."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2784,13 +2783,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
                       <t>No supported codec is present in the offer.</t>
 
-                      <t>The bundle policy is "max-bundle", the m=
-                      section is not in a bundle group, and this is not
-                      the first m= section.</t>
+                      <t>The bundle policy is "max-bundle", this is
```
juberti: incomplete sentence - suggest "this is" -> "and this is"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2784,13 +2783,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                      not the first m= section or in the same
+                      bundle group as the first m= section.</t>
 
-                      <t>The bundle policy is "balanced", the m=
-                      section is not in a bundle group, and this is not
-                      the first m= section for this media type.</t>
+                      <t>The bundle policy is "balanced", this is not
```
juberti: see above

PR #336, opened 2016-10-07
Add Justin's addTrack text from the list. Fixes #288

### LINKED ISSUE #288: Review addTrack track assignment algorithm
> The track will be attached to the first compatible transceiver
> (of the same media type) which has never had a direction of
> "sendonly" or "sendrecv". If no such transceiver exists,
> then one will be constructed as described in
> <xref target="sec.addTransceiver"/>.

The reuse of 'recvonly'/'inactive' m= lines may not be correct, as this would require renegotiation of the various parameters of an already established m= line. But this may affect the W3C spec too.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1217,11 +1217,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
                     addTrack attempts to minimize the number of transceivers as follows:
-                    The track will be attached to the first compatible transceiver
-                    (of the same media type) which has never had a direction of
-                    "sendonly" or "sendrecv". If no such transceiver exists,
-                    then one will be constructed as described in
-                    <xref target="sec.addTransceiver"/>.
+                    If the PeerConnection is in the have-remote-offer
+                    state, the track will be attached to the first compatible transceiver
+                    that was created by the most recent call to setRemoteDescription() and does not
+                    have a local track. Otherwise, a new transceiver will be created.
```
juberti: need a comma instead of a period

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1217,11 +1217,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                     tracks in the same MediaStream, so that they can be added to the
                     same "LS" group when creating an offer or answer.
                     addTrack attempts to minimize the number of transceivers as follows:
-                    The track will be attached to the first compatible transceiver
-                    (of the same media type) which has never had a direction of
-                    "sendonly" or "sendrecv". If no such transceiver exists,
-                    then one will be constructed as described in
-                    <xref target="sec.addTransceiver"/>.
+                    If the PeerConnection is in the have-remote-offer
```
juberti: add quotes around "have-remote-offer"

Issue #261, opened 2016-03-22
Need to document API points for RtpSender/RtpReceiver


juberti:
In section 4, Interface, the PC methods are listed but not the sender/receiver/transceiver methods.

juberti:
I think we need to call out:
- PC.addTrack
- RtpSender.active, .mid
- RtpReceiver.active, .mid
- RtpTransceiver.mid, stop(). Maybe .sender, .receiver

taylor-b:
Should JSEP also describe when properties of an RtpSender/RtpReceiver (transport and rtcpTransport for instance) are updated? Presumably in "applying a local/remote description"?

juberti:
I'd prefer not to have to clone everything from webrtc-pc into JSEP. Properties that aren't relevant to SDP can hopefully be omitted.

juberti:
Eric added placeholders for addTrack, addTransceiver, createDataChannel today.

ekr:
@juberti I have written the above sections (and in any case there are placeholders and associated GH issues). Can we close this issue?

juberti:
I think we still need .stop for the transceiver, and maybe the .mid and .active getters.

pthatcherg:
An RtpSender/RtpReceiver doesn't have an .active or .mid, so I don't think we need to add anything for them. 

RtpTransceiver has .stop(), which makes the port 0.  I'll add that. 
RtpTransceiver also has .setDirection, which changes the m= line direction.  I'll add that.

I think that's all we need.

pthatcherg:
Oh, there's a big one we forgot: setCodecPreferences.

PR #338, opened 2016-10-07
Add RtpTransceiver methods stop, stopped, setDirection, and setCodecPreferences.

pthatcherg:
Fix https://github.com/rtcweb-wg/jsep/issues/261


ekr:
@pthatcherg: needs rebase

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1097,6 +1097,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           simulcast as described in 
           <xref target="sec.simulcast" />).</t>
         </section>
+
+        <section title="stopTransceiver" anchor="sec.transceiver-stop">
+          <t>
+            The stopTransceiver method stops an RtpTransceiver.
+            See <xref target="sec.transceiver-stopped"/> for more details.
```
ekr: "see below" seems more appropriate when this is the next section

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1097,6 +1097,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          </t>
+        </section>
+        <section title="isTransceiverStopped" anchor="sec.transceiver-stopped">
```
ekr: Did we decide that to do methodFoo() or methodFoo

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1097,6 +1097,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          <t>
+            The isTransceiverStopped method returns true if
```
ekr: "true" and "false" need to be quoted

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1097,6 +1097,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            the transceiver has been stopped, either by a call
+            to stopTransceiver or by a call to
+            setRemoteDescription, and false otherwise.
+          </t>
+          <t>
+            A stopped RtpTransceiver does not send any
```
ekr: Can stopped transceivers be re-started?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1097,6 +1097,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            outgoing RTP or RTCP or process any incoming RTP
+            or RTCP.
+          </t>
+        </section>
+        <section title="setTransceiverDirection" anchor="sec.transceiver-set-direction">
+          <t>
+            The setTransceiverDirection method sets the
+            direction of a transceiver, which affects the
+            direction attribute of an associted m=section on
```
ekr: associated

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1097,6 +1097,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            createAnswer.
+          </t>
+          <t>
+            The codec preferences of an RtpTransceiver can
+            affect calls to createOffer and createAnswer by
+            either excluding codecs (in which case the
+            corresponding media formats will be excluded) or
+            by changing the order of codecs (in which case the
+            order of the media formats will match).  The codec
+            preferenes cannot add media formats that would
```
ekr: preferences.

Also, can this cause onnegotiationneeded.

draft-ietf-rtcweb-jsep.xml

pthatcherg:
That's twice now.  Stop moving stuff around!

draft-ietf-rtcweb-jsep.xml
pthatcherg: methodFoo is consistent with the existing text.

draft-ietf-rtcweb-jsep.xml
pthatcherg: True.

For some definition of "true".

draft-ietf-rtcweb-jsep.xml
pthatcherg: no

draft-ietf-rtcweb-jsep.xml
pthatcherg: done

pthatcherg: onnegotiationneeded is covered in the W3C spec.  I don't see anywhere in this spec anything that references onnegotiationneeded.   Is there one?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1097,6 +1097,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           simulcast as described in
           <xref target="sec.simulcast" />).</t>
         </section>
+
+        <section title="stopTransceiver" anchor="sec.transceiver-stop">
```
juberti: To avoid confusion, rather than jamming these on PeerConnection, I think we probably want to make a separate section for RtpTransceivers, e.g.

    4.  Interface
         4.1.  PeerConnection
           4.1.1.  Constructor
           4.1.2.  addTrack
           4.1.3.  addTransceiver
         4.2.  RtpTransceiver
           4.2.1. setDirection 
           4.2.2. stop
           4.2.3. isStopped

(we can probably remove the empty 'Methods' section)
   
...

draft-ietf-rtcweb-jsep.xml
juberti: I think we typically use parens though when referring to a method.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1097,6 +1097,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          <t>
+            The isTransceiverStopped method returns true if
+            the transceiver has been stopped, either by a call
+            to stopTransceiver or by a call to
+            setRemoteDescription, and false otherwise.
+          </t>
+          <t>
+            A stopped RtpTransceiver does not send any
+            outgoing RTP or RTCP or process any incoming RTP
+            or RTCP.
```
fluffy: Might help to add a note that stopped will correspond to a m line with a zero port and that a m-line that has a direction of "inactive" will not be stopped.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1097,6 +1097,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          <t>
+            The setCodecPreferences method sets the codec
+            preferences of a transceiver, which affect the
+            presence and order of codecs of an associated
+            m=section on future calls to createOffer and
+            createAnswer.
+          </t>
+          <t>
+            The codec preferences of an RtpTransceiver can
+            affect calls to createOffer and createAnswer by
```
fluffy: I think it can impact createOffer but I don't think it should impact createAnswer. The answer should accept the highest priority codec in the offer regardless of any priorities of this browser or the JS running on side processing the answer.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2501,6 +2551,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>The &lt;proto&gt; field MUST be set to exactly match the
             &lt;proto&gt; field for the corresponding m= line in the
             offer.</t>
+
+            <t>If codec preferences have been set for the
+            associated RtpTransceiver, media formats MUST
+            be generated in the corresponding order.</t>
```
fluffy: I'm not reading this in context but this should only be for Offer not answer. Perhaps it is OK as it.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2531,7 +2585,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>A direction attribute which is the same as that of the
             associated transceiver.</t>
 
-            <t>For each supported codec that is present in the offer,
+            <t>For each supported codec that is present in the offer
+            and not excluded by the codec preferences of associated transceiver,
```
fluffy: again offer only not answer but perhaps this is right

draft-ietf-rtcweb-jsep.xml
pthatcherg: I added a note about making a zero port.  

But what's this about inactive m= sections not being stopped?  That's new to me.

draft-ietf-rtcweb-jsep.xml
pthatcherg: Then how does an answerer specify which codec the prefer to receive?  The W3C spec specifically says "This method allows applications to disable the negotiation of specific codecs. It also allows an application to cause a remote peer to prefer the codec that appears first in the list for sending."  Would that not apply to both the offerer and the answerer?

draft-ietf-rtcweb-jsep.xml
pthatcherg: That's a bit more remodeling of the text than I wanted to do.  Can we land this as-is and then do another PR to remodel?  Or would you like to make a PR of my PR?  I have a feeling you typing what you want will be faster than you typing a description of what you want me to type.

draft-ietf-rtcweb-jsep.xml
pthatcherg: Ever place I see addTrack and addTransceiver don't have parens.  Are we supposed to change them all?

draft-ietf-rtcweb-jsep.xml
fluffy: So 3264 has the text 

   Although the answerer MAY list the formats in their desired order of
   preference, it is RECOMMENDED that unless there is a specific reason,
   the answerer list formats in the same relative order they were
   present in the offer.  In other words, if a stream in the offer lists
   audio codecs 8, 22 and 48, in that order, and the answerer only
   supports codecs 8 and 48, it is RECOMMENDED that, if the answerer has
   no reason to change it, the ordering of codecs in the answer be 8,
   48, and not 48, 8.  This helps assure that the same codec is used in
   both directions.

Which I think supports you argument that if the JS going the create answer set the a preference, then we reorder the codecs in the offer to match that preference when generating the answer

pthatcherg:
Parens added.

draft-ietf-rtcweb-jsep.xml
taylor-b: So for an inactive transceiver, isTransceiverStopped will return true? Then how do you tell the difference between an inactive port 0 transceiver and a plain old inactive one?

draft-ietf-rtcweb-jsep.xml
juberti: Also: what should happen if A offers codecs X and Y but B prefers Y over X, and does a reoffer?

I generally think this should only affect initial offers.

draft-ietf-rtcweb-jsep.xml
ekr: that was supposed to be rhetorical. I think you should change it :)

draft-ietf-rtcweb-jsep.xml
ekr: I think you might say "this may require a new offer/answer transaction"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1901,8 +1962,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>A direction attribute which is the same as that of the
               associated transceiver.</t>
 
-              <t>For each supported codec, "a=rtpmap" and "a=fmtp"
-              lines, as specified in
+              <t>For each supported codec,
+              not excluded by the codec preferences of associated transceiver,
```
juberti: For each supported codec **that has not been** excluded by the codec preferences of **the** associated transceiver

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2304,7 +2366,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
             <t>The m= line and corresponding "a=rtpmap" and "a=fmtp"
             lines MUST only include codecs present in the most recent
-            answer.</t>
+            answer which are not excluded by the codec preferences of
```
juberti: are -> have not been
of -> of the

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2501,6 +2564,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>The &lt;proto&gt; field MUST be set to exactly match the
             &lt;proto&gt; field for the corresponding m= line in the
             offer.</t>
+
+            <t>If codec preferences have been set for the
```
juberti: I think you want to remove this, per our discussion.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2531,7 +2598,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>A direction attribute which is the same as that of the
             associated transceiver.</t>
 
-            <t>For each supported codec that is present in the offer,
+            <t>For each supported codec that is present in the offer
+            and not excluded by the codec preferences of associated transceiver,
```
juberti: of the

taylor-b:
Alright, I've addressed all the comments so far (except for "transceiver with inactive direction should return true for isStopped", which I disagree with.

Most notably, we now have the rule that the codec preferences can *remove* codecs at any time, but can only *reorder* them for the initial offer. This prevents codecs being reordered back and forth on subsequent offer/answer exchanges, and prevents having different orders in the offer and answer.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1577,7 +1577,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             The isStopped method returns "true" if
             the transceiver has been stopped, either by a call
             to stopTransceiver or by a call to
-            setLocalDescription, and "false" otherwise.
+            setRemoteDescription, and "false" otherwise.
```
pthatcherg: I think there's a problem here.   An offer can't stop a transceiver, because it's possible that will be rolled back.  An answer must be the thing that can stop a transceiver.   And that would only happen if the remote side offered to stop the transceiver, which means the answer that actually stops it (in a way that can't be rolled back) is the setLocalDescription(answer), not the setRemoteDescription(offer).

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1602,14 +1602,22 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          <t>
+            The codec preferences of an RtpTransceiver can also determine
+            the order of codecs in subsequent calls to createOffer (if an
+            answer containing the transceiver has not yet been applied),
+            in which case the order of the media formats in the associated
+            m= section will match. The codec preferences cannot change the
+            order of codecs in an answer, and cannot change the order of
+            codecs after an answer containing the transceiver has been
+            applied.
           </t>
```
pthatcherg: Ah, I see you addressed the "change on re-offer" part of this by restricting it to the initial offer.  But that definitely contradicts the W3C spec, which says "Codec preferences remain in effect for all calls to createOffer and createAnswer that include this RTCRtpTransceiver until this method is called again."

If this is what we really want, then I think we need to go change it in the W3C to match.  And I don't think it's what we decided on in the W3C before.

However, I personally don't care either way.  I think anyone who wants to change what codec to send should just use RtpSender.setParameters.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1980,6 +1980,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <xref target="I-D.ietf-rtcweb-video"></xref>(see Section
               5).</t>
 
+              <t>If codec preferences have been set for the associated
+              transceiver, media formats MUST be generated in the
+              corresponding order.</t>
+
```
pthatcherg: I think it will be a very confusing API if this happens to a developer:

1.  Set the codec preferences.  Receive a call.  Codec preferences ignored.
2.  Set the codec preferences again (or mid-call).  Re-offer.  Codec preferences ignored.
3.  Learn to remove all but my favorite codec.  It works!  Until it doesn't.
4.  Receive a call from an endpoint that doesn't support my favorite codec.  No video.
5.  Learn to check RtpTransceiver.sender.getParameters().codecs first, then pick my favorite out of that group and then exclude the rest.  It works!

Still better than SDP munging?

draft-ietf-rtcweb-jsep.xml
taylor-b: Well they do remain in effect with these rules... just only for the purpose or removing codecs. :)

But I understand your reasoning.

taylor-b:
I looked through our mailing archives and meeting minutes to see what we've previously decided.

When this API was first discussed on the mailing list, [it was proposed](https://lists.w3.org/Archives/Public/public-webrtc/2015Oct/0035.html) that the preferences *do* affect `createAnswer`, and no one voiced any opposition.

Then a couple weeks later [at TPAC 2015](https://www.w3.org/2015/10/28-webrtc-minutes#item04), we decided that the order in `setCodecPreferences` affects the initial offer *and* answer, but not subsequent offers/answers.

So maybe we should just go with what was previously decided?

draft-ietf-rtcweb-jsep.xml
juberti: This is a good argument. You've convinced me we should apply this for initial offers and answers (which sounds like the decision the W3C came to).

taylor-b:
Ready for review again. Rules about codec preferences now match the decision at TPAC 2015 described above.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1562,6 +1563,74 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            The codec preferences of an RtpTransceiver can also determine
+            the order of codecs in subsequent calls to createOffer and
+            createAnswer, in which case the order of the media formats in
+            the associated m= section will match. However, the codec
+            preferences cannot change the order of the media formats after
+            an answer containing the transceiver has been applied. At this
+            point, codecs can only be removed, not reordered.
+          </t>
+          <t>
+            Note that, as described in <xref target="RFC3264">
```
juberti: This guidance is somewhat irrelevant here since we have a specific set of directives below indicating what should be sent. (i.e., the 'applying a remote description' section). 

So, you could either point at that, or omit this clause and say 'Note that the order of codecs in sCP does not affect which codec the implementation decides to send. It only affects ...'

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1901,15 +1970,20 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              the codec preferences of the associated transceiver,
+              "a=rtpmap" and "a=fmtp" lines, as specified in
               <xref target="RFC4566"></xref>, Section 6. The audio and
               video codecs that MUST be supported are specified in
               <xref target="I-D.ietf-rtcweb-audio"></xref>(see Section
               3) and
               <xref target="I-D.ietf-rtcweb-video"></xref>(see Section
               5).</t>
 
+              <t>If codec preferences have been set for the associated
```
juberti: I think this needs to go into the discussion of the m= line generation itself (e.g. the section after "Each m= section should be generated as specified in [RFC4566],
   Section 5.14.  For the m= line itself, the following rules MUST be
   followed:")

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2541,6 +2620,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <xref target="I-D.ietf-rtcweb-video"></xref>(see Section
             5).</t>
 
+            <t>If codec preferences have been set for the associated
```
juberti: I think this should go into the m= line generation discussion.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1562,6 +1563,70 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            outgoing RTP or RTCP or process any incoming RTP
+            or RTCP.  It cannot be restarted.
+          </t>
+        </section>
+        <section title="setDirection" anchor="sec.transceiver-set-direction">
+          <t>
+            The setTransceiverDirection method sets the
+            direction of a transceiver, which affects the
+            direction attribute of the associated m= section on
+            future calls to createOffer and createAnswer.
```
ekr: This seems a bit confusing. It's not like you can unilaterally change something that's recvonly to sendonly without a re-offer. 

How about a TODO for an example.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1562,6 +1563,70 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          </t>
+        </section>
+        <section title="setCodecPreferences" anchor="sec.transceiver-set-codec-preferences">
+          <t>
+            The setCodecPreferences method sets the codec
+            preferences of a transceiver, which affect the
```
ekr: affects

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1562,6 +1563,70 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            preferences cannot change the order of the media formats after
+            an answer containing the transceiver has been applied. At this
+            point, codecs can only be removed, not reordered.
+          </t>
+          <t>
+            Note that the order of codecs in setCodecPreferences does not
+            affect which codec the implementation decides to send. It only
+            affects which codecs the implementation indicates that it prefers
+            to receive, via the offer or answer.
+          </t>
```
ekr: Can you move this note up. The rest would be clearer if you understood that it only impacts new O/A transactions.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1901,8 +1970,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>A direction attribute which is the same as that of the
               associated transceiver.</t>
 
-              <t>For each supported codec, "a=rtpmap" and "a=fmtp"
-              lines, as specified in
+              <t>For each supported codec that has not been excluded by
+              the codec preferences of the associated transceiver,
+              "a=rtpmap" and "a=fmtp" lines, as specified in
```
ekr: I think you would be better if you phrased this as
"For each codec that is available for this transceiver" rather than "supported filtered by excluded"

draft-ietf-rtcweb-jsep.xml
taylor-b: Changed to "in turn affect" to make it unambiguous that this means "the preferences affect".

draft-ietf-rtcweb-jsep.xml
taylor-b: Done. Also added something clarifying that excluding a codec doesn't mean you immediately stop sending it.

draft-ietf-rtcweb-jsep.xml
taylor-b: As discussed I think this would require some restructuring (to define "codec available for a transceiver" somewhere) so we can save this for later, in the interest of landing the PR.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1861,6 +1930,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             specified in
             <xref target="I-D.nandakumar-mmusic-proto-iana-registration" /> if
             the default candidate uses TCP transport.</t>
+
+            <t>If codec preferences have been set for the associated
```
juberti: I think you need to talk about excluded codecs here too.

draft-ietf-rtcweb-jsep.xml
juberti: You might just want to say "for each media format specified on the m= line", since that should cover it.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2501,6 +2579,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>The &lt;proto&gt; field MUST be set to exactly match the
             &lt;proto&gt; field for the corresponding m= line in the
             offer.</t>
+
+            <t>If codec preferences have been set for the associated
```
juberti: See note above on excluded codecs

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2531,7 +2613,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>A direction attribute which is the same as that of the
             associated transceiver.</t>
 
-            <t>For each supported codec that is present in the offer,
+            <t>For each supported codec that is present in the offer and
```
juberti: See note above about referencing m= line formats

draft-ietf-rtcweb-jsep.xml
taylor-b: Done. Also moved "must support these mandatory codecs" up here.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1974,15 +1983,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-              the codec preferences of the associated transceiver,
+              <t>For each media format on the m= line,
               "a=rtpmap" and "a=fmtp" lines, as specified in
-              <xref target="RFC4566"></xref>, Section 6. The audio and
-              video codecs that MUST be supported are specified in
-              <xref target="I-D.ietf-rtcweb-audio"></xref>(see Section
-              3) and
-              <xref target="I-D.ietf-rtcweb-video"></xref>(see Section
-              5).</t>
+              <xref target="RFC4566"></xref>, Section 6.</t>
```
juberti: suggest mentioning 3264, S 5.1, similar to the answer section below.

PR #343, opened 2016-10-07
Reference transceiver direction as opposed to "active" senders/receivers

taylor-b:
Reference transceiver direction as opposed to "active" senders/receivers

We initially thought there would be an "active" bit on senders/receivers
to control the direction of a media description, but instead we now have
a "setDirection" API on the transceiver. So we should reference that.


pthatcherg:
lgtm

PR #344, opened 2016-10-07
Issue230 a setup

ekr:
@juberti I don't think this is quite ready, but PTAL

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2071,7 +2071,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               in the offer MUST be "actpass".</t>
 
               <t>An "a=dtls-id" line, as specified in
-              <xref target="I-D.ietf-mmusic-dtls-sdp"/> Section 4.</t>
+              <xref target="I-D.ietf-mmusic-dtls-sdp"/> Section 5.2.</t>
 
               <t>An "a=rtcp-mux" line, as specified in
               <xref target="RFC5761"></xref>, Section 5.1.1.</t>
@@ -2095,7 +2095,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           4.1.</t>
 
           <t>Within the data m= section, the "a=mid", "a=ice-ufrag",
-          "a=ice-pwd", "a=fingerprint", and "a=setup" lines MUST be
+          "a=ice-pwd", "a=fingerprint", "a=dtls-id", and "a=setup" lines MUST be
           included as mentioned above, along with an
           "a=fmtp:webrtc-datachannel" line and an "a=sctp-port" line
           referencing the SCTP port number as defined in
@@ -2639,12 +2639,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               algorithm used for the fingerprint MUST match that used
               in the certificate signature.</t>
 
+              <t>An "a=dtls-id" line, as specified in
+              <xref target="I-D.ietf-mmusic-dtls-sdp"/> Section 5.3.</t>
+
               <t>An "a=setup" line, as specified in
               <xref target="RFC4145"></xref>, Section 4, and clarified
               for use in DTLS-SRTP scenarios in
               <xref target="RFC5763"></xref>, Section 5. The role value
               in the answer MUST be "active" or "passive"; the "active"
-              role is RECOMMENDED.</t>
+              role is RECOMMENDED. The role value MUST be consistent
+              with the existing DTLS connection, if one exists and is
+              being continued.</t>
 
               <t>If present in the offer, an "a=rtcp-mux" line, as
               specified in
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2636,12 +2639,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               algorithm used for the fingerprint MUST match that used
               in the certificate signature.</t>
 
+              <t>An "a=dtls-id" line, as specified in
```
juberti: Suggest matching the offer text and putting this after a=setup

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2996,6 +3004,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <xref target="RFC4145" />, Section 4, and the setup value
             is stored.</t>
 
+            <t>If present, a single "a=dtls-id" line is parsed as
```
juberti: Is this thing valid at session level?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3199,12 +3212,20 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               the size limits specified in
               <xref target="RFC5245" />, Section 15.4.</t>
 
+              <t>dtls-id value, which MUST be set according to
+              <xref target="I-D.ietf-mmusic-dtls-sdp"/> Section 5.
+              If this is a re-offer and the dtls-id value is different
+              from that presently in use, then a new DTLS connection
+              MUST be established, and the ufrag and password values
+              MUST also be different. If processing an answer, the
```
juberti: I would spell this out, i.e., if we're doing a new DTLS connection here, it MUST be as part of a proposed ICE restart, with new ICE ufrag and password values.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3199,12 +3212,20 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
               <t>DTLS setup value, which MUST be set according to the
-              rules specified in
-              <xref target="RFC5763" />, Section 5, and MUST be
-              consistent with the selected role of the current DTLS
-              connection, if one exists.[TODO: may need revision, i.e.,
-              use of actpass</t>
+              rules specified in <xref target="RFC5763"/>,
+              Section 5 and MUST
+              be consistent with the selected role of the current DTLS
+              connection, if one exists and is being continued.</t>
```
juberti: I also think you need some text in the "applying a remote description" section below that talks about creating a new DTLS connection.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3199,12 +3212,20 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               the size limits specified in
               <xref target="RFC5245" />, Section 15.4.</t>
 
+              <t>dtls-id value, which MUST be set according to
+              <xref target="I-D.ietf-mmusic-dtls-sdp"/> Section 5.
+              If this is a re-offer and the dtls-id value is different
+              from that presently in use, then a new DTLS connection
```
juberti: Not sure if this is where you need to point out that a new DTLS connection MUST be established - that seems like something for the  "applying a remote desc" section. I think you just need to note that the previous connection is not being continued, relevant to the para below.

draft-ietf-rtcweb-jsep.xml
ekr: Thankfully, no
https://tools.ietf.org/html/draft-ietf-mmusic-dtls-sdp-14#section-4

draft-ietf-rtcweb-jsep.xml
ekr: done

ekr:
@juberti PTAL

draft-ietf-rtcweb-jsep.xml
juberti: Did you mean to remove this then?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3199,12 +3212,21 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               the size limits specified in
               <xref target="RFC5245" />, Section 15.4.</t>
 
+              <t>dtls-id value, which MUST be set according to
+              <xref target="I-D.ietf-mmusic-dtls-sdp"/> Section 5.
+              If this is a re-offer and the dtls-id value is different
+              from that presently in use, the DTLS connection is not being
+              continued and the remote description MUST be part of an
+              ICE restart, together with new ufrag and password
+              values. If processing an answer, the
```
juberti: suggest changing 'processing' to 'this is' to match text above

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3389,6 +3411,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           the end-of-candidates indication as described in
           <xref target="I-D.ietf-ice-trickle" /> Section 11.</t>
 
+          <t>If the dtls-id has changed from the previous remote description,
```
juberti: Looking at this more closely, I think you probably want to move this text to the 'applying an answer' section, which already has a para on when to create a DTLS connection.

PR #345, opened 2016-10-07
Address weird SAR cases. Fixes #310

### LINKED ISSUE #310: Support for SAR in imageattr
In https://rtcweb-wg.github.io/jsep/#rfc.section.3.6.2 we say 

For an "a=imageattr recv" attribute, only size limits are considered. Any other values, e.g. aspect ratio, MUST be ignored.

Ignoring sar seem problematic because we may get squashed video if the sender is not sending square pixels. (some video codecs can indicated the sar in the video stream but some can't)

juberti:
Fixes #310

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -665,6 +665,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         what it receives due to application preferences, e.g. a
         specific size for the window in which the video will be
         displayed.</t>
+
+        <t>Note that certain codecs support transmission of samples with aspect
+        ratios other than 1.0 (i.e., non-square pixels).
+        JSEP implementations will not transmit non-square pixels, but SHOULD
+        receive and render such video with the correct aspect ratio.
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -665,6 +665,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         what it receives due to application preferences, e.g. a
         specific size for the window in which the video will be
         displayed.</t>
+
+        <t>Note that certain codecs support transmission of samples with aspect
+        ratios other than 1.0 (i.e., non-square pixels).
+        JSEP implementations will not transmit non-square pixels, but SHOULD
+        receive and render such video with the correct aspect ratio.
```
ekr: I think this would do with a forward pointer about how you would know that.

fluffy: So if we get SDP that requires sending with non square, we should reject that ?

juberti: Sorry, not sure exactly what is needed here. What did you want to see a pointer to?

PR #347, opened 2016-10-07
Per Justin's comments, remove all description of generating

### LINKED ISSUE #149: Limitation on SSRC change is in violation with RFC7160 regarding RTP PT change
I want raise for discussion the fact that I think JSEP does not handle RTP payload type switching as currently specified.

So lets walk through a case of RTP payload type switching and what it means according to the specifications we do have.

So we have an established PeerConnection with one audio track for which we have negotiated the possibility to send both PCM and OPUS. If the sender decides to switch RTP payload type, lets say from OPUS to PCM. The reason is actually not important, but lets say due to a "replace track" API call as currently being discussed on W3C WebRTC group.

When that RTP payload type switch from OPUS to PCM we also have an RTP timestamp rate change, from 48 kHz to 8 kHz. As specified in RTP usage (Section 4.3):

   If performing changes between two RTP payload types that use
   different RTP clock rates, an RTP sender MUST follow the
   recommendations in Section 4.1 of [RFC7160].

So lets look at what RFC 7160 says:

4.1.  RTP Sender (with RTCP)

   An RTP Sender with RTCP turned on MUST use a different SSRC for each
   different clock rate.  An RTCP BYE MUST be sent and a new SSRC MUST
   be used if the clock rate switches back to a value already seen in
   the RTP stream.

Thus, this Payload type change forces the use of a new SSRC.

This in itself should not be a significant issue. The binding to the right media block in the SDP is done using MID. Which, will be sent in an RTP header extension to ensure that stream upon reception is correctly associated. The RTP synchronization information will arrive in RTCP or even in the RTP header extension if the peers both supports RFC 6051. Thus minimal media impacts is to be expected.

But, discussing what happens in this case with Stefan Håkansson we did look into JSEP. And here we clearly have some short commings in the specification.

A. First of all there is an assumption that the value of a=ssrc will not change. For example Section 5.2.2 (Subsequent Offers):

  o  For MediaStreamTracks that are still present, the "a=msid",
      "a=ssrc", and "a=ssrc-group" lines MUST stay the same.

This limitaiton can't be left in place.

ekr:
a=ssrc and a=ssrc-group. Fixes #149.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1954,30 +1954,6 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <xref target="I-D.ietf-rtcweb-rtp-usage"></xref>, Section
               5.1.</t>
 
-              <t>An "a=ssrc" line, as specified in
-              <xref target="RFC5576"></xref>, Section 4.1, indicating
-              the SSRC to be used for sending media, along with the
-              mandatory "cname" source attribute, as specified in
-              Section 6.1, indicating the CNAME for the source. The
-              CNAME MUST be generated in accordance with Section 4.9 of
-
-              <xref target="I-D.ietf-rtcweb-rtp-usage"></xref>.</t>
-
-              <t>If RTX is supported for this media type, another
-              "a=ssrc" line with the RTX SSRC, and an "a=ssrc-group"
-              line, as specified in
-              <xref target="RFC5576"></xref>, section 4.2, with
-              semantics set to "FID" and including the primary and RTX
-              SSRCs.</t>
-
-              <t>If FEC is supported for this media type, another
-              "a=ssrc" line with the FEC SSRC, and an "a=ssrc-group"
-              line with semantics set to "FEC-FR" and including the
-              primary and FEC SSRCs, as specified in
-              <xref target="RFC5956"></xref>, section 4.3. For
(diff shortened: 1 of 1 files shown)


ekr:
@juberti: note I still assume you might *receive* a=ssrc because it might come from non-JSEP endpoints. LMK if you want that removed.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2212,10 +2164,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             with an m= section, and the m= section is not being
             recycled as described above, an m= section MUST be
             generated for it with the port set to zero and the
-            "a=msid", "a=ssrc", and "a=ssrc-group" lines removed.</t>
+            "a=msid" line removed.</t>
 
             <t>For RtpTransceivers that are not stopped, the "a=msid",
```
juberti: remove superfluous comma

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2756,7 +2683,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             omitted.</t>
 
             <t>For RtpTransceivers that are not stopped, the "a=msid",
-            "a=ssrc", and "a=ssrc-group" lines MUST stay the same.</t>
+            line MUST stay the same.</t>
```
juberti: spurious comma

Issue #341, opened 2016-10-07
Need rules for combining transceiver direction with offered directional attribute.


taylor-b:
The steps for createAnswer say each m= section must contain:

> o  A direction attribute which is the same as that of the associated
      transceiver.

But this doesn't take into account the directional attribute in the offer.

Imagine the following sequence of operations:

pc.setRemoteDescription(offer_with_sendonly_msection);
transceiver.setDirection("sendrecv");
pc.createAnswer();

Even though the transceiver's direction is "sendrecv", the answer must contain "recvonly", and the application must perform another offer/answer to change the negotiated direction.

juberti:
Peter/Taylor - can you look into this before Friday (10/21)?

taylor-b:
I'm gonna start making a PR now.

juberti:
Fixed in commit 490ca75 (manual marge of an updated version of Taylor's PR).

PR #357, opened 2016-10-20
Adding back a direction attribute table for answers.

taylor-b:
Needed because the direction set on the RtpTransceiver may be outside
the bounds of what was in the remote offer.

Fixes issue #341.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1990,8 +1990,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               the dummy value "9 IN IP4 0.0.0.0", because no candidates
               have yet been gathered.</t>
 
-              <t>A direction attribute which is the same as that of the
-              associated transceiver.</t>
+              <t>A direction attribute for the associated RtpTransceiver,
+              described by <xref target="sec.answer-dirattr"></xref>.</t>
 
               <t>For each media format on the m= line,
               "a=rtpmap" and "a=fmtp" lines, as specified in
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2824,6 +2824,35 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                     |  sendrecv  |  sendonly  |  recvonly  |  inactive  |
+    +---+------------+------------+------------+------------+------------+
+    |   |  sendrecv  |  sendrecv  |  sendonly  |  recvonly  |  inactive  |
+    | o +------------+------------+------------+------------+------------+
+    | f |  sendonly  |  recvonly  |  inactive  |  recvonly  |  inactive  |
+    | f +------------+------------+------------+------------+------------+
+    | e |  recvonly  |  sendonly  |  sendonly  |  inactive  |  inactive  |
+    | r +------------+------------+------------+------------+------------+
+    |   |  inactive  |  inactive  |  inactive  |  inactive  |  inactive  |
+    +---+------------+------------+------------+------------+------------+
```
taylor-b: I opted for a hand-drawn table, since this lets us get away with a 5x5 table instead of a 3x17 flattened table (one row for each combination).

I considered just writing "choose the directional attribute that's the closest match to the RtpTransceiver's direction within the set of allowed values." But I thought maybe "closest match" would be too vague.

juberti: This looks good to me, but not sure it needs its own section. I think it might be best to simply include this inline where it is referenced.

taylor-b: I couldn't figure out how to put a drawing inside a bulleted list and have the HTML turn out right. Any suggestions? Maybe this is the problem Peter ran into (I just copied his old commit and changed the contents of the table).

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2824,6 +2824,35 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           </section>
         </section>
+        <section title="Direction Attribute in Answers"
+        anchor="sec.answer-dirattr">
+
+          <t><xref target="RFC3264"></xref> direction attributes
+          (defined in Section 6.1) in answers are chosen according to
+          the direction attribute in the remote offer and the direction
+          of the corresponding RtpTransceiver, by choosing the
+          intersection of the two directions in the following table:</t>
```
ekr: I had always assumed the algorithm was as follows:

enum {
  inactive = 0,
  sendonly = 1,
  recevonly = 2,
  sendrecv =3
};

int transceiver;   // How the transceiver is set
int offer;             // How the offer was set

int offer_swapped = (offer & 1) << 1  | (offer & 2) >> 1
answer = offer_swapped & transceiver;

Is this right?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1990,8 +1990,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               the dummy value "9 IN IP4 0.0.0.0", because no candidates
               have yet been gathered.</t>
 
-              <t>A direction attribute which is the same as that of the
-              associated transceiver.</t>
+              <t>A direction attribute, determined by intersecting the
+              direction of the associated RtpTransceiver with the 
+              rules regarding the offer direction specified
+              in <xref target="RFC3264" />, Section 6.1. For example,
```
ekr: This prose feels a bit clunky though I'm not entirely sure what the best prose would be.

Perhaps "determined by applying the rules for offer direction specified in [...] and then intersecting with the direction of the associated RtpTransceiver"

juberti:
Manually merged in commit 490ca75.

Issue #353, opened 2016-10-08
Resolve TODO: Implementation requirements


juberti:
https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-16#section-5.1.1

```
R-16  TODO: any others?
```

juberti:
R-10 (a=ssrc) should probably be removed or adjusted to reflect the fact that JSEP won't generate these lines.

fluffy:
We need to look at the PR started on this and see if this is something we want to do or if we should go the direction of just calling out the non obvious implementation requirements here.

fluffy:
This seems fixed. Closing it

juberti:
Cullen to inspect document and ensure all relevant SDP RFCs are cited in the generating/parsing offer/answer sections (using full list in #360)

juberti:
Is this resolved by #577?

juberti:
Yes. Closing.

PR #360, opened 2016-10-21
start adding missing refs

fluffy:
Fixes https://github.com/rtcweb-wg/jsep/issues/353


fluffy:
Plan is to eventually kill this PR and not merge it but left around for a bit till we figure out what parts of it need to go elsewhere

PR #370, opened 2016-10-28
Fixing a minor issue in the "rollback" section.

taylor-b:
This still refers to RtpSenders being "activated", a concept that isn't
defined anywhere. I replaced this with "track attached to the
transceiver" since that's the terminology that's used elsewhere.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1336,7 +1336,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             be set to null. RtpTransceivers that were created by
             applying a remote offer that was subsequently rolled back
             MUST be removed. However, a RtpTransceiver MUST NOT be
-            removed if the RtpTransceiver's RtpSender was activated by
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1336,7 +1336,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             be set to null. RtpTransceivers that were created by
             applying a remote offer that was subsequently rolled back
             MUST be removed. However, a RtpTransceiver MUST NOT be
-            removed if the RtpTransceiver's RtpSender was activated by
+            removed if a track was attached to the RtpTransceiver via
```
juberti: What happens if addTrack is not used, but someone does RtpSender.setTrack?

taylor-b: `addTrack` is the only thing that has this strange latching behavior. If an application is using the transceiver API directly, we expect it knows what it's doing. I can try to find the minutes if you like, but I'm very confident this is what we decided.

Issue #384, opened 2016-11-11
When "current direction" changes


aboba:
According to WebRTC 1.0 Section 5.4 setDirection() does not take effect immediately: 

"The setDirection method sets the direction of the RTCRtpTransceiver. Calls to setDirection() do not take effect immediately. Instead, future calls to createOffer and createAnswer mark the corresponding media description as sendrecv, sendonly, recvonly or inactive as defined in [JSEP] (section 5.2.2. and section 5.3.2.). Calling setDirection() sets the negotiation-needed flag."

My interpretation of the above is that setDirection immediately sets the transceiver.direction attribute, but that the "current direction" (distinct from the direction attribute) is only changed by calls to setLocal/setRemoteDescription. 

I would like to suggest that JSEP indicate precisely when the "current direction" changes.  My reading of JSEP is that this is determined from the currentLocalDescription and currentRemoteDescription.  For example: 

1. The direction provided in the currentLocalDescription (e.g. the "current direction") is "sendrecv" and setDirection("recvonly") is called.  Since setDirection does not change the direction in currentLocalDescription, the RtpSender does not stop sending.
2. createOffer() is called.  A "recvonly" offer is generated.  
3. setLocalDescription(RecvOnlyOffer) is called.  Now the direction in pendingLocalDescription changes to "recvonly", but the direction in currentLocalDescription is still "sendrecv", so the RtpSender does not stop sending.
4. setRemoteDescription(SendOnlyAnswer) is called.  Now the direction in currentLocalDescription is "recvonly", so the RtpSender will stop sending.

juberti:
Bernard's description makes sense to me; currentLocal/Remote indicate the current direction state.

juberti:
Fixed by #410

PR #410, opened 2016-11-15
Adding RtpTransceiver currentDirection (and direction) attributes.

taylor-b:
This distinguishes between the direction which was last set via
setDirection (which simply affects offer/answer generation) from the
direction that was negotiated, which affects whether a transceiver can
send/recv media.

Addresses issue #384.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1628,9 +1628,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             "recvonly", currentDirection returns "sendonly".
           </t>
           <t>
-            If an answer has never been applied containing an m= section
-            associated with the transceiver, or if the transceiver is
-            stopped, currentDirection returns null.
+            If an answer that references this transceiver has not yet been
+            applied, or if the transceiver is stopped, currentDirection
+            returns null.
           </t>
         </section>
         <section title="setCodecPreferences" anchor="sec.transceiver-set-codec-preferences">
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1609,6 +1609,30 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             section below.
           </t>
         </section>
+        <section title="direction" anchor="sec.transceiver-direction">
+          <t>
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1609,6 +1609,30 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            direction for the transceiver's associated m= section. More
+            specifically, it returns the <xref target="RFC3264"></xref>
+            directional attribute of the associated m= section in the last
+            applied answer, with "send" and "recv" directions reversed if
+            it was a remote answer. For example, if the directional
+            attribute for the associated m= section in a remote answer is
+            "recvonly", currentDirection returns "sendonly".
+          </t>
+          <t>
+            If an answer has never been applied containing an m= section
```
juberti: awkward - suggest 
"If the transceiver has no associated m= section, perhaps because no answer that references this m= section has been applied, or if the transceiver is stopped..."

taylor-b: The reason I worded it this way is that a transceiver is associated with an m= section as soon as an offer is applied (with how JSEP is currently written)... But we only want currentDirection to be populated after applying an *answer*.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1609,6 +1609,30 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             section below.
           </t>
         </section>
+        <section title="direction" anchor="sec.transceiver-direction">
+          <t>
+            The direction method returns the last value passed into
+            setDirection. If setDirection has never been called, it
+            returns the direction the transceiver was initialized with.
+          </t>
```
taylor-b: If you initialize a transceiver with "sendrecv" and the answer is "recvonly", `direction` will be "sendrecv" and `currentDirection` will be "sendonly". Isn't this what we want?

I thought the last point was captured by the fact that `currentDirection` depends on the last applied answer for the transceiver.

draft-ietf-rtcweb-jsep.xml
juberti: OK, I see. In that case, suggest this simplification:

"If an answer that references the transceiver has not yet been applied, or..."

Open to other ideas too.

PR #434, opened 2016-12-09
Remove unneeded TODO. fixes #392

### LINKED ISSUE #392: Section 5.1.1 Remove TODO
Raised by Magnus.

ekr:


juberti:
I think there may still be more work to do here, but we can track in the existing bug (#353) rather than a TODO.

PR #435, opened 2016-12-09
Editorial. Fixes #395

### LINKED ISSUE #395: 5.1.3 editorial requests
Originally raised by Magnus:

Otherwise, assume that AVPF is being used in an AVP
      compatible mode and use AVP timing, i.e., "trr-int=4".

I find this sentence confusing. I think this should be clarified to say:

Otherwise, assume that AVPF is being used in an AVP
      compatible mode, i.e. use AVPF timing configured with a "trr-int=4".

15. Section 5.1.3:

Spurious space in:

For data m= sections, JSEP endpoints MUST support receiving the
      "UDP/ DTLS/SCTP", "TCP/DTLS/SCTP", or "DTLS/SCTP" (for backwards
      compatibility) profiles.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1822,11 +1822,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               rules used for RTCP feedback. If AVPF is provided, or an
               "a=rtcp-fb" attribute is present, assume AVPF timing,
               i.e., a default value of "trr-int=0". Otherwise, assume
-              that AVPF is being used in an AVP compatible mode and use
-              AVP timing, i.e., "trr-int=4".</t>
+              that AVPF is being used in an AVP compatible mode and
+              use AVPF timing configured with "trr-int=4".</t>
```
juberti: In both cases we describe here, we are saying to use AVPF timing, which is confusing. I will ask Magnus in the bug to clarify.

juberti:
Adding link to #395

draft-ietf-rtcweb-jsep.xml
fluffy: I think this is right. AVP does not have configurable timing but AVPF does. And it has two main configurations, the one that is AVP compatible which you get by setting trr-int=4 and the one that is the default for AVPF which is not AVP compatible which you get by setting trr-int=0

fluffy: Might be better if we moved "use AVPF timing"  -> "use AVPF with the timing"

juberti: >    If AVPF is provided, or an
    "a=rtcp-fb" attribute is present, assume AVPF timing,
    i.e., **use** the default value of "trr-int=0". Otherwise, assume
    that AVPF is being used in an AVP compatible mode, and
    **use a value of "trr-int=4000"**.

PR #437, opened 2016-12-09
Require that MIDs and RIDs be generated safely. Fixes #398. Fixes #380

### LINKED ISSUE #380: MID and RID construction
Note that the resolution to this may be here, in Bundle, or in  Security Architecture document; authors are requested to discuss.  The same text would likely be needed in any case.

Magnus Westerlund has proposed the following text for Bundle:

"The identfication-tag when included in the RTP MID SDES item, independent of transport, RTCP SDES packet or RTP header extension, can expose the value to parties beyond the signaling chain. Therefore, the identification-tag MUST NOT contain any user related information for privacy reasons, nor hardware based identifiers. That as such information could enable tracking of the sending end point.

The identification-tag is used to route the media stream to the right application functionality, thus is important that the value received is the one intended by the sender. Malicious modifications can result in that a media stream is wrongly attributed or fails to be played. Thus, verifying integrity and the authenticity of the source are RECOMMENDED to prevent these attacks on the application. Security mechanisms for RTP/RTCP are discussed in Options for Securing RTP Sessions [RFC7201]."

This clarifies that the MID should have no PII and should not enable tracking of the sending end point.  If this advice is followed, the usual requirement to encrypt SDES items in the presence of encrypted RTCP might be waived, if there is consensus that this confidentiality would not serve any purpose commensurate with the difficulty involved.  This is permitted by this text in RFC 7941:

   If the security level is chosen to be different for an SDES item in RTCP and
   an RTP header extension, it is important to justify the exception and
   to consider the security properties as the worst in each aspect for
   the different configurations.  

This text requires, however, that a justification be provided.  It seems best that this justification originate in the RTCWEB working group, no matter where it is eventually documented.

### LINKED ISSUE #398: Section 5.2.1:  Should this reference a specific generation algorithm?
o  An "a=mid" line, as specified in [RFC5888], Section 4.  When
      generating mid values, it is RECOMMENDED that the values be 3
      bytes or less, to allow them to efficiently fit into the RTP
      header extension defined in
      [I-D.ietf-mmusic-sdp-bundle-negotiation], Section 11.

Just to note that to my understanding we are having discussion if this should be referencing a more specific generation algorithm.

Such a change is likely to affect text on RID also:

19. Section 5.2.1:
The list of
      RTCP feedback mechanisms that SHOULD/MUST be supported is
      specified in [I-D.ietf-rtcweb-rtp-usage], Section 5.1.

I would request that we don't use the quite confusing "SHOULD/MUST" and instead write something like:

mechanisms that are recommended or mandated to be supported

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1984,7 +1984,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               be 3 bytes or less, to allow them to efficiently fit into
               the RTP header extension defined in
               <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation">
-              </xref>, Section 11.</t>
+              </xref>, Section 11. MID values MUST be generated in a fashion
```
juberti: In the preceding sentence, we use "mid" rather than "MID".

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1984,7 +1984,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              that does not leak user information, e.g., randomly
```
juberti: repeated "randomly"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1984,7 +1984,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               be 3 bytes or less, to allow them to efficiently fit into
```
juberti: Might want to just write this as a single sentence, e.g.
"When generating mid values, they MUST be generated in a fashion that does not leak user information, e.g. randomly or from a counter, and SHOULD be 3 bytes of less, to allow..."

ekr: Done

draft-ietf-rtcweb-jsep.xml
ekr: Fixed by your change above?

ekr:
@juberti PTAL

ekr:
@juberti ready for you

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2062,11 +2063,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-              implementation. When generating RID values, it is
-              RECOMMENDED that the values be 3 bytes or less, to allow
-              them to efficiently fit into the RTP header extension
-              defined in
-              <xref target="I-D.ietf-avtext-rid"></xref>, Section 11.
+              implementation. RID values MUST be generated in a fashion that does not leak user
+              information, e.g., randomly or using a per-PeerConnection counter, and
+              SHOULD be 3 bytes or less, to allow them to efficiently fit into
+              the RTP header extension defined in
+              <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation">
```
juberti: xref is wrong here

ekr: This just needs to be -14, right?

juberti: I think you need I-D.ietf-avtext-rid here; maybe the xref above needs its section changed as well.

ekr:
@juberti PTAL

Issue #368, opened 2016-10-27
Section 3.7: Simulcast


aboba:
JSEP currently does not provide an API to configure receipt of
    simulcast.  This means that if simulcast is offered by the remote
    endpoint, the answer generated by a JSEP endpoint will not indicate
    support for receipt of simulcast, and as such the remote endpoint
    will only send a single encoding per m= section.  

[BA] In the description of addTransceiver, WebRTC 1.0 Section 5.1 states: 

However when setRemoteDescription is called with a corresponding remote description that is able to send multiple RTP encodings as defined in [JSEP], the RTCRtpReceiver may receive multiple RTP encodings and the parameters retrieved via the transceiver's receiver.getParameters() will reflect the encodings negotiated.

This appears to indicate that a JSEP endpoint may indicate support for receipt of simulcast in an answer. 

    In addition, when
    the JSEP endpoint is the answerer, the permitted encodings for the
    RTPSender must be consistent with the offer, but this information is
    currently not surfaced through any API.  

[BA] Not sure what this is trying to say. Once setRemoteDescription is
called on the JSEP endpoint, won't it be possible to retrieve the sender 
settings via sender.getParameters() and the receiver settings via 
receiver.getParameters()? 

    This means that established
    simulcast streams will continue to work through a received re-offer,
    but setting up initial simulcast by way of a received offer requires
    out-of-band signaling or SDP inspection.  Future versions of this
    specification may add additional APIs to provide this control.

[BA] Does this mean that a JSEP endpoint cannot respond to an offer to receive
simulcast with an Answer indicating that it will send simulcast? Note sure why
this would be the case, particularly if addTransceiver was called prior to 
setRemoteDescription indicating the ability to send simulcast.

juberti:
Seems like there are 2 points here.
1), receiving simulcast streams. This seems like something we don't want to support (unclear which stream we decode, etc)
2), being able to answer a simulcast offer. This seems possible; the reason we discouraged this in the current text was that we didn't have a way for the app to get the desired simulcast sizes to send. How would you see this working with getParameters?

alvestrand:
Room consensus at IETF97 seems to be that trying to get the ability to answer a simulcast offer is not worth it at this stage in document lifecycle; re-offer allows us to get to the same place.

juberti:
@aboba, are you OK with us closing this issue?

juberti:
Related to #417

aboba:
[...]

PR #438, opened 2016-12-09
remove comments about API limitations as may not be true

### LINKED ISSUE #417: Remove claim that API doesn't show number of simultaneous tracks?
Section 3.7 Simulcast - I'm not sure it's true that the information
about number of layers is not surfaced through any API - the number of
elements in the encoding list, and their active/passive state, which is
accessible through getSettings(), should surely give a hint about this.

Suggest removing the claim, since what the API exposes (apart from what's essential for doing JSEP) is beyond the scope of this document.

fluffy:
Fixes #417

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -827,8 +827,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         remote endpoint will only send a single encoding per m=
         section. In addition, when the JSEP endpoint is the answerer,
         the permitted encodings for the RTPSender must be consistent
-        with the offer, but this information is currently not surfaced
-        through any API. This means that established simulcast streams
+        with the offer. This means that established simulcast streams
         will continue to work through a received re-offer, but setting
         up initial simulcast by way of a received offer requires
         out-of-band signaling or SDP inspection. Future versions of


draft-ietf-rtcweb-jsep.xml
```diff
@@ -827,8 +827,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         remote endpoint will only send a single encoding per m=
         section. In addition, when the JSEP endpoint is the answerer,
         the permitted encodings for the RTPSender must be consistent
-        with the offer, but this information is currently not surfaced
-        through any API. This means that established simulcast streams
+        with the offer. This means that established simulcast streams
```
juberti: The surrounding text doesn't really make sense when this clause is removed, e.g. the later points about needing SDP inspection. 

If we are going to say that we can support a simulcast offer, we need to describe how that would work.

juberti:
This may be OBE, as discussed in #368.

juberti:
Not sure we can land this today, we may not need to do anything here, but need to wait to hear from Bernard/Harald.

draft-ietf-rtcweb-jsep.xml
alvestrand: We already have that - the text says that there's no API to munge those settings, which is what's needed to support a simulcast offer. Whether an inspection API exists or not is irrelevant.

alvestrand: Specifically, the section starts "JSEP currently does not provide an API to configure receipt of
   simulcast. " That's true, and sufficient basis for the rest of the text.

If you want to improve the text further, you can change

   This means that established
   simulcast streams will continue to work through a received re-offer,
   but setting up initial simulcast by way of a received offer requires
   out-of-band signaling or SDP inspection.

into

   This means that the UA can ensure that established
   simulcast streams will continue to work through a received re-offer,
   but setting up initial simulcast by way of a received offer won't work.

Reiterating: I'm happy to see it not work.

alvestrand:
High level bit: I'm happy with the proposed PR.

draft-ietf-rtcweb-jsep.xml
juberti: I don't follow this reasoning, unless you interpret "does not provide an API to configure receipt of simulcast" to mean "is unable to receive an offer to send simulcast", which I have been treating as two separate things.

Regarding receipt of a simulcast offer, the difference between an outgoing simulcast offer and an incoming one is that the application knows what it's trying to do in the first case, and can configure the session appropriately. In the incoming case, the application doesn't know that it's received a simulcast offer. IOW, the lack of inspection capability is material, and the explanation of why this can't work doesn't make much sense without it.

fluffy:
This has been replaced by PR #494

Issue #447, opened 2016-12-21
What happens when RTCP mux policy is "require" and an answer is set without "a=rtcp-mux"?


taylor-b:
I may be overlooking something, but I don't think this behavior is documented anywhere.

I think the choices, in order of harshness, are:

- Silently fails, and you don't get RTCP
- negotiationneeded fires, and the m= section is port-zeroed in the next offer
- setRemoteDescription throws an error

I'd lean towards the middle option, but that assumes that applications listen to negotiationneeded.

juberti:
I lean toward the last option, as a violation of an invariant set by the offerer.

juberti:
Agree that this should be documented somewhere.

fluffy:
+1 on the throw an error option

juberti:
The general policy that we discussed here in Seattle was that any spec violation should lead to a setRD exception, not a zero port. For the rtcp-mux-only doc, we probably should have some language that makes rtcp-mux in the answer a MUST.

juberti:
Resolved by #449

PR #449, opened 2017-01-05
Return error if rtcp-mux policy is "require" and "rtcp-mux" is missing.

taylor-b:
This will happen when applying either an offer or an answer, for
consistency.

Also, any "semantics verification" failure will now result in an error,
which was assumed but not explicitly stated.

Fixes issue #447.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1064,8 +1064,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t hangText="require:">The browser will only gather RTP
             candidates. This halves the number of candidates that the
             offerer needs to gather. Applying a description with an
-            m= section missing an "a=rtcp-mux" attribute will cause
-            an error to be returned.</t>
+            m= section that does not contain an "a=rtcp-mux" attribute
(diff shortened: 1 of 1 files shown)


juberti:
This is a good start, but you'll also want to update 4.1.1, 5.3.1 to reflect this (new) policy.

Note though that this will end up being inconsistent with the handling of max-bundle as it is described in 4.1.1, although I still think this is the right approach.

taylor-b:
Updated 4.1.1. I thought I already made the needed change to 5.3.1, what else needs to be changed?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1063,9 +1063,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             non-multiplexing endpoints.</t>
             <t hangText="require:">The browser will only gather RTP
             candidates. This halves the number of candidates that the
-            offerer needs to gather. When acting as answerer, the
-            implementation will reject any m= section that does not
-            contain an "a=rtcp-mux" attribute.</t>
+            offerer needs to gather. Applying a description with an
```
juberti: This feels awkward. Perhaps "missing" -> "that does not contain"

Issue #409, opened 2016-11-14
What happens when browser receives offer with different capabilities than currently negotiated


fluffy:
Browser offer MCU, VP9, and VP10. MCU answers with VP10. Offer / answer completes. 

Later the MCU offers VP9 to browser on m-line that was currently use VP10. What does the browser do? A strict reading of spec might suggest that since MCU only offered VP9, and current is VP10, the m-line should be rejected. 

Do we need to change this ? Is it OK for browser to respond with VP9.

taylor-b:
Previously I've heard people say "the non-JSEP endpoint can do whatever it wants with codecs." For example, in the thread for issue #266.

andyhutton:
It is definitely OK for the browser to accept the VP9 in the updated offer and I would go further and say that it must accept VP9 in such an updated offer if it is capable of doing so.  The spec should make it clear that browsers MUST accept such codec changes in updated offers.

juberti:
If we solve this, #266 should resolve.

PR #453, opened 2017-01-05
Clarifying that re-offers from non-JSEP endpoints can add codecs.

taylor-b:
Fixes issue #409.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2366,9 +2366,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             lines MUST only include codecs present in the most recent
             answer which have not been excluded by the codec preferences
             of the associated transceiver. Note that non-JSEP endpoints
-            are not subject to these restrictions, and MAY offer media
+            are not subject to these restrictions, and might offer media
             formats that were not present in the most recent answer, as
-            specified in <xref target="RFC3264"></xref>, Section 8.</t>
+            specified in <xref target="RFC3264"></xref>, Section 8.
+            Therefore, JSEP endpoints MUST be prepared to receive such
+            offers.</t>
 
             <t>The media formats on the m= line MUST be generated in the
             same order as in the current local description.</t>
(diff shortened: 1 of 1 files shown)


taylor-b:
Note that I made this a MAY and not a MUST, because the only precedent for a MUST in codec selection is with regards to the mandatory codecs.

ekr:
This seems like it still needs some more work, as we discussed earlier.

taylor-b:
I added some more informative text about non-JSEP endpoints. Does this address your concerns Ekr? There are still some questions about when (if ever) a JSEP endpoint is allowed to include previously discarded codecs in an offer (issue #266), but this PR should at least address issue #409.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2365,7 +2365,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>The m= line and corresponding "a=rtpmap" and "a=fmtp"
             lines MUST only include codecs present in the most recent
             answer which have not been excluded by the codec preferences
-            of the associated transceiver.</t>
+            of the associated transceiver. Note that non-JSEP endpoints
+            are not subject to these restrictions, and MAY offer media
```
ekr: I would not make this a capitalized MAY because this is informative not normative.

One way to handle this is to say ", and therefore JSEP endpoints MUST be prepared to receive..."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2572,7 +2575,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>If codec preferences have been set for the associated
             transceiver, media formats MUST be generated in the
             corresponding order, and MUST exclude any codecs not present
-            in the codec preferences or not present in the offer.</t>
+            in the codec preferences or not present in the offer. Note
+            that non-JSEP endpoints are not subject to this restriction,
+            and MAY add media formats in the answer that are not present
+            in the offer, as specified in <xref target="RFC3264"></xref>,
+            Section 6.1.</t>
```
ekr: See above.

PR #455, opened 2017-01-05
add ref to 5761-update

### LINKED ISSUE #372: Include ref to draft-ietf-avtcore-5761-update
Should also have a ref to draft-ietf-avtcore-5761-update

fluffy:
Fixes #372


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2131,7 +2131,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               have yet been gathered.</t>
 
               <t>An "a=rtcp-mux" line, as specified in
-              <xref target="RFC5761"></xref>, Section 5.1.1.</t>
+              <xref target="RFC5761"></xref>, Section 5.1.1
```
juberti: Let's just point to the update, S 3.1, since it obsoletes 5761.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3618,7 +3619,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>If the media section has RTCP mux enabled, discard any
             RTCP component, and begin or continue muxing RTCP over the
             RTP component, as specified in
-            <xref target="RFC5761" />, Section 5.1.3. Otherwise,
+           <xref target="RFC8035"></xref>, Section 3.1. Otherwise,
```
juberti: I think this should still be 5761

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2303,7 +2304,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>If the m= section is not bundled into another
             m= section, an "a=rtcp" attribute line MUST be added with
             of the default RTCP candidate, as indicated in
-            <xref target="RFC5761"></xref>, section 5.1.3.</t>
+            <xref target="RFC5761"></xref>.</t>
```
juberti: I think this should still be 5761

fluffy:
Changed back as Justin suggested.

juberti:
So I reviewed RFC 8035 and RFC 5761 and came to the conclusion that RFC 5761, S 5.1.3 already says everything we need, and RFC 8035 only changes S 5.1.1 of RFC 5761, so it's not useful in this context (5.1.1 is only when ICE is not used).

As such I simply changed this PR to refer to 5761, S 5.1.3, and removed the 8035 reference.

PTAL

PR #457, opened 2017-01-05
Address concern about when RTCP mux has been selected

### LINKED ISSUE #403: Section 5.8: RTCP mux indication may change
*  If RTCP mux is indicated, prepare to demux RTP and RTCP from
         the RTP ICE component, as specified in [RFC5761],
         Section 5.1.1.  If RTCP mux is not indicated, but was indicated
         in a previous description, this MUST result in an error.

I note that there can be difference here between an offer and an answer if the RTCP mux indication may have changed or not between these two description.

juberti:
Fixes #403


ekr:
LGTM

Issue #377, opened 2016-11-02
Section 3.4.1 Request for explanatory text & editorial change


hardie:
Originally raised by Christer

Section 3.4.1 talks about ³RtpTranceivers², ³RtpSenders² and
³RtpReceivers².

1) I think they should be introduced in their own section (not within the
Session Description Protocol section).

2) I think it should be explained what an RtpXXX really is. At a minimum
there should be a reference to the appropriate W3C document. But, as there
is text on how they related to SDP m- lines etc, 1 or 2 sentences
describing what they are would also be useful.

juberti:
Section 3.4.1 is its own section, and explains what the RtpTransceiver is.

Agree that Sections 3.4.2/3.4.3 (on senders/receivers) could be expanded somewhat.

@cdh4u FYI

juberti:
Resolved by #458

PR #458, opened 2017-01-05
Describing RtpSender and RtpReceiver in more detail.

taylor-b:
Attempts to address issue #377.

Also changed "control" to "inspect" in RtpReceiver description, since an RtpReceiver is readonly and doesn't let an application control anything.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -455,7 +455,7 @@ setRemote(ANSWER) |                                   |
         <section title="RtpSenders" anchor="sec.rtpsenders">
 
           <t>RtpSenders allow the application to control how RTP media
-          are sent. An RtpSender is conceptually responsible for the
+          is sent. An RtpSender is conceptually responsible for the
           outgoing RTP stream(s) described by an m= section. This includes
           encoding the attached MediaStreamTrack, sending RTP media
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -455,12 +455,21 @@ setRemote(ANSWER) |                                   |
         <section title="RtpSenders" anchor="sec.rtpsenders">
 
           <t>RtpSenders allow the application to control how RTP media
-          is sent.</t>
+          is sent. An RtpSender is conceptually responsible for the
+          outgoing RTP stream(s) described by an m= section. This includes
+          encoding the attached MediaStreamTrack, sending RTP media
+          packets, sending RTCP sender reports, and processing received
```
juberti: One issue here is that sender reports also include information about received media. That is, a SR is a superset of an RR, and thereby must include information about **received** media. Also, when SRs are sent, RRs are typically not sent.

My suggestion would be to abstract this text somewhat to refer to "generating and processing RTCP about the outgoing RTP stream(s)", and the text below to "generating and processing RTCP about the incoming RTP stream(s)".

taylor-b: Good point. Done.

ekr:
LGTM

PR #459, opened 2017-01-05
Fix up for Issue #418

### LINKED ISSUE #418: 3.8.2 Media flow management - delete suggestion for flipping SDP direction?
- Section 3.8.2 - "the media flow from these sessions can be managed by
specifying SDP direction attributes in the descriptions". This seems out
of place. Surely flipping track.muted is a much simpler way of managing
them if the problem is what to play to the user?

Suggestion: Remove text.

fluffy:
Fixes #418

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -944,8 +944,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <t>As a result of sharing the media streams, the application
           will end up with N parallel PeerConnection sessions, each
           with a local and remote description and their own local and
-          remote addresses. The media flow from these sessions can be
-          managed, or the application can choose to play out the
+          remote addresses.
+          Which media flow(s) from these sessions
+          are played can be selected by the application,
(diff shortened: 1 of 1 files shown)


juberti:
Since the goal is to only have one session active at a time, I think the best way of handling this would be to use transceiver.setDirection (rather than muted, which requires all sessions to be active, and some receive silent media)

juberti:
Right, but I think that matches the desired outcome best.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -945,8 +945,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           will end up with N parallel PeerConnection sessions, each
           with a local and remote description and their own local and
           remote addresses. The media flow from these sessions can be
-          managed by specifying SDP direction attributes in the
-          descriptions, or the application can choose to play out the
+          managed, or the application can choose to play out the
```
ekr: What does managed mean here?

juberti:
Is this now OBE by #472?

PR #461, opened 2017-01-05
aditional info about PR answer

### LINKED ISSUE #374: Section 3.2 Provisional Answer text
Originally raised by Christer

Section 3.2 says:

   "JSEP also allows for an answer to be treated as provisional by the
    application.  Provisional answers provide a way for an answerer to
    communicate initial session parameters back to the offerer, in order
    to allow the session to begin, while allowing a final answer to be
    specified later. This concept of a final answer is important to the
    offer/answer model; when such an answer is received, any extra
    resources allocated by the caller can be released, now that the exact
    session configuration is known."

1) I would like explicit text saying that this is a deviation from RFC
3264, where there is no such thing as a provisional answer. I would like
to have a dedicated ³Deviations to RFC3264² (or something similar) section.

2) I would like explicit text saying that the final answer may be
completely different (in terms of candidates, codecs, accepted m- line
etc) from the provisional answer.

3) Related to 2), the final answer may also require NEW resources to be
allocated.

fluffy:
Fixes #374

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -322,7 +322,9 @@ ipr="trust200902">
         components, TURN candidates, or video decoders. Provisional
         answers, on the other hand, do no such deallocation; as
         a result, multiple dissimilar provisional answers can be
-        received and applied during call setup.</t>
+        received and applied during call setup. Note the final offer
+        may be different than the provisional offer and may have
+        different IP addresses, ports, codec choices, and so on. </t>
 
         <t>In
         <xref target="RFC3264"></xref>, the constraint at the signaling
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -322,7 +322,9 @@ ipr="trust200902">
         components, TURN candidates, or video decoders. Provisional
         answers, on the other hand, do no such deallocation; as
         a result, multiple dissimilar provisional answers can be
-        received and applied during call setup.</t>
+        received and applied during call setup. Note the final offer
+        may be different than the provisional offer and may have
+        different IP addresses, ports, codec choices, and so on. </t>
```
ekr: Shouldn't this say "answer"?

fluffy: oops yes

PR #464, opened 2017-01-05
Editorial changesfrom Colin Perkins. Fixes #365. Fixes #390

### LINKED ISSUE #365: Editorial/Nits for -17
Raised by Colin Perkins:

"- Section 5.1.1, 1st paragraph: The list of mandatory to implement standards is “derived from the requirements outlined in [I-D.ietf-rtcweb-rtp-usage]”. Should the draft say somewhere that “RTP media transport MUST be implemented according to [I-D.ietf-rtcweb-rtp-usage]”? 
- Section 5.1.1: It’s easy to skim-read R-5 this as “MUST”, since all the other requirements are MUST. It might be more visually distinct if written as  “SHALL NOT” rather than “MUST NOT”.
- Section 5.1.1: R-16 should be removed.
- Section 5.2.1, top of page 38: The header extension is in Section 14 of BUNDLE, not Section 11 (also second bullet point on page 39).
- Section 5.8, 3rd bullet on page 60: I suggest changing “If the MID header extension is supported, prepare to demux RTP data intended for this media section“ to “…prepare to demux RTP streams intended…”, to match the recent discussion around terminology in BUNDLE."

### LINKED ISSUE #390: Section 5.1.1 editorial split request
Originally raised by Magnus

Section 5.1.1:

   This list of mandatory-to-implement specifications is derived from
   the requirements outlined in [I-D.ietf-rtcweb-rtp-usage].

   R-5   [RFC4568] MUST NOT be implemented to signal SDES SRTP keying
         information.

So, the list of mandatory-to-implement includes also some to not implement. That is fine, but probably should be moved to its own list, or rename the list so that it can contain both types.

Secondly, this specific requirement I think comes from the security docs, rather than rtp-usage.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1699,7 +1702,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             DTLS certificate fingerprints.</t>
 
             <t>
-            <xref target="RFC4568"></xref> MUST NOT be implemented to
+            <xref target="RFC4568"></xref> SHALL NOT be implemented to
```
juberti: Would it make sense to move this to the end? Seems a bit odd to encounter it in the middle.

Also, we might want to adjust the prose here to say something like

> The SDES SRTP keying mechanism from RFC 4568 SHALL NOT be implemented, as discussed in <xref security-arch/>.

This might also obviate the need for SHALL NOT, which feels clunky to me.

ekr:
Merged by hand

PR #468, opened 2017-01-05
Remove MUST language from discussion of ICE policy

### LINKED ISSUE #379: Section 3.5.3 Browser requirements for selecting candidates
Originally raised by Christer

Section 3.5.3 contains requirements regarding what a browser MUST do -
e.g., allowing the application to select what types of ICE candidates are
used.

1) I am not sure why such requirements are listed here. But, in any case,
is there a mechanism for applications for selecting candidate types etc?
If so, what mechanism?

juberti:
It’s an API requirement, so the MUST doesn’t belong here.
Fixes #379
Also remove discussion of application filtering of local addresses,
which is no longer believed to be useful.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -579,13 +579,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           server reflexive, and relay. However, in certain cases,
           applications may want to have more specific control over the
           gathering process, due to privacy or related concerns. For
-          example, one may want to suppress the use of host candidates,
-          to avoid exposing information about the local network, or go
-          as far as only using relay candidates, to leak as little
-          location information as possible (note that these choices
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -579,13 +579,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          to avoid exposing information about the local network, or go
-          as far as only using relay candidates, to leak as little
-          location information as possible (note that these choices
-          come with corresponding operational costs). To accomplish
-          this, the browser MUST allow the application to restrict
-          which ICE candidates are used in a session. Note that this
+          example, one may want to only use relay candidates, to leak as little
+          location information as possible (keeping in mind that this choice
+          comes with corresponding operational costs). To accomplish this, the
+          browser allows the application to restrict which ICE candidates
```
ekr: Maybe "the browser" -> JSEP?

PR #470, opened 2017-01-06
Editorial. Fixes #423

### LINKED ISSUE #423: Editorial, section 4.1.8 to 4.1.16
- Section 4.1.8 setLocalDescription - "the PeerConnection must be able
to simultaneously support... until a final answer is received". This is
unclear for several reasons - suggest "while the connection is in the
have-local-offer or have-remote-pranswer state". (There's a race
condition here between media and signalling, of course.)

- Section 4.1.10 currentLocalDescription - "a copy of the current
negotiated local description" is ambiguous (see issue raised in
webrtc-1.0 - https://github.com/w3c/webrtc-pc/issues/921). Suggest
making language here less normative - "shows the current negotiated
local description" would probably be weak enough.

- Section 4.1.15 setConfiguration - "and may result in a change to media
state". Not sure this belongs here - it can lead to a cascade of events
that may end up changing media state, but it is not the call that
directly causes the change. Suggest deleting.

- Section 4.1.16 addIceCandidate - suggest deleting the
"end-of-candidates" text here, and saying separately that there is a
means of indicating end of candidates. There seems to be consensus that
mixing the two was an API design mistake, discussions are about how to
get out of it.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1360,10 +1360,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           things, it sets up local resources for receiving and decoding
           media. In order to successfully handle scenarios where the
           application wants to offer to change from one media format to
-          a different, incompatible format, the PeerConnection must be
-          able to simultaneously support use of both the current and
-          pending local descriptions (e.g. support codecs that exist in
-          both descriptions) until a final answer is received, at which
+          a different, incompatible format, while in the have-local-offer
```
juberti: This sentence is complicated. I would suggest splitting this into two sentences, e.g.

> In order to successfully handle scenarios where the application wants to offer to change from one media format to a different, incompatible format, the PeerConnection must be able to simultaneously support use of both the current and pending local descriptions (e.g. support the codecs that exist in either description). This dual processing begins when the PeerConnection enters the have-local-offer state, and continues until setRemoteDescription is called with either a final answer, at which point the PeerConnection can fully adopt the pending local description, or a rollback, which results in a revert to the current local description.

ekr:
@juberti: updated.

Issue #402, opened 2016-11-14
Section 5.8: Clarification of language setting the MID attribute for m= section


hardie:
Originally raised by Magnus:

Section 5.8:

         +  Find the RtpTransceiver that corresponds to the m= section
            with the same MID in the created offer.

         +  Set the value of the RtpTransceiver's mid attribute to the
            MID of the m= section.

I find this text strange. If you already know the MID, why is it being set in the second sub-bullet?

fluffy:
seems like we just need to check this  - prob. editorial

juberti:
I think this is just an artifact of how the createOffer sequence works:
1) addTrack - creates transceiver with no MID
2) createOffer - creates MIDs but doesn't assign them
3) setLocal - assigns MIDs to transceivers
@pthatcherg, is this correct?

Agree that it's a bit confusing at first glance and perhaps we could explain it better.

juberti:
@gloinul for his comments

taylor-b:
Speaking for pthatcherg, yes, that's how we designed it. You can think of the MID in the offer as a "proposed" MID, until setLocalDescription is called.

Part of the reason for this is to simplify some corner cases. For example: createOffer is called and proposes a MID "foo", but before the offer is applied, setRemoteDescription is called, and one of the remote m= sections also happens to use MID "foo". Now when generating a subsequent offer, a different MID would need to be proposed for the new m= section, because you can't have 2 m= sections with the same MID.

juberti:
ping @gloinul

pthatcherg:
Justin's understanding and Taylor explanation are correct.  Before setLocal is called, transceivers aren't "associated" with any m=section, and thus have no MID.

pthatcherg:
I think Taylor's clarification PR is good.

gloinul:
Yes, the PR appears to resolve my issue. Thanks!

juberti:
Resolved by #471

PR #471, opened 2017-01-06
Clarifying when the mid attribute of an RtpTransceiver is set.

taylor-b:
Also describing explicitly what it means for an RtpTransceiver to be
"associated" with an m= section. Namely, that it has a mid attribute
that matches the MID of that m= section.

Addresses issue #402.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -446,7 +446,7 @@ setRemote(ANSWER) |                                   |
           associated with a m= section, or if RtpTransceivers have been
           stopped and disassociated from m= sections. An RtpTransceiver
           is said to be associated with an m= section if its mid property
-          is non-null. Otherwise it is said to be disassociated.
+          is non-null; otherwise it is said to be disassociated.
           The associated m= section is determined using a mapping between
           transceivers and m= section indices, formed when creating an
           offer or applying a remote offer. An RtpTransceiver is never
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2027,8 +2027,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               </xref>, Section 14. Note that the mid attribute of the
               RtpTransceiver is not set when creating an offer; it's only
               set when applying a description, as described in
-              <xref target="sec.applying-a-local-desc"> and
-              <xref target="sec.applying-a-remote-desc">. Thus the
+              <xref target="sec.applying-a-local-desc"></xref> and
+              <xref target="sec.applying-a-remote-desc"></xref>. Thus the
               generated MID value is only considered a "proposed" MID at
               this point.</t>
 
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -449,7 +449,11 @@ setRemote(ANSWER) |                                   |
           session description is applied, a m= section is always
           associated with exactly one RtpTransceiver. An RtpTransceiver
           is said to be associated with an m= section if and only if
-          its mid attribute is equal to the MID of that m= section.</t>
+          its mid attribute is equal to the MID of that m= section, or a
+          mapping exists between the transceiver and the index of that m=
+          section. The index-based mapping is required for remote
+          endpoints that don't support the MID extension; see
+          <xref target="sec.applying-a-remote-desc" />.</t>
 
           <t>RtpTransceivers can be created explicitly by the
           application or implicitly by calling setRemoteDescription
@@ -1339,7 +1343,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             that some RtpTransceivers that were previously associated
             will no longer be associated with any m= section; in such
             cases, the value of the RtpTransceiver's mid property MUST
-            be set to null. RtpTransceivers that were created by
+            be set to null, and the mapping between the transceiver and
+            its m= section index MUST be discarded, if it exists.
+            RtpTransceivers that were created by
             applying a remote offer that was subsequently rolled back
             MUST be removed. However, a RtpTransceiver MUST NOT be
             removed if a track was attached to the RtpTransceiver via
@@ -3380,7 +3386,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <xref target="sec-create-offer"></xref>.</t>
 
               <t>Set the value of the RtpTransceiver's mid property to
-              the MID of the m= section.</t>
+              the MID of the m= section, and establish a mapping between
+              the transceiver and the index of the m= section. The
+              index-based mapping is necessary in case the remote endpoint
+              does not support the MID extension.</t>
             </list></t>
 
             <t>If RTCP mux is indicated, prepare to demux RTP and RTCP
@@ -3497,7 +3506,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>If the m= section is being recycled (see
             <xref target="sec.subsequent-offers"></xref>), dissociate
             the currently associated RtpTransceiver by setting its mid
-            attribute to null.</t>
+            attribute to null and discarding the mapping between the
+            transceiver and its m= section index.</t>
 
             <t>If the m= section is not associated with any
             RtpTransceiver (possibly because it was dissociated in the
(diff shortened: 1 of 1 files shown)


taylor-b:
There probably are better ways of wording this, I'm open to suggestions.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -446,7 +446,9 @@ setRemote(ANSWER) |                                   |
           stopped and disassociated from m= sections. An RtpTransceiver
           is never associated with more than one m= section, and once a
           session description is applied, a m= section is always
-          associated with exactly one RtpTransceiver.</t>
+          associated with exactly one RtpTransceiver. An RtpTransceiver
```
juberti: I would move this sentence before the previous sentence, so that you can explain what association is before you provide additional detail on what it isn't.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -446,7 +446,9 @@ setRemote(ANSWER) |                                   |
...
+          is said to be associated with an m= section if and only if
+          its mid attribute is equal to the MID of that m= section.</t>
```
ekr: Is this in fact true? Note that we only generate MIDs if the other side provided one:

```
   o  If and only if present in the offer, an "a=mid" line, as specified
      in [RFC5888], Section 9.1.  The "mid" value MUST match that
      specified in the offer.
```

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1996,13 +1998,19 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <list style="symbols">
 
               <t>An "a=mid" line, as specified in
-              <xref target="RFC5888"></xref>, Section 4. All mid values
+              <xref target="RFC5888"></xref>, Section 4. All MID values
```
juberti: Is MID capitalized or not? Not sure we've been consistent on this, although I think I prefer MID to mid for consistency with RID.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1996,13 +1998,19 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
               MUST be generated in a fashion that does not leak user
               information, e.g., randomly or using a per-PeerConnection counter, and
               SHOULD be 3 bytes or less, to allow them to efficiently fit into
               the RTP header extension defined in
               <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation">
-              </xref>, Section 14.</t>
+              </xref>, Section 14. Note that the mid attribute of the
```
juberti: Not sure we need to go into this level of detail here in this bullet list. I would just say

> Note that this does not set the RtpTransceiver mid attribute, as that only occurs when the description is applied. The generated MID value can be considered a "proposed" MID at this point.

taylor-b:
Ekr pointed out that this doesn't handle situations where "a=mid" is not used. So I added text describing a transceiver->m= section index mapping.

draft-ietf-rtcweb-jsep.xml
taylor-b: Done.

draft-ietf-rtcweb-jsep.xml
taylor-b: Done.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -444,6 +444,12 @@ setRemote(ANSWER) |                                   |
           sections when RtpTransceivers are created but not yet
           associated with a m= section, or if RtpTransceivers have been
           stopped and disassociated from m= sections. An RtpTransceiver
+          is said to be associated with an m= section if and only if
+          its mid attribute is equal to the MID of that m= section, or a
+          mapping exists between the transceiver and the index of that m=
+          section. The index-based mapping is required for remote
```
ekr: I think this is formally correct, but if and only if <blah> or <blah> is hard.

Maybe "An RtpTransceiver is said to be associated with an m= section if <blah> or <blah>. Otherwise the transceiver is said to be "disassociated"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -444,6 +444,12 @@ setRemote(ANSWER) |                                   |
...
+          endpoints that don't support the MID extension; see
+          <xref target="sec.applying-a-remote-desc" />. An RtpTransceiver
```
ekr: I think you could lose this sentence.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1327,7 +1333,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             that some RtpTransceivers that were previously associated
             will no longer be associated with any m= section; in such
             cases, the value of the RtpTransceiver's mid attribute MUST
-            be set to null. RtpTransceivers that were created by
+            be set to null, and the mapping between the transceiver and
+            its m= section index MUST be discarded, if it exists.
```
ekr: "any mapping between" would let you get read of ", if it exists".

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3339,11 +3350,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             the following steps:
             <list style="symbols">
 
-              <t>Find the RtpTransceiver that corresponds to the m=
-              section with the same MID in the created offer.</t>
+              <t>Find the RtpTransceiver that corresponds to this m=
+              section, by matching the MID of this m= section with the MID
+              that was proposed for the transceiver in the last created
+              offer, as described in
+              <xref target="sec-create-offer"></xref>.</t>
```
ekr: Doesn't this need index text again?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3458,7 +3475,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>If the m= section is being recycled (see
             <xref target="sec.subsequent-offers"></xref>), dissociate
             the currently associated RtpTransceiver by setting its mid
-            attribute to null.</t>
+            attribute to null and discarding the mapping between the
+            transceiver and its m= section index.</t>
```
ekr: Is it possible that there isn't a mapping?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1996,13 +2004,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-              <xref target="RFC5888"></xref>, Section 4. All mid values
+              <xref target="RFC5888"></xref>, Section 4. All MID values
               MUST be generated in a fashion that does not leak user
               information, e.g., randomly or using a per-PeerConnection counter, and
               SHOULD be 3 bytes or less, to allow them to efficiently fit into
               the RTP header extension defined in
               <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation">
-              </xref>, Section 14.</t>
+              </xref>, Section 14. Note that this does not set the
+              RtpTransceiver mid attribute, as that only occurs when
```
juberti: Note that we are referring to attributes as 'properties' in this specification.

taylor-b:
I think I've addressed all of ekr and juberti's comments; ready for another review.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -445,9 +445,14 @@ setRemote(ANSWER) |                                   |
...
           associated with a m= section, or if RtpTransceivers have been
           stopped and disassociated from m= sections. An RtpTransceiver
-          is never associated with more than one m= section, and once a
-          session description is applied, a m= section is always
-          associated with exactly one RtpTransceiver.</t>
+          is said to be associated with an m= section if its mid property
+          is non-null. Otherwise it is said to be disassociated.
+          The associated m= section is determined using a mapping between
+          transceivers and m= section indices, formed when creating an
+          offer or applying a remote offer. An RtpTransceiver is never
```
juberti: do you need to spell out how this mapping is created? It's not obvious to me; I was assuming the mapping would always exist, as an array of transceivers that is populated by calls to addTrack/addTransceiver.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3360,15 +3374,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            section (which will only happen when applying an offer),
             find one and associate it with this m= section according to
             the following steps:
             <list style="symbols">
 
-              <t>Find the RtpTransceiver that corresponds to the m=
-              section with the same MID in the created offer.</t>
+              <t>Find the RtpTransceiver that corresponds to this m=
+              section, using the mapping between transceivers and m=
+              section indices established when creating the offer.</t>
```
juberti: See above regarding when the mapping is created

draft-ietf-rtcweb-jsep.xml
taylor-b: The "creating offer" and "applying remote description" sections have text that mention specifically when a mapping is created.

How the mapping is implemented seemed like an implementation decision, so I tried to not get too specific. I was trying to use language similar to what I saw elsewhere in JSEP, but if you have another suggestion I'm all for it.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -445,9 +445,14 @@ setRemote(ANSWER) |                                   |
           sections when RtpTransceivers are created but not yet
           associated with a m= section, or if RtpTransceivers have been
           stopped and disassociated from m= sections. An RtpTransceiver
-          is never associated with more than one m= section, and once a
-          session description is applied, a m= section is always
-          associated with exactly one RtpTransceiver.</t>
+          is said to be associated with an m= section if its mid property
+          is non-null. Otherwise it is said to be disassociated.
```
juberti: Think a semicolon would be ideal here.

> ; otherwise it is said to be disassociated.

PR #472, opened 2017-01-06
Use setDirection instead of rewriting SDP. Fixes #418

### LINKED ISSUE #418: 3.8.2 Media flow management - delete suggestion for flipping SDP direction?

ekr:


fluffy:
LGTM

Issue #385, opened 2016-11-11
removeTrack()


aboba:
JSEP Section 4.1.2 talks about addTrack(), but there is no equivalent section relating to removeTrack(). 

The current text relating to removeTrack in WebRTC 1.0 Section 5.1 is a bit confusing:

    Stops sending media from sender. The RTCRtpSender will still appear in getSenders. Doing so     
    will cause future calls to createOffer to mark the media description for the corresponding    
    transceiver as recvonlyor inactive, as defined in [JSEP] (section 5.2.2.).

[BA] Since the above text refers to future calls marking the media description, this sounds a lot like a call to setDirection("recvonly") or setDirection("inactive").  However, this would not result in the sender immediately stopping the sending of media.

Since removeTrack is reversible, I would presume it is not equivalent to transceiver.stop() which would be permanent.  While setting the encodings[].active attribute to false would stop sending, that would seem inconsistent with the reference to future calls, which seems to lean more towards setting the direction of the transceiver.

ekr:
I'm not sure I understand the concern here. removeTrack() *both* stops sending and arranges that future createOffers() have the track marked as non-sending

juberti:
I think we have three distinct things:
- sender.setTrack(null): stops sending, no signaling
- removeTrack(sender): stops sending, queues a=recvonly signaling
- transceiver.stop: stops sending, queues port zero signaling

juberti:
@aboba, what would you like to see here?

aboba:
@juberti Taylor had a suggestion to have removeTrack() set sender.track to null (like sender.replaceTrack(null)).  This stops sending immediately, but not permanently, and queues a=recvonly signaling.

juberti:
ok, agreed, as indicated above. What needs to go into JSEP?

juberti:
ping @aboba

PR #474, opened 2017-01-06
Add removeTrack section

aboba:
Fix for Issue https://github.com/rtcweb-wg/jsep/issues/385

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1098,7 +1098,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         <section title="removeTrack" anchor="sec.removeTrack">
 
           <t>The removeTrack method removes a MediaStreamTrack from the
-          PeerConnection, using the the RTCRTpSender argument to indicate
+          PeerConnection, using the RTCRTpSender argument to indicate
           which track is removed. The sender's track attribute is set to
           null, and the sender stops sending.  Future calls to createOffer
           will mark the media description corresponding to the transceiver
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1097,13 +1097,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         </section>
         <section title="removeTrack" anchor="sec.removeTrack">
 
-          <t>The removeTrack method remove a MediaStreamTrack from the
-          PeerConnection, using the the sender argument to indicate which
-          track is removed. The sender's track attribute is set to null,
-          and the sender stops sending.  Future calls to createOffer mark
-          the media description for the corresponding transceiver as recvonly
-          (if transceiver.currentDirection is sendrecv) or as inactive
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1095,6 +1095,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           described in
           <xref target="sec.addTransceiver" />.</t>
         </section>
+        <section title="removeTrack" anchor="sec.removeTrack">
+
+          <t>The removeTrack method remove a MediaStreamTrack from the
```
ekr: removes

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1095,6 +1095,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          PeerConnection, using the the sender argument to indicate which
```
ekr: for addTrack, we used the type (i.e., "MediaStreamTrack") rather than the name of the formal argument.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1095,6 +1095,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          track is removed. The sender's track attribute is set to null,
+          and the sender stops sending.  Future calls to createOffer mark
```
ekr: mark -> will mark I think

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1095,6 +1095,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          the media description for the corresponding transceiver as recvonly
```
ekr: Maybe "the media description corresponding to the transceiver"?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1095,6 +1095,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           described in
           <xref target="sec.addTransceiver" />.</t>
         </section>
+        <section title="removeTrack" anchor="sec.removeTrack">
+
+          <t>The removeTrack method removes a MediaStreamTrack from the
+          PeerConnection, using the RTCRTpSender argument to indicate
+          which track is removed. The sender's track attribute is set to
```
juberti: It's not obvious why you pass in a RtpSender here - perhaps we should say "which sender should have its track removed".

Also, we haven't discussed the track attribute in this document, so this should probably say "The sender's track is cleared, and the sender stops sending."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1095,6 +1095,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          null, and the sender stops sending.  Future calls to createOffer
+          will mark the media description corresponding to the transceiver
```
juberti: since we are talking about the sender, should this say "media description associated with the sender"?

ekr:
I took this over in PR#490

PR #477, opened 2017-01-06
fix maxptime descrpiption

### LINKED ISSUE #399: Section 5.3.1: maxptime clarification
o  If this m= section is for media with configurable frame sizes,
      e.g. audio, an "a=maxptime" line, indicating the smallest of the
      maximum supported frame sizes out of all codecs included above, as
      specified in [RFC4566], Section 6.

In most cases the maxptime attribute doesn't affect the frame size of the encoder, instead it affects the maximum amount of media that are packetized into a RTP payload, i.e. how many frames that are included.

fluffy:
Fixes #399

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2031,10 +2031,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
               <t>If this m= section is for media with configurable
               durations of media per packet, e.g. audio, an
-              "a=maxptime" line, indicating the smallest of the
-              maximum amount of media, expressed as time in
-              milliseconds, that can be encapsulated in each packet
-              out of all codecs included above, as specified in <xref
+              "a=maxptime" line, indicating the smallest, out of all
+              codecs included above, of the maximum amount of media,
+              expressed as time in milliseconds, that can be
+              encapsulated in each packet as specified in <xref
               target="RFC4566"></xref>, Section 6.</t>
 
               <t>If this m= section is for video media, and there are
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2030,10 +2030,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <xref target="RFC3264"></xref>, Section 5.1.</t>
 
               <t>If this m= section is for media with configurable
-              frame sizes, e.g. audio, an "a=maxptime" line, indicating
-              the smallest of the maximum supported frame sizes out of
-              all codecs included above, as specified in
-              <xref target="RFC4566"></xref>, Section 6.</t>
+              durations of media per packet, e.g. audio, an
```
ekr: e.g. -> e.g.,

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2030,10 +2030,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              "a=maxptime" line, indicating the smallest, out of all
+              codecs included above, of the maximum amount of media,
+              expressed as time in milliseconds, that can be
+              encapsulated in each packet as specified in <xref
```
ekr: This seems awkward. I would say

"indicating the maximum amount of media, specified in milliseconds, that can be encapsulated in each packet, as specified in [RFC4566] This value is set to the smallest of the maximum values across all the codecs included in the m= section"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2630,11 +2632,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-            smallest of the maximum supported frame sizes out of all
-            codecs included above, as specified in
-            <xref target="RFC4566"></xref>, Section 6.</t>
+             <t>If this m= section is for media with configurable
+             durations of media per packet, e.g. audio, an
+             "a=maxptime" line, indicating the smallest, out of all
+             codecs included above, of the maximum amount of media,
+             expressed as time in milliseconds, that can be
+             encapsulated in each packet as specified in <xref
+             target="RFC4566"></xref>, Section 6.</t>
```
ekr: Is there a reason we can't just make this a reference to the ection above

PR #478, opened 2017-01-06
discuss what to do if in undefined state

### LINKED ISSUE #429: Apply local description - return to previous state on failure?
Section 5.8 Applying a local description - on errrors, "processing
MUST stop and an error MUST be returned" - should it also say that the
state of the PC needs to be returned to the pre-apply-LD state?

fluffy:
Fixes #429

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3406,6 +3406,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           </list></t>
         </list></t>
 
+         <t> During the processing, if an error is returned, the
+         session is left in an undefined state. The applications needs
+         to resolve this by either termination the session or doing a
+         rollback to a previously defined state as described in <xref
+         target="sec.rollback" />. </t>
+         
         <t>Finally, if this description is of type "pranswer" or
         "answer", follow the processing defined in the
         <xref target="sec.applying-an-answer" /> section below.</t>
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3406,6 +3406,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           </list></t>
         </list></t>
 
+         <t> During the processing, if an error is returned, the
```
juberti: I think I would state this at the outset, i.e., after 

> The following steps are performed at the media engine level to apply a local description.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3406,6 +3406,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+         session is left in an undefined state. The applications needs
```
juberti: applications -> application

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3406,6 +3406,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+         to resolve this by either termination the session or doing a
```
juberti: termination -> terminating

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3406,6 +3406,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+         rollback to a previously defined state as described in <xref
```
juberti: I don't think rollback is guaranteed to recover the session at this point. I think we either need to say we automatically perform rollback, or that the session is left in a bad state.

juberti:
Plan is to write the text in the apply local/remote sections to indicate that upon an error, the session state should be restored to the pre-call state.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3419,7 +3421,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         false.</t>
 
         <t>The following steps are performed at the media engine level
-        to apply a remote description.</t>
+        to apply a remote description.  If an error is returned, the
+        session MUST be restored to the state it was in before
```
juberti: the text here is ambiguous regarding whether canTrickle is reverted - need to resolve this.

PR #479, opened 2017-01-06
Describing when the ICE candidate pool is used and can be resized.

### LINKED ISSUE #381: Need to detail the effects of changing the candidate pool size mid-session.
It's pretty straightforward how `setConfiguration` affects the candidate pool size *before* the pooled candidates are used. By the way, when I say "candidate" below, assume I mean "stream's worth of candidates", since a pool size of 1 really means you gather a bunch of candidates for one ICE "stream".

```
setPoolSize(2);  // 2 candidates gathered
setPoolSize(3);  // 1 additional candidate gathered
setPoolSize(1);  // 2 candidates discarded (if they were gathered by this point)
```

But what about after some (or all) of the pooled candidates begin to be used, as a result of setLocalDescription? Here are some examples:

**Increasing pool size after it's depleted**
```
setPoolSize(2);
setLocalDescription(offerWithTwoStreams);  // Pooled candidates used.
setPoolSize(3);  // Is 1 additional candidate gathered, or are 3?
```

**Decreasing size of depleted pool, then increasing it**
```
setPoolSize(2);
setLocalDescription(offerWithTwoStreams);  // Pooled candidates used.
setPoolSize(1);  // Pool size decreased, but no pooled candidates to discard.
setPoolSize(2);  // Does this do anything? We just decreased the size and increased it again.
```

**Creating pool after SLD**
```
setLocalDescription(offerWithTwoStreams);  // No pooled candidates to use.
setPoolSize(3);  // Is 1 candidate gathered, or are 3?
```

You could think of many other examples.

I can think of two reasonable solutions:

1. When a candidate is taken from the pool, the pool size decreases. This doesn't seem so hard, but it means we need some synchronization between the ICE Agent and User Agent, and the result of `getConfiguration` can now change (which is new).
2. The ICE agent keeps track of both candidates in the pool + taken out of the pool. Additional candidates are only gathered if the size is increased beyond this sum. Going back to the examples:

```
setPoolSize(2);  // Size = 2, Pooled = 2, Taken = 0
setLocalDescription(offerWithTwoStreams);  // Size = 2, Pooled = 0, Taken = 2
setPoolSize(3);  // Size = 3, Pooled = 1, Taken = 2
setPoolSize(1);  // Size = 1, Pooled = 0, Taken = 2
setPoolSize(2);  // Size = 2, Pooled = 0, Taken = 2
```

This is what I implemented in Chrome's WebRTC library (though the feature isn't yet available from JS). It may not be ideal but it's the best thing I could think of that fulfills the requirements of "pool size does not change unless you call `setConfiguration`" and "setting the same size is a no-op".

Or of course, there's also the option:

3. Setting a nonzero pool size after `setLocalDescription` is not allowed. :)

taylor-b:
Fixes #381

The pool is only used in the first offer/answer exchange. After applying
a local description, the pool size is frozen and changing the ICE
servers will *not* result in new candidates being pooled. After an answer
is applied, any leftover candidates are discarded. This only happens
when applying an answer so that candidates aren't prematurely discarded
for use cases that involve pranswer, or multiple offers.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -642,7 +642,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           larger-than-expected number of ICE components is used, or
           because the pool has not had enough time to gather
           candidates, the remaining candidates are gathered as
-          usual.</t>
+          usual. This only occurs for the first offer/answer exchange,
+          after which the candidate pool is emptied and no longer used.
+          </t>
 
           <t>One example of where this concept is useful is an
           application that expects an incoming call at some point in
@@ -1513,9 +1515,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               new ICE credentials, for the purpose of forcing an ICE
               restart and kicking off a new gathering phase, in which
               the new servers will be used. If the ICE candidate pool
-              has a nonzero size, any existing candidates will be
-              discarded, and new candidates will be gathered from the
-              new servers.</t>
+              has a nonzero size, and a local description has not yet been
+              applied, any existing candidates will be discarded, and new
+              candidates will be gathered from the new servers.</t>
 
               <t>Any change to the ICE candidate policy affects the
               next gathering phase. If an ICE gathering phase has
@@ -1526,9 +1528,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               phase occurs, and so any necessary filtering can still be
               done on any pooled candidates.</t>
 
-              <t>Any changes to the ICE candidate pool size take effect
-              immediately; if increased, additional candidates are
-              pre-gathered; if decreased, the now-superfluous
+              <t>The ICE candidate pool size MUST NOT be changed after
+              applying a local description. If a local description has not
+              yet been applied, any changes to the ICE candidate pool size
+              take effect immediately; if increased, additional candidates
+              are pre-gathered; if decreased, the now-superfluous
               candidates are discarded.</t>
 
               <t>The bundle and RTCP-multiplexing policies MUST NOT be
@@ -3732,6 +3736,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         m= sections accordingly, as described in
         <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation" />,
         Section 8.2.</t>
+
+        <t>If the description is of type "answer", and there are still
+        remaining candidates in the ICE candidate pool, discard them.</t>
       </section>
     </section>
     <section title="Processing RTP/RTCP packets" anchor="sec.rtp.demux">


draft-ietf-rtcweb-jsep.xml
```diff
@@ -642,7 +642,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           larger-than-expected number of ICE components is used, or
           because the pool has not had enough time to gather
           candidates, the remaining candidates are gathered as
-          usual.</t>
+          usual. This only occurs for the first offer/answer exchange,
+          after which the candidate pool is emptied and no longer used.
+          </t>
```
ekr: This is actually news to me. What's the reason for this?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1513,7 +1515,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               new ICE credentials, for the purpose of forcing an ICE
               restart and kicking off a new gathering phase, in which
               the new servers will be used. If the ICE candidate pool
-              has a nonzero size, any existing candidates will be
+              has a nonzero size, and a non-provisional local description
+              has not yet been applied, any existing candidates will be
```
ekr: This doesn't seem like quite the right conditional, if I understand the code correctly. Consider

- set pool size = 5
- addTrack()
- createOffer()/setLocal(). We now have 1 set of candidates and a pool size of 4
- setConfiguration() (but we don't discard the pool)l

draft-ietf-rtcweb-jsep.xml
ekr: And it doesn't agree with the text below about setLocal() being the triggering element, not the O/A exchange.

taylor-b: > What's the reason for this?

To simplify the API, since we don't have a use case for using the pool mid-session (or at least, that's the conclusion I think we came to).

> it doesn't agree with the text below about setLocal() being the triggering element
The pool size is set in stone after setLocal(), but Cullen pointed out that if using a local pranswer, you don't want to discard pooled candidates quite yet, until a final answer is applied.

draft-ietf-rtcweb-jsep.xml
taylor-b: So you'd prefer that leftover candidates can still be used? Cullen and my reasoning was: "if you can't resize the pool mid-session, it makes sense to go a step further and say the pool isn't used at all mid-session".

Also, without the ability to lower the pool size, this would mean that if you pooled more candidates than you'll ever need, there's *no* way to discard them.

fluffy:
I can live with this as is. Want to read again if we end up making changes to address EKR comments

draft-ietf-rtcweb-jsep.xml
ekr: Sorry, I don't think I was clear, I am fine with the leftover candidates being unusable. My point is that I think it's unclear when they are discarded. Is it at setLocal() or stable state? Some of the text suggests one and some the other. And if it's at setLocal, why does setConfiguration do anything?

taylor-b:
Updated PR and its description based on our discussion.

PR #481, opened 2017-01-06
Only accept the *last* generated offer/answer in setLocalDescription.

### LINKED ISSUE #366: Need rules to prevent applying an old, obsolete offer/answer
There is now the requirement:

> The SDP returned from createOffer or createAnswer MUST NOT be changed before passing it to setLocalDescription.

However, this requirement by itself doesn't prevent an application from doing the following:

```
fullCapabilityOffer = createOffer();
setLocalDescription(fullCapabilityOffer);
// Send offer, receive answer with reduced capabilities...
setRemoteDescription(reducedCapabilityAnswer);
// Later...
setLocalDescription(fullCapabilityOffer);  // Hey, I didn't modify the SDP, so it's legal!
```

Setting an old offer here is effectively the same as generating a new offer and modifying the SDP (adding codecs etc.), which we want to prevent.

So I think there also need to be rules about when an offer/answer is invalidated. Meaning it will be rejected in the same way that munged SDP would be rejected. Does this occur:
- Whenever the signaling state changes to `stable`? (I _think_ this is sufficient)
- Whenever the signaling state changes at all?
- Whenever the `PeerConnection` is modified by the application at all?

There's also the more innocuous example of:

```
addTransceiver();
oneTransceiverOffer = createOffer();
addTransceiver();
setLocalDescription(oneTransceiverOffer);
```

Where this just means that an implementation needs to be prepared to accept offers that don't match the current state of its transceivers. Which it already must do for answers, so this may not be a big deal.

taylor-b:
Fixes #366

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2864,7 +2864,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         <t>The SDP returned from createOffer or createAnswer MUST NOT
         be changed before passing it to setLocalDescription. If precise
         control over the SDP is needed, the aforementioned
-        createOffer/createAnswer options or RtpSender APIs MUST be
+        createOffer/createAnswer options or RtpTransceiver APIs MUST be
         used.</t>
 
         <t>Note that the application MAY modify the SDP to reduce the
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3327,9 +3327,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         to apply a local description.</t>
 
         <t>First, the parsed parameters are checked to ensure that they
-        have not been altered after their generation in
-        createOffer/createAnswer, as discussed in
-        <xref target="sec.modifying-sdp" />; otherwise, processing MUST
+        are identical to those generated in the last call to
+        createOffer/createAnswer, and have not been altered as discussed
```
juberti: comma after "altered"

PR #486, opened 2017-01-06
Editorial issues from alvestrand. Fixes #422

### LINKED ISSUE #422: Editorials, section 4.1.7
Grouping these as one bug.

Section 4.1.7.1 - the term "inactive" final answer is introduced
without a defintion. Need to either define "inactive" or use another term.

Section 4.1.7.2 Rollback - "and possibly the answerer as well" is
wishy-washy. Suggest "and the answerer if it has done
set-remote-description".

- Section 4.1.7.2 This is the first mention of "pending local / remote
description". Suggest link to definition.

- Section 4.1.7.2 "RTPTransceivers that were created by applying a
remote offer that was subsequently rolled back MUST be removed". The
term "removed" is undefined. Deleted, discarded, destroyed or detached
from the PeerConnection? Suggestion: "stopped and removed from the
PeerConnection". (if the JS app still has a reference to them, it may
still inspect the object).

ekr:
Fixes #422

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1315,7 +1315,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
             <t>A rollback discards any proposed changes to the session,
             returning the state machine to the stable state, and
-            setting the pending local and/or remote description back to
+            setting the pending local and/or remote description
+            (See <xref target="sec.pendinglocaldescription"/>
+            and <xref target="sec.pendingremotedescription"/>) to
             null. Any resources or candidates that were allocated by
             the abandoned local description are discarded; any media
             that is received will be processed according to the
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1315,7 +1315,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
             <t>A rollback discards any proposed changes to the session,
             returning the state machine to the stable state, and
-            setting the pending local and/or remote description back to
+            setting the pending local and/or remote description
+            (See <xref target="sec.pendinglocaldescription"/>
```
juberti: See -> see

juberti:
Is there any other action to be taken for the issues mentioned in #422?

ekr:
@juberti: I concluded no.

PR #411, opened 2016-11-16
Demux in bundle

pthatcherg:

PR #489, opened 2017-01-06
Move Magnus and Collin's demux algorithm to BUNDLE (appendix B for now) 

pthatcherg:


fluffy:
I think that this has a long ways to go before we could use it and https://github.com/rtcweb-wg/jsep/pull/411 we can go with now.

aboba:
@fluffy I agree this has a long ways to go.

ekr:
Overtaken by what is currently in the spec.

PR #493, opened 2017-01-07
Reoffer codecs

pthatcherg:
If we want to allow applications to control when we do or don’t reoffer codecs, here is how we can do it.


taylor-b:
This should update the "subsequent offer" section. Also: should the previously rejected media formats be required to appear at the *end* of the list in the new offer? This may be necessary to avoid codec-flipping.

rshpount:
What we are trying to implement is https://tools.ietf.org/html/rfc3261#section-14.2 :

 A UAS providing an offer in a 2xx (because the INVITE did not contain an offer) SHOULD construct the offer as if the UAS were making a brand new call, subject to the constraints of sending an offer that updates an existing session, as described in [13] in the case of SDP. Specifically, this means that it SHOULD include as many media formats and media types that the UA is willing to support.  The UAS MUST ensure that the session description overlaps with its previous session description in media formats, transports, or other parameters that require support from the peer.  This is to avoid the need for the peer to reject the session description.

I assume this means something like "SHOULD include all codecs that would be included in the initial offer" instead of "MAY include other codecs".

pthatcherg:
Good idea, I changed it to "SHOULD include all codecs ..."

taylor-b:
LGTM, but we may still want to clarify whether the previously discarded codecs should go at the end of the priority list, or at the same position as if it were an initial offer. @rshpount previously mentioned this:

> The order of codecs should ideally be the same as currently negotiated to avoid codec flips. Essentially, what this means all supported codecs not currently negotiated should be added to the end of the offer after the currently negotiated ones.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2513,6 +2518,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             described in
             <xref target="RFC6464"></xref>, Section 4.</t>
           </section>
+          <section title="ReofferCodecs" anchor="sec.reoffer-codecs">
+            <t>If the "ReofferCodecs" option is specified, with a
+            value of "true", the offer SHOULD include all codecs
+            offered by an initial offer, as specified in <xref
+            target="sec.initial-offers"></xref> even if they have been
```
ekr: Perhaps "which it would have included in an initial offer"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2409,13 +2409,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            preferences of the associated transceiver.  However, if
+            the offer was created with the "ReofferCodecs" option
+            (<xref target="sec.reoffer-codecs"></xref>), these lines
+            SHOULD include all codecs that would be offered in an
+            initial offer, as defined in <xref
+            target="sec.initial-offers"></xref>. Note that non-JSEP
+            endpoints are not subject to these restrictions, and might
+            offer media formats that were not present in the most
+            recent answer, as specified in <xref
+            target="RFC3264"></xref>, Section 8.  Therefore, JSEP
```
ekr: This sentence is kind of run-on. Perhaps '; this is permitted by <xref...>'

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2409,13 +2409,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-            formats that were not present in the most recent answer, as
-            specified in <xref target="RFC3264"></xref>, Section 8.
-            Therefore, JSEP endpoints MUST be prepared to receive such
-            offers.</t>
+            answer which have not been excluded by the codec
+            preferences of the associated transceiver.  However, if
+            the offer was created with the "ReofferCodecs" option
+            (<xref target="sec.reoffer-codecs"></xref>), these lines
+            SHOULD include all codecs that would be offered in an
+            initial offer, as defined in <xref
```
ekr: Not a big fan of absolute statements followed by retractions. How about
"Unless the "ReoffereCodecs option was specified, the m=..."

draft-ietf-rtcweb-jsep.xml
ekr: I don't actually think we need this "Note that" because we just told you that JSEP endpoints may generate this.

pthatcherg:
Yes, we do need W3C consensus on the API point.  In fact, don't we need RTCWEB consensus on this as well?

juberti:
Obsoleted by #599

PR #494, opened 2017-01-07
Be more generic about why we can't receive an incoming offer to send simulcast

juberti:
Fixes #438


ekr:
Merged by hand

juberti:
oops, fixes #417

PR #496, opened 2017-01-13
add example text explaining lip sync

### LINKED ISSUE #427: Lipsync algorithm not clear enough for easy verification
Section 5.3.1 I'm unsure about the language on lipsync groups. I can't
figure out without more careful analysis if it works for the case of
1A+1V replying to 1A+1V - in almost any other case, I think the result
is that answer doesn't have LS groups.

fluffy:
Fixes #427


fluffy:
@alvestrand - Would love it if you could have a read of this and see if it was going the right direction. If not, help suggest text - I'm struggling to figure out what is needed here.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2563,25 +2563,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          this indicates a difference of opinion between the offerer
-          and answerer regarding lip sync status, and as such, the
-          offered group MUST be ignored and no corresponding "LS" group
+          <t>The next step is to generate session-level lip sync
+          groups as defined in <xref target="RFC5888" />, Section
+          7. For each group of type "LS" present in the offer,
+          determine which of the local RtpTransceivers identified by
+          that group's mid values reference a common local MediaStream
+          (as specified in the addTrack and addTransceiver
+          methods). If at least two such RtpTransceivers exist, a
```
alvestrand: I think this is not terribly clear - it talks about existing associations between transceivers and LS groups, while I think we should be talking about what we do when deciding to associate a track with a particular transceiver.

This means that the language needs to go into section 5.9 "applying a remote offer", not here - this section should just say "If the ID value of the msid attributes of the m-lines covered by an LS group differ, the answer MUST NOT contain that LS group".

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2563,25 +2563,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          group of type "LS" with the mid values of these
+          RtpTransceivers MUST be added. Otherwise, this indicates a
+          difference of opinion between the offerer and answerer
+          regarding lip sync status, and as such, the offered group
+          MUST be ignored and no corresponding "LS" group
           generated.</t>
 
+          <t> Consider a simple case where both sides have single
```
alvestrand: I think the examples need to go into a separate section, with example SDP. Something like:

Offer:

a=group:LS:aaa bbb (please fix syntax)
m=audio
mid=aaa
a=msid:foo bar
m=video
mid=bbb
a=msid:foo baz

answer, streams match

a=group:LS:aaa bbb
m=audio
a=mid:aaa
a=msid:xyzzy foo
m=video
a=mid:bbb
a=msid:xyzzy bar

answer, streams don't match

<no group line>
m=audio
mid=aaa
msid=xyzzy foo
m=video
mid=bbb
msid=wwwww bar

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2563,25 +2563,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           <t>The next step is to generate m= sections for each m=
           section that is present in the remote offer, as specified in
           <xref target="RFC3264"></xref>, Section 6. For the purposes
-          of this discussion, any session-level attributes in the offer
-          that are also valid as media-level attributes SHALL be
+          of this discussion, any session-level attributes in the
+          offer that are also valid as media-level attributes SHALL be
           considered to be present in each m= section.</t>
 
           <t>The next step is to go through each offered m= section.
```
alvestrand: And then in section 5.9, right after "either find an RTPTransceiver or create one", add:

* If a set of unassociated RTPTransceivers exist that are linked to the same MediaStream, and a set of available m= sections exist that belong to a single LS group and match the types of the streams being sent, choose that set of available m= sections for this set of RTPTransceivers. This will allow the generated answer to contain the LS group.

draft-ietf-rtcweb-jsep.xml
juberti: I think we could have a simple example (with SDP) inline, simply to illustrate the meaning of the text above. 

We can then extend that with a more comprehensive example in the examples section (I think we could use the 7.2 example to show a non-trivial LS situation).

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2563,25 +2563,54 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          <t> Complex examples where MediaStreamTrack A is
+          synchronized with B and at the same time MediaStreamTrack C,
+          D, E are also synchronized with each other but not with A
+          and B are also possible. To make these cases work, the
+          sender need to ensure A and B are in one MediaStream while
+          C,D, and E are in a different MediaStream. This will cause
+          two groups of type "LS" to be created in the Offer and as
+          long as the association of MediaStreamTracks and
+          MediaStreams are not changed between applying the remote
+          Offer and creating the Answer, the groups of type "LS" will
```
taylor-b: It's not that the remote MediaStreams can't be changed, it's that matching local MediaStreams must be added. If I understand the text above correctly.

draft-ietf-rtcweb-jsep.xml
fluffy: Hmm - on this one I think the current text is correct. I wish it was clearer but this is happening when the answer is created.

fluffy:
@juberti  - OK given the new examples, what needs to happen with this text.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2557,31 +2557,125 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          <t>The next step is to generate session-level lip sync
+          groups, as defined in <xref target="RFC5888" />, Section 7.
+          For each group of type "LS" present in the offer, select the
+          local RtpTransceivers that are referenced by the MID values in the
+          specified group, and determine which of them
+          either reference a common local MediaStream
+          (specified in the calls to addTrack/addTransceiver used to create
+          them), or have no MediaStream to reference because they were not
+          created by addTrack/addTransceiver.
+          implicitly by setRemoteDescription. If at least two such
```
ekr: sentence fragment starting with "implicitly"

juberti:
We realized that we messed up the a=recvonly case. I adapted the algorithm to handle it better.
@taylor-b, please check that I got it right.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2557,31 +2557,125 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          will not reference any MediaStreams, causing the preferences of
+          the offerer to be maintained, and so the subsequent answer will
+          contain an identical "LS" group.
+          </t>
+
+          <t>
+            <figure>
+              <artwork>
+                <![CDATA[
+          m=audio 20000 UDP/TLS/RTP/SAVPF 0
```
fluffy: I think we need an LS group here

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2557,31 +2557,125 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          either reference a common local MediaStream
+          (specified in the calls to addTrack/addTransceiver used to create
+          them), or have no MediaStream to reference because they were not
+          created by addTrack/addTransceiver.
+          implicitly by setRemoteDescription. If at least two such
+          RtpTransceivers exist, a
+          group of type "LS" with the mid values of these
+          RtpTransceivers MUST be added. Otherwise, this indicates a
+          difference of opinion between the offerer and answerer
+          regarding lip sync status, and as such, the offered "LS" group
```
ekr: I would just strike everything between the commas here, I.e.,

```
Otherwise, the offered "LS" group..."
```

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2557,31 +2557,125 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          MUST be ignored and no corresponding group generated in the answer.
+          </t>
+
+          <t>As a simple example, consider the following offer of a single
+          audio and single video track contained in the same MediaStream.
+          (Note that SDP lines not relevant to this example have been removed
```
ekr: I would make this a non-parenthetical.
"SDP lines not relevant..."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2557,31 +2557,125 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          Finally, if the answerer does not add any tracks, its transceivers
+          will not reference any MediaStreams, causing the preferences of
+          the offerer to be maintained, and so the subsequent answer will
+          contain an identical "LS" group.
+          </t>
+
+          <t>
+            <figure>
+              <artwork>
+                <![CDATA[
```
fluffy: Don't we need a
> a=group:LS a1 v1
here ?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2557,31 +2557,125 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          For each group of type "LS" present in the offer, select the
+          local RtpTransceivers that are referenced by the MID values in the
+          specified group, and determine which of them
+          either reference a common local MediaStream
+          (specified in the calls to addTrack/addTransceiver used to create
+          them), or have no MediaStream to reference because they were not
+          created by addTrack/addTransceiver. If at least two such
+          RtpTransceivers exist, a
+          group of type "LS" with the mid values of these
+          RtpTransceivers MUST be added. Otherwise, this indicates a
```
taylor-b: Isn't this same thing necessary for creating an offer? In other words, generate an LS group for all offered "m=" sections with no associated streams? This could happen either after:

`createOffer({offerToReceiveAudio:1, offerToReceiveVideo:1})`

Or an offer/answer in the reverse direction of the original:

```
pc1.addTrack(audio);
pc1.addTrack(video);
// Do offer/answer between pc1 and pc2
pc2.doSomethingThatRequiresNegotiation();
pc2.createOffer();
```

fluffy:
We need to add text than when creating a re-offer, if things were previously grouped in LS, we keep them grouped in new LS

fluffy:
This LGTM

fluffy:
Waiting for OK from Taylor then we can merge

fluffy:
@taylor-b - Can you check the fix Justin made looks OK to you. Thanks

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2359,16 +2359,23 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          <t>The "LS" groups are generated in the same way as with
-          initial offers.</t>
+          <t>"a=group:LS" attributes are generated in the same way as for
+          initial offers, with the additional stipulation that any lip sync
+          groups that were present in the most recent answer MUST continue to
+          exist and MUST contain any previously existing MID identifiers, as
+          long as the identified m= sections are not rejected, and the
+          group still contains at least two MID identifiers.
+          This ensures that any synchronized "recvonly" m= sections continue to
+          be synchronized in the new offer.
```
taylor-b: Does this produce the intended result when an "m=" section is recycled in a subsequent offer?

juberti: The MIDs will change upon such recycling, but do you think that needs to be called out, e.g.

> as long as the identified m= sections still exist and are not rejected...

draft-ietf-rtcweb-jsep.xml
juberti: Fixed

juberti:
Given that we are only maintaining OfferToReceiveX for backwards compatibility, I'm inclined not to try to support offering LS groups in initial recvonly offers.

draft-ietf-rtcweb-jsep.xml
taylor-b: That sounds good.

taylor-b:
>Given that we are only maintaining OfferToReceiveX for backwards compatibility, I'm inclined not to try to support offering LS groups in initial recvonly offers.

Ok, this seems good to me then, after you add the text you suggested for recyling.

PR #498, opened 2017-01-14
Make effects of createOffer/createAnswer more clear.

### LINKED ISSUE #375: Section 3.2 Set local description interaction with provisional offer
Originally raised by Christer

Section 3.2 says:

    "To support this, the JSEP media layer
     can provide an offer via the createOffer() method whenever the
     Javascript application needs one for the signaling.  The answerer can
     send back zero or more provisional answers, and finally end the
     offer-answer exchange by sending a final answer.  The state machine
     for this is as follows:"

1) I would like to explicitly note that the generated offer is not applied
(and any previously applied offer is not removed) until
setLocalDescription() is called.

juberti:
Fixes #375

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1204,9 +1204,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           state of the system, so that setLocalDescription will succeed
           when it attempts to acquire those resources.</t>
 
-          <t>Calling this method may do things such as generate new ICE
+          <t>Calling this method may do things such as generating new ICE
           credentials, but does not result in candidate gathering, or
-          cause media to start or stop flowing.</t>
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1204,9 +1204,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           state of the system, so that setLocalDescription will succeed
           when it attempts to acquire those resources.</t>
 
-          <t>Calling this method may do things such as generate new ICE
+          <t>Calling this method may do things such as generating new ICE
           credentials, but does not result in candidate gathering, or
-          cause media to start or stop flowing.</t>
+          cause media to start or stop flowing. Specifically, the offer
```
ekr: I don't think "Specifically" helps here, but it's harmelss.

juberti:
@cdh4u has given his thumbs up as well.

PR #499, opened 2017-01-14
Remove mandatory constraint impact on imageattr

### LINKED ISSUE #414: 3.6.1 Mandatory constraint change on incoming track causing an offer/answer exchange?
Section 3.6.1 describes mandatory constraints applied to an incoming
track resulting in a new offer-answer exchange. Since this is "action at
a distance" (modify the track and the PeerConnection emits a
negotiation-needed), I think this needs to be synced with the webrtc-1.0
API spec, and possibly even media-capture.

Also, what's the state of the track between the changing of constraints
at the receiver and the renegotiation? It seems to me that the track
should be scaled at the receiver to fit the agreed-upon constraints.

juberti:
Fixes #414

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -678,10 +678,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         receiver can use the "a=imageattr" SDP attribute
         <xref target="RFC6236" /> to indicate what video frame sizes it
         is capable of receiving. A receiver may have hard limits on
-        what its video decoder can process, or it may wish to constrain
-        what it receives due to application preferences, e.g. a
-        specific size for the window in which the video will be
-        displayed.</t>
+        what its video decoder can process, or it may have some maximum set
+        by policy.</t>
 
         <t>Note that certain codecs support transmission of samples with aspect
         ratios other than 1.0 (i.e., non-square pixels).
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -694,35 +692,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          MediaStreamTracks until it receives the answer, the offer can
-          only reflect decoder hard limits. If the offerer wishes to
-          set mandatory constraints on video resolution, it must do so
-          after receiving the answer, and the result will be a new
-          offer-answer to communicate them.</t>
-
-          <t>If there are no known decoder limits or mandatory
-          constraints, the "a=imageattr" attribute SHOULD be
-          omitted.</t>
+          <t>The receiver will first intersect any known local limits (e.g.,
```
ekr: I think given that this is an "e.g.", "merge" might be a happier word than "intersect" but I can live with this.

I note that you have now removed the viewport, so that if you attach to a 32x32 window you may still get 1080p. Is that what we agreed.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -694,35 +692,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           <t>Otherwise, an "a=imageattr" attribute is created with
           "recv" direction, and the resulting resolution space formed
-          by intersecting the decoder limits and constraints is used to
+          from the aforementioned intersection is used to
           specify its minimum and maximum x= and y= values. If the
-          intersection is the null set, i.e., there are no resolutions
-          that are permitted by both the decoder and the mandatory
-          constraints, this MUST be represented by x=0 and y=0
+          intersection is the null set, i.e., the degenerate case of no
+          permitted resolutions, this MUST be represented by x=0 and y=0
```
fluffy: It seems that in the degenerate case, one should not offer video to be received at all. ( I don't feel super strongly about this )

draft-ietf-rtcweb-jsep.xml
juberti: I believe this is what we have agreed, based on the input from @alvestrand. Harald, please advise if I have misunderstood.

draft-ietf-rtcweb-jsep.xml
juberti: Do you mean we should reject the m= line or become a=sendonly? I'm not sure either of these are appropriate, since rejecting the m= line prevents sending, and directional attributes can be changed by the app, so simply treating this case like a null track (i.e. just not sending) seems like the simplest solution.

alvestrand: The new text has local policy + hardware limits as the only things to intersect. It would be weird to have those end up with the empty set, but the language seems to me to be be OK for handling this weird case.

PR #492, opened 2017-01-06
Replace almost every use of browser with endpoint

ekr:

PR #501, opened 2017-01-17
Define m= section. Fixes #387

### LINKED ISSUE #387: Terminology
Initially requested by Magnus:

Terminology: m= section. I would propose that the document creates a terminology section and are explicit that m= section and SDP Media Description are the same thing.

2. There are a couple of terms that would really need a definition and also possibly some reflection of how they relate to the RTCWeb terms, like WebRTC Endpoint. So the terms I really like defined are:

JSEP endpoint
JSEP implementation

ekr:


ekr:
I decided it was worth defining m= section, etc. m= line is properly defined already. I decided ICE generation was clear enough. JSEP endpoint etc. are in PR #492

draft-ietf-rtcweb-jsep.xml
```diff
@@ -276,9 +276,11 @@ ipr="trust200902">
 
         <t>In addition, various RFCs put different conditions on the
         format of offers versus answers. For example, an offer may
-        propose an arbitrary number of media streams (i.e. m=
-        sections), but an answer must contain the exact same number as
-        the offer.</t>
+        propose an arbitrary number of media streams (i.e., media
+        descriptions as described in <xref target="RFC4566"/>, Section
+        5.14) but an answer must contain the exact same number as
```
juberti: comma before |but|

draft-ietf-rtcweb-jsep.xml
```diff
@@ -276,9 +276,11 @@ ipr="trust200902">
...
+        the offer. Note: this document uses the terms "m= section",
```
juberti: This feels like an odd place for this note, as it's largely irrelevant to this section. 

I suggest mentioning this in the "Session Description Format" section below.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -276,9 +276,11 @@ ipr="trust200902">
...
+        "media description", and "media section" interchangeably.</t>
```
pthatcherg: Can we just pick one and use it everywhere?  Why do we have to use 3 words for the same thing?  Is it not just a matter of doing a search and replace?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -276,8 +276,9 @@ ipr="trust200902">
 
         <t>In addition, various RFCs put different conditions on the
         format of offers versus answers. For example, an offer may
-        propose an arbitrary number of media streams (i.e. m=
-        sections), but an answer must contain the exact same number as
+        propose an arbitrary number of m= sections (i.e., media
+        descriptions as described in <xref target="RFC4566"/>, Section
+        5.14) but an answer must contain the exact same number as
```
juberti: restore comma

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1030,7 +1031,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation">
           </xref>. Regardless of policy, the application will always
           try to negotiate bundle onto a single transport, and will
-          offer a single bundle group across all media section; use of
+          offer a single bundle group across all m= section; use of
```
juberti: m= section**s**

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3131,9 +3132,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
 
           <t>Once all the session-level lines have been parsed,
-          processing continues with the lines in media sections.</t>
+          processing continues with the lines in m= sections.</t>
         </section>
-        <section title="Media Section Parsing"
+        <section title="M= Section Parsing"
         anchor="sec.media-level-parse">
 
           <t>Like the session-level lines, the media session lines MUST
```
juberti: should this be media section?

draft-ietf-rtcweb-jsep.xml
juberti: (i.e. plural)

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3131,9 +3132,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           </list></t>
 
           <t>Once all the session-level lines have been parsed,
-          processing continues with the lines in media sections.</t>
+          processing continues with the lines in m= sections.</t>
         </section>
-        <section title="Media Section Parsing"
+        <section title="M= Section Parsing"
```
juberti: Not super excited about this change. Since these terms are isomorphic I think it would be fine to use 'm= section' when talking about a specific section and 'media section' when being abstract.

Issue #356, opened 2016-10-12
Update examples to match text


juberti:
Follow new rules for a=rtcp, duplication of attributes
Add dtls-id, rtcp-mux-only
Maybe add imageattr

juberti:
#503 adds dtls-id, and I think follows the new rules for a=rtcp and attrib duplication.
It doesn't yet handle the second example or do rtcp-mux-only.

juberti:
Fixed by #526

Issue #298, opened 2016-07-21
Consider machine-generating examples


juberti:
This would help avoid simple errors. We could polyfill the current impls to generate JSEP-compatible SDP.

juberti:
I started work on a Python script that will do this.

PR #503, opened 2017-01-17
Machine-generate examples

juberti:
Python generator for SDP. Super simplified for the purposes of our examples, but should be reasonably flexible. Right now it only generates the simple example.

Changes to the document:
- session attribs ordering now match text
- a=msid format now matches latest spec (I think)
- use GUIDs for MSIDs consistently
- use short ufrags, as one should
- group transport attribs together
- tweak indenting of fingerprint attrib
- Add support for dtls-id
- Only include a=rtcp in offers (I think this is right)
- Use simpler candidate foundations
- pick a more interesting port for the answer

Relevant to #298 and #356.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3901,14 +3901,12 @@ v=0
 o=- 4962303333179871722 1 IN IP4 0.0.0.0
 s=-
 t=0 0
-a=group:BUNDLE a1 v1
 a=ice-options:trickle
+a=group:BUNDLE a1 v1
 m=audio 56500 UDP/TLS/RTP/SAVPF 96 0 8 97 98
 c=IN IP4 192.0.2.1
 a=mid:a1
-a=rtcp:56501 IN IP4 192.0.2.1
 a=msid:47017fee-b6c1-4162-929c-a25110252400
-       f83006c5-a0ff-4e0a-9ed9-d3e6747be7d9
 a=sendrecv
 a=rtpmap:96 opus/48000/2
 a=rtpmap:0 PCMU/8000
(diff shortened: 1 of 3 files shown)


ekr:
@juberti: are these offers acceptable to chrome?

juberti:
I will verify this.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3901,63 +3901,59 @@ v=0
...
 s=-
 t=0 0
-a=group:BUNDLE a1 v1
 a=ice-options:trickle
+a=group:BUNDLE a1 v1
 m=audio 56500 UDP/TLS/RTP/SAVPF 96 0 8 97 98
 c=IN IP4 192.0.2.1
 a=mid:a1
-a=rtcp:56501 IN IP4 192.0.2.1
 a=msid:47017fee-b6c1-4162-929c-a25110252400
```
fluffy: Why such a large msid ?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3901,63 +3901,59 @@ v=0
...
 a=fingerprint:sha-256
-              19:E2:1C:3B:4B:9F:81:E6:B8:5C:F4:A5:A8:D8:73:04
-             :BB:05:2F:70:9F:04:A9:0E:05:E9:26:33:E8:70:88:A2
+              19:E2:1C:3B:4B:9F:81:E6:B8:5C:F4:A5:A8:D8:73:04:
+              BB:05:2F:70:9F:04:A9:0E:05:E9:26:33:E8:70:88:A2
 a=setup:actpass
+a=dtls-id:1
+a=rtcp:56501 IN IP4 192.0.2.1
 a=rtcp-mux
 a=rtcp-rsize
```
fluffy: I think we require 4584 support so likely need an "a=rtcp-fb: ack" line

draft-ietf-rtcweb-jsep.xml
fluffy: Reading section 3.2.1 of https://tools.ietf.org/html/draft-ietf-mmusic-msid-15, it seems like the msid should have the "id" attribute of the MS set as app data on the msid line.

fluffy:
@ekr , are theses offers acceptable to FF ?

ekr:
@fluffy: working on it

ekr:
Looks like you are duplicating a=fingerprnt when you have a=bundle-only

taylor-b:
"a=setup" is also "TRANSPORT" category and shouldn't be duplicated in answers or bundle-only/already-bundled offers.

> Are these offers acceptable to FF?

It doesn't like the absence of "a=ice-ufrag" or "a=fingerprint". I filed this bug just yesterday: https://bugzilla.mozilla.org/show_bug.cgi?id=1332163

Chrome has the same issues, which I'm working on fixing.

gen_examples.py
```diff
@@ -0,0 +1,244 @@
...
+    a=rtcp-fb:100 nack
+    a=rtcp-fb:100 nack pli
+    """
+
+  DATA_SDP = \
+ """m=application {0[local_port]} UDP/DTLS/SCTP webrtc-datachannel
+    c=IN IP4 {0[local_ip]}
+    a=mid:{0[mid]}
+    a=fmtp:webrtc-datachannel max-message-size=65536
+    a=sctp-port 5000
```
nils-ohlmeier: This is not compliant to https://tools.ietf.org/html/draft-ietf-mmusic-sctp-sdp-21#section-5.2
It should be 'a=sctp-port:5000' instead.

gen_examples.py
```diff
@@ -0,0 +1,244 @@
...
+    a=rtcp-fb:100 ccm fir
+    a=rtcp-fb:100 nack
+    a=rtcp-fb:100 nack pli
+    """
+
+  DATA_SDP = \
+ """m=application {0[local_port]} UDP/DTLS/SCTP webrtc-datachannel
+    c=IN IP4 {0[local_ip]}
+    a=mid:{0[mid]}
+    a=fmtp:webrtc-datachannel max-message-size=65536
```
nils-ohlmeier: According to https://tools.ietf.org/html/draft-ietf-mmusic-sctp-sdp-21#section-6 there is no fmtp attribute any more. It should be 'a=max-message-size:65536' instead.

draft-ietf-rtcweb-jsep.xml
juberti: Yes, I misunderstood this. Will fix.

draft-ietf-rtcweb-jsep.xml
juberti: For this PR I would like to avoid adding any new stuff to the SDP, just fixing the existing stuff. I will add the appropriate new stuff as part of #356.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3901,63 +3901,59 @@ v=0
...
-a=rtcp:56503 IN IP4 192.0.2.1
 a=mid:v1
 a=msid:61317484-2ed4-49d7-9eb7-1414322a7aae
-       f30bdb4a-5db8-49b5-bcdc-e0c9a23172e0
 a=sendrecv
 a=rtpmap:100 VP8/90000
 a=rtpmap:101 rtx/90000
 a=fmtp:101 apt=100
-a=ice-ufrag:BGKkWnG5GmiUpdIV
+a=extmap:3 urn:ietf:params:rtp-hdrext:sdes:mid
```
juberti: I thought there was some reason they needed to be unique. But TBH I am not 100% sure.

What does the attribute grouping doc say?

gen_examples.py
juberti: Thanks, will fix

gen_examples.py
juberti: Thanks, will fix

juberti:
@ekr/@taylor-b: Note that bundle is not assumed in the offer, so each m= section should have its own transport attributes (IIUC).

juberti:
OK. All comments addressed. PTAL.

juberti:
Passes tester.js.

Issue #373, opened 2016-11-02
Section 3.2 DTLS parameter


hardie:
Originally raised by Christer

Section 3.2 says:

   "the DTLS-SRTP parameters [RFC5763] sent to a remote party indicate what
   certificate the local side will use in DTLS setup"

1) I am not sure what "parameters" the text is referring to. If it¹s the
fingerprint, I think the text should reference RFC 4572.

2) I think we should simply have a generic statement that, following the
rules in RFC3264, parameters indicate receiving capabilities - unless
explicitly indicated otherwise.

juberti:
The parameters include a=setup and a=fingerprint. 5763 seems like it covers both.

juberti:
@cdh4u, thoughts?

ekr:
Authors believe this text is fine. @cdh4u, please send us a PR with the specific change you want if you disagree.

juberti:
@cdh4u said in email:

> But, in order to do that, I need to know what is meant by "the DTLS-SRTP parameters [RFC5763] sent to a remote party indicate what
certificate the local side will use in DTLS setup".
 
> Justin said: “The parameters include a=setup and a=fingerprint. 5763 seems like it covers both.”
 
> But, a=setup has nothing to do with the certificate, and in case of multiple possible certificates a=fingerprint won’t indicate which one will be used in DTLS setup.
 
>So, I would suggest to replace:
 
    “However, not all parameters follow this rule; for example, the DTLS-
    SRTP parameters [RFC5763] sent to a remote party indicate what
    certificate the local side will use in DTLS setup, and thereby what
    the remote party should expect to receive;”
 
>…with:
 
    “However, not all parameters follow this rule; for example, the fingerprints
      sent to a remote party are calculated based on the local certificate(s)
      offered;”
 
>If you agree with that understanding, I’m happy to submit a pull request.

juberti:
OK, I now see what you are saying, and I think the text you propose looks reasonable. Thanks for offering to submit a PR!

juberti:
[waiting on PR from @cdh4u]

cdh4u:
It seems like I don't have rights to submit new branches (needed for the PR).

ekr:
cd4u: fork the repo and push the branch to your fork, then submit the PR from there.

cdh4u:
PR created: https://github.com/rtcweb-wg/jsep/pull/504

juberti:
Fixed by #504

PR #504, opened 2017-01-17
Editorial change (issue #373)

cdh4u:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -266,13 +266,11 @@ ipr="trust200902">
...
-        <xref target="RFC5763"></xref> sent to a remote party indicate
-        what certificate the local side will use in DTLS setup, and
-        thereby what the remote party should expect to receive; the
-        remote party will have to accept these parameters, with no
-        option to choose different values.</t>
+        parameters follow this rule; for example, the fingerprints
+        <xref target="I-D.ietf-mmusic-4572-update"></xref>
+        sent to a remote party are calculated based on the local
+        certificate(s) offered; the remote party will have to accept
+        these parameters, with no option to choose different values.</t>
```
ekr: perhaps:
"must either accept these parameters or reject the offer, with no..."

cdh4u: PR updated. I used "reject the parameter" instead of "reject the offer", because one can reject one m- line without rejecting the whole offer. Or, does JSEP require accept-all-or-nothing?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -266,13 +266,12 @@ ipr="trust200902">
...
-        parameters
-        <xref target="RFC5763"></xref> sent to a remote party indicate
-        what certificate the local side will use in DTLS setup, and
-        thereby what the remote party should expect to receive; the
-        remote party will have to accept these parameters, with no
-        option to choose different values.</t>
+        parameters follow this rule; for example, the fingerprints
+        <xref target="I-D.ietf-mmusic-4572-update"></xref>
+        sent to a remote party are calculated based on the local
+        certificate(s) offered; the remote party MUST either accept
```
juberti: I would prefer

> the remote party MUST either accept these parameters or reject them altogether, with no option to choose different values.

cdh4u: Done.

PR #510, opened 2017-01-18
Describe how "a=msid" should be generated when there's no track/stream.

### LINKED ISSUE #507: What should a=msid contain for a sender created with a null track?

taylor-b:
Fixes #507 and #586

Suggesting the same thing as mediacapture-streams, which is to generate
UUIDs using the forms in section 4.4 or 4.5 of RFC4122.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2097,8 +2097,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                 the "a=msid" lines MUST use that track's ID. If no
                 MediaStreamTrack is attached, a valid ID MUST be generated
                 instead. To avoid fingerprinting, it is RECOMMENDED to generate
-                a UUID <xref target="RFC4122"></xref> using a form in section
-                4.4 or 4.5 of RFC 4122.</t>
+                a UUID <xref target="RFC4122"></xref> using the form in section
+                4.4 of RFC 4122.</t>
               </list></t>
 
               <t>If the RtpTransceiver has a sendrecv or sendonly
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2089,9 +2089,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               direction:
               <list style="symbols">
 
-                <t>An "a=msid" line, as specified in
-                <xref target="I-D.ietf-mmusic-msid"></xref>, Section
-                2.</t>
+                <t>For each MediaStream that was associated with the
+                transceiver when it was created via addTrack or addTransceiver,
+                an "a=msid" line, as specified in
+                <xref target="I-D.ietf-mmusic-msid"></xref>, Section 2. If a
+                MediaStreamTrack is attached to the transceiver's RtpSender,
+                the "a=msid" lines MUST use that track's ID. If no
+                MediaStreamTrack is attached, a valid ID MUST be generated
+                instead. To avoid fingerprinting, it is RECOMMENDED to generate
+                a UUID <xref target="RFC4122"></xref> using a form in section
+                4.4 or 4.5 of RFC 4122.</t>
               </list></t>
 
               <t>If the RtpTransceiver has a sendrecv or sendonly
@@ -5321,6 +5328,27 @@ a=rtcp-fb:100 nack pli
 
         <seriesInfo name="Internet-Draft" value="draft-ietf-mmusic-dtls-sdp-14"/>
         <format type="TXT" target="http://www.ietf.org/internet-drafts/draft-ietf-mmusic-dtls-sdp-14.txt"/>
+      </reference>
+      <reference anchor="RFC4122">
+        <front>
+          <title>A Universally Unique IDentifier (UUID) URN Namespace</title>
+          <author fullname="P. Leach" initials="P." surname="Leach">
+            <organization></organization>
+          </author>
+          <author fullname="M. Mealling" initials="M."
+          surname="Mealling">
+            <organization></organization>
+          </author>
+          <author fullname="R. Salz" initials="R."
+          surname="Salz">
+            <organization></organization>
+          </author>
+          <date month="July" year="2005" />
+        </front>
+        <seriesInfo name="RFC" value="4122" />
+        <format octets="53319"
+        target="http://www.rfc-editor.org/rfc/rfc4122.txt"
+        type="TXT" />
       </reference>
 
     </references>
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2312,7 +2312,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
             <t>For RtpTransceivers that are not stopped, the "a=msid"
             line(s) MUST stay the same if
-            they are present in the current description.</t>
+            they are present in the current description, regardless of
+            changes to the transceiver's direction or track. If no
+            "a=msid" line is present in the current description,
+            "a=msid" line(s) MUST be generated according to the same
+            rules as for an initial offer.</t>
 
             <t>Each "m=" and c=" line MUST be filled in with the port,
             protocol, and address of the default candidate for the m=
@@ -2356,9 +2360,6 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             If the m= section is bundled into another m= section, both
             "a=candidate" and "a=end-of-candidates" MUST be
             omitted.</t>
-
-            <t>For RtpTransceivers that are still present, the
-            "a=msid" line(s) MUST stay the same.</t>
 
             <t>For RtpTransceivers that are still present, the "a=rid"
             lines MUST stay the same.</t>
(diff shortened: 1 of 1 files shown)


taylor-b:
About the reference: would it be preferable to reference mediacapture-streams directly, or reference RFC4122 (what I did here)? Also, is there any established ordering scheme for the references?

fluffy:
So two concerns on this ... 

1) if multiple things are created with no track, I think it would be better to add them all to the same default track instead of creating separate tracks for them. 

2) these are pretty big and there is no need for a globally unique track name. 

So I wonder if we could solve this by when the track is null setting the MSID to the string "NULL" or "DEFAULT".  

It seems like if we use the system in 4.5, if an app got a fixed node ID, that would end up in the UUID which might be sent in the clear in RTP SDES headers in some cases. This would allow correlation of different calls as being from the same node by on path attackers which seems to create some other privacy issue. So if we do go with the UUID, I'd rather restrict it to just the section 4.4 UUID and drop the 4.5 version.

taylor-b:
Your suggestion to drop 4.5 UUIDs sounds fine to me.

As for using the same track ID for multiple "m=" sections: That seems likely to cause problems. This ID is what's used to create the remote MediaStreamTracks when `setRemoteDescription` is called, so this means you'd get multiple tracks with the same ID, which may break the assumptions of applications that deal with MediaStreamTracks.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2089,9 +2089,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                <t>An "a=msid" line, as specified in
-                <xref target="I-D.ietf-mmusic-msid"></xref>, Section
-                2.</t>
+                <t>For each MediaStream that was associated with the
+                transceiver when it was created via addTrack or addTransceiver,
+                an "a=msid" line, as specified in
+                <xref target="I-D.ietf-mmusic-msid"></xref>, Section 2. If a
+                MediaStreamTrack is attached to the transceiver's RtpSender,
+                the "a=msid" lines MUST use that track's ID. If no
+                MediaStreamTrack is attached, a valid ID MUST be generated
```
juberti: Is a single UUID generated, or two - one for the MediaStream and one for the MST?

juberti:
MSIDs (not MIDs) don't appear on the wire IIUC. Right?

draft-ietf-rtcweb-jsep.xml
taylor-b: You mean if no MediaStreams were provided in `addTrack` or `addTransceiver`? I think that's a separate issue; Harald sent an email about this earlier today: https://www.ietf.org/mail-archive/web/rtcweb/current/msg16240.html

fluffy:
I agree the trackID need to all be different - disregard previous comment, I was thinking about a default stream not track.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2089,9 +2089,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+                instead. To avoid fingerprinting, it is RECOMMENDED to generate
+                a UUID <xref target="RFC4122"></xref> using the form in section
```
juberti: I think this may have come up previously, but do we need to specify a UUID here? Could this just be any valid locally unique id?

That is, could we just say, generate an ID as the implementation would do for a new track?

taylor-b: Any valid ID is allowed, the UUID is just a recommendation to avoid fingerprinting. I copied this recommendation from mediacapture-streams.

But your suggestion seems good to me (and there's less duplication that way). Whatever you prefer.

juberti: I prefer less duplication.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2089,9 +2089,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                <xref target="I-D.ietf-mmusic-msid"></xref>, Section
-                2.</t>
+                <t>For each MediaStream that was associated with the
+                transceiver when it was created via addTrack or addTransceiver,
+                an "a=msid" line, as specified in
+                <xref target="I-D.ietf-mmusic-msid"></xref>, Section 2. If a
+                MediaStreamTrack is attached to the transceiver's RtpSender,
+                the "a=msid" lines MUST use that track's ID. If no
+                MediaStreamTrack is attached, a valid ID MUST be generated,
+                in the same way that the implementation generates IDs for
```
juberti: Note that this text may not be accurate if replaceTrack has been called. There is text for subsequent offers that says that a=msid needs to remain the same, but it's not clear what should happen if replaceTrack occurs before createOffer.

In any case it would be good idea to spell this out.

Also: I think the interpreting-a-remote-description section only expects there to be a single msid attribute. We'll need to file a bug on that.

taylor-b: As currently written, `addTransceiver`/`replaceTrack`/`createOffer` would use the track's ID. If that's not what we want, what do we want exactly? For the ID to be determined at transceiver creation time? But we also have the special case of `addTrack` latching onto an existing transceiver, in which case we do want the track's ID to be used (I assume). What would you suggest?

juberti: yes, we probably want the track ID to be used in the addTrack case you mention.

Wdyt then - should we just maintain the first track id passed to setLocal?

taylor-b: I don't see any problem with that. The track IDs really only have value to applications using "addTrack", since if you're using the transceiver API you can use MIDs.

Along those lines, I guess another option would be to say "Always generate the ID, unless a track was attached to the transceiver via addTrack".

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2089,9 +2089,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               direction:
```
juberti: Don't we need to generate this ID regardless of transceiver direction?

Specifically, in example 7.2, should Alice have no MSID in her recvonly answer for the video tracks?

taylor-b: No, we'd decided track IDs only go in sendrecv or sendonly descriptions. The "track" event of RTCPeerConnection only fires when applying a media description with a "send" direction, so that's the only place where a track ID is needed.

The reasoning was: if an application is using `addTrack`, and they send an offer and receive a "recvonly" answer, because the answerer hasn't called `addTrack` yet, the "track" event shouldn't fire for the offerer.

mparis:
Hello,
I totally agree with you in creating an msid "If no MediaStreamTrack is attached". Actually, we have already implemented this in our stack.

Moreover, I would like to say that due to the msid can be generated internally, we should expose a way of getting this msid to can attach metainfo at application level like "The msid XYZ is the front-camera, and the msid TUV is the screen-share".
I propose to add a readonly property in RtpSender that will have the value of the **first** MediaStreamTrack's msid or the value autogenerated internally as proposed in:
https://github.com/w3c/webrtc-pc/issues/705
We also have this implemented in our stack.

I hope this can be useful ;)
Best!!

draft-ietf-rtcweb-jsep.xml
juberti: OK, makes sense. 

Does this also affect senders that have been set to "recvonly" by setDirection? (i.e. if you set recvonly direction and call createOffer, is a=msid still present? If not, and you change the track with replaceTrack and call setDirection("sendrecv"), which MSID is used?)

taylor-b: > if you set recvonly direction and call createOffer, is a=msid still present

No; creating a sendrecv transceiver and changing it to recvonly should have the same effect as just creating it as recvonly.

> If not, and you change the track with replaceTrack and call setDirection("sendrecv"), which MSID is used?

Currently, the MSID from the track, which I don't see a problem with. Basically, the first time "a=msid" is generated is when it gets locked in.

juberti: OK, got it. In this case I think the text that says "a=msid must stay the same" needs to change:

      For RtpTransceivers that are not stopped, the "a=msid" line MUST
      stay the same if they are present in the current description.

taylor-b: I thought we were just talking about in initial offers; in subsequent offers, the "a=msid" line should stay the same, since according to mmusic-msid, removing it will cause the track to end:
https://tools.ietf.org/html/draft-ietf-mmusic-msid-16#section-3.2.5

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2089,9 +2089,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
               <list style="symbols">
 
-                <t>An "a=msid" line, as specified in
-                <xref target="I-D.ietf-mmusic-msid"></xref>, Section
-                2.</t>
+                <t>For each MediaStream that was associated with the
+                transceiver when it was created via addTrack or addTransceiver,
+                an "a=msid" line, as specified in
+                <xref target="I-D.ietf-mmusic-msid"></xref>, Section 2. If a
+                MediaStreamTrack is attached to the transceiver's RtpSender,
```
ekr: Sorry to be nitpicking, but I'm not sure that this is quite right.

Suppose you do addTransceiver() and hence there's no MS or MST associated with the transceiver
Now you do replaceTrack() so there's a track. Don't you still need to generate an MSID line? As Justin points out, if no MS is associated, you need a '-' MSID.

taylor-b: Actually, since this refers to mmusic-msid, which itself tells what to do when there's no stream, I think we're ok.

taylor-b: Oh, the problem is "for each MediaStream", I see now

taylor-b:
Ready for another review. I also noticed that '"a=msid" line' was singular in some places where there may be multiple lines, so I fixed that as well.

Issue #367, opened 2016-10-27
Section 6: RTP/RTCP routing rules


aboba:
Some comments.

```
  Construct a table mapping payload type to RtpReceiver for each
  RtpReceiver configured to receive from this transport and for each
  payload type that RtpReceiver is configured to receive.  The
  payload types of a given RtpReceiver are found in the m= section
  corresponding to that RtpReceiver in the local description.  If
  any payload type could map to more than one RtpReceiver, map to
  the RtpReceiver whose m= section appears earliest in the local
  description.
```

[BA] In the case where a payload type can map to  more than one RtpReceiver, what if the earliest m-line is "send-only" or "inactive"?  Wouldn't you want to map to the  RtpReceiver whose m-line is "recv-only" or "send-recv" and is earliest? 

```
  If the packet's payload type is in the payload type table, update
  the the incoming SSRC mapping table to include an entry that maps
  the packet's SSRC to the RtpReceiver for that payload type.  In
  addition, deliver the packet to the associated RtpReceiver and
  stop.
```

[BA] I believe that PT-based SSRC latching breaks two of Cullen's proposed use cases (see: https://mailarchive.ietf.org/arch/msg/rtcweb/cvcpy-Rnj4ntL66haNnKHy2_Yvw ): 

Codec change, case 1: The PTs are all unique, and no SSRCs were signaled in
SDP  There is no RID, MID, FEC or RTX. Sender switches codecs, and the PT
changes (but SSRC stays the same).  The packet goes to the receiver that
matches the new PT.

Codec change, case 2:  The PTs are all unique and SSRCs are signaled in
SDP.  There is no RID, MID, FEC or RTX. Sender switches codecs and the PT
changes (but SSRC stays the same).  SSRC does not match any signaled ones.
The packet goes to the receiver that matches the new PT.

[BA] For the RTCP routing, it is more clear to use the RFC 3550 terms for the various SSRC fields.  For exchange, change:

```
  "If the packet is of type SR, and the sender SSRC for the packet is
  found in the incoming SSRC table, deliver a copy of the packet to
  the RtpReceiver associated with that SSRC.  In addition, for each
  report block in the report whose SSRC is found in the outgoing
  SSRC table, deliver a copy of the RTCP packet to the RtpSender
  associated with that SSRC."
```

To:

```
  "If the packet is of type SR, and the "SSRC of sender" for the packet is
  found in the incoming SSRC table, deliver a copy of the packet to
[...]

PR #511, opened 2017-01-18
PT check needed on SSRC match

aboba:
After an SSRC match, need to check that the packet.PT matches a PT for the "m=" line.

Related to Issue https://github.com/rtcweb-wg/jsep/issues/367

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -5832,8 +5832,10 @@ a=rtcp-fb:100 nack pli
           "m=" line for that MID.</t>
 
           <t>If the packet's SSRC is in the incoming SSRC mapping
-          table, route the packet to the associated "m=" line and
-          stop.</t>
+          table, check that the packet's PT matches a PT included
+          on the associated "m=" line.  If so, route the packet to
+          that associated "m=" line and stop, otherwise drop the
+          packet.</t>
 
           <t>If the packet's payload type is in the payload type
           table, update the the incoming SSRC mapping table to include


fluffy:
I know that some people might argue the drop implicitly happens later when it reaches the normal m-line processing but I prefer the very explicit text proposed here. So I'm in favor of adding this.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -5832,8 +5832,10 @@ a=rtcp-fb:100 nack pli
           "m=" line for that MID.</t>
 
           <t>If the packet's SSRC is in the incoming SSRC mapping
-          table, route the packet to the associated "m=" line and
-          stop.</t>
+          table, check that the packet's PT matches a PT included
+          on the associated "m=" line.  If so, route the packet to
+          that associated "m=" line and stop, otherwise drop the
```
juberti: Change comma to semicolon

pthatcherg: I don't think we want this check here.  I think we we want to route the packet by SSRC and then have the RtpTransceiver drop the packet because the PT doesn't match.  If we do this check here, we end up with a strange edge case where we receive a packet with the SSRC of RtpTransceiver A and PT or RtpTransceiver B, and we end up routing it to B.  I think we should route it to A and then have A drop it.   Anyone sending such a packet is in error and it's better to fail to decode that to route media in a surprising way.

aboba: If the packet has the SSRC of Transceiver A but the PT does not match, the intent was to drop the packet, not route it to Transceiver B (which does match the PT).  Rereading the text, perhaps this is not clear.  Should we instead have text about the RtpTransceiver examining packets routed to it and dropping packets where the PT does not match?

PR #524, opened 2017-01-21
Finish machine generation of examples

juberti:
Fixes #298

PR #526, opened 2017-01-23
Fill in missing attributes in samples

juberti:
a=rtcp-mux-only
LS groups


juberti:
Relevant to #356, although there are a few nits remaining which will be handled in a followup PR

juberti:
FWIW, this is atop #524

juberti:
Added machine-filling of candidate artwork tags as well in this PR.

PR #527, opened 2017-01-23
Update IPs

### LINKED ISSUE #525: Use approved IP addresses in examples
@fluffy sayeth:
> The IP address in IETF stuff have to come out of one of 192.0.2.0/24 198.51.100.0/24
or 203.0.113.0/24. I suggest we go with the convention for all our examples of something along lines of local for offer side is 192.0.2.100 and answer side local is 198.51.100.200 , and the IPs on the outside of there nats are 203.0.113.101 and 203.0.113.201 respectively and relay IP of 203.0.113.102 and 203.0.113.202. The idea being to make it so you can look at last 3 digits and quickly have a clue what address it is.

> Oh, and around the start of the example in the draft add some text that explains our naming conventions to readers.

> And use the last tuple of the IP as the last 3 digits of the port.

juberti:
Fixes #525 

On top of PR #524

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3902,8 +3902,8 @@ t=0 0
 a=ice-options:trickle
 a=group:BUNDLE a1 v1
 
-m=audio 56500 UDP/TLS/RTP/SAVPF 96 0 8 97 98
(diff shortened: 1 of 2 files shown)


juberti:
Rebased

PR #529, opened 2017-01-24
Output candidate objects, not just candidate lines, into examples

juberti:


juberti:
This is atop #526

juberti:
Merge conflicts addressed.

Issue #523, opened 2017-01-21
Appendix B: Routing of RTCP XR packets


aboba:
It has been pointed out to me that the text in JSEP Appendix B does not include text relating to the routing of RTCP XR packets, described in RFC 3611. 

Since RTCP XR packets include report blocks with different formats coming up with appropriate text is challenging. 

For example, the following text covers Loss RLE Report Blocks (Section 4.1), Duplicate RLE Report Blocks (Section 4.2), Packet Receipt Times Report Blocks (Section 4.3), Statistics Summary Report Blocks (Section 4.6) and VoIP Metric Report Blocks (Section 4.7):

   If the packet is of type XR with a BT of 1 (Loss RLE Report Block), 2 (Duplicate RLE Report
   Block),  3 (Packet Receipt Times Report Block), 6 (Statistics Summary Report Block), 
  7 (VoIP Metric Report Block), for each report block in the packet whose "SSRC of source"
   is found in the outgoing SSRC table, deliver a copy of the RTCP packet to the
   "m=" line associated with that SSRC.

However, RFC 3611 also includes the Receiver Reference Report Block (Section 4.4) whose format looks like this: 

    0                   1                   2                   3
    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |     BT=4      |   reserved    |       block length = 2        |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |              NTP timestamp, most significant word             |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |             NTP timestamp, least significant word             |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+

Should this be forwarded to the RtpSender whose outgoing SSRC table that matches the
"SSRC" field in the RTCP XR packet? 

There is also is the DLRR Report Block (Section 4.5) whose format looks like this: 

  0                   1                   2                   3
[...]

PR #530, opened 2017-01-25
Routing of RTCP XR packets

aboba:
Fix for Issue https://github.com/rtcweb-wg/jsep/issues/523


juberti:
This seems fine. From #523, I thought the situation was more complicated.

aboba:
@juberti  In SR/RR the sender SSRC and "SSRC of source" is sufficient for routing. XR is similar since there are reports from senders (where the sender SSRC can be used) and receivers (where the "SSRC of source" can be used).  Perhaps Varun can review.

fluffy:
I have not properly reviewed this but I am fine with it being merged it look OK to others.

juberti:
Is this obsoleted by #538?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -5909,6 +5920,13 @@ a=rtcp-fb:100 nack pli
           <xref target="RFC4585" />, but not previously handled, and the media
           source SSRC for the packet is found in the outgoing SSRC table,
           deliver the packet to the "m=" line associated with that SSRC.</t>
+		
+          <t>If the packet is of type XR, as defined in <xref target="RFC3611" />,
```
fluffy: Given the bundle does not make XR MTI, I think that this should be prefixed with "If the device supports XR, and ...." ANd make the 3611 ref informative not normative.

aboba:
@juberti Probably.

juberti:
Closing for now.

juberti:
Reopening per discussion in #550. I think this is a better approach as well.

juberti:
LGTM. @fluffy, any concerns before I land this?

PR #534, opened 2017-01-27
Add transport warmup example

### LINKED ISSUE #15: Provide example of how to do an inactive stream

juberti:
Fixes #15


juberti:
@fluffy, this is a reworked version of your original text. PTAL and let me know what you think.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1275,38 +1275,41 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            the caller cannot send an updated offer during this time.</t>
+
+            <t>As an example, consider a typical JSEP application that wants to
+            set up audio and video. When the callee
+            receives an offer with these MediaStreamTracks, it could send an
+            answer accepting the audio and video tracks as sendonly (meaning
+            that the caller will not send the callee any media yet, and because
+            the callee has not yet added its own MediaStreamTracks, the callee
+            will not send any media either). It could
+            then ask the user to accept the call, acquire the needed local
```
ekr: I feel like "could" -> "would" would make this graf stronger

draft-ietf-rtcweb-jsep.xml
```diff
@@ -4444,6 +4447,152 @@ a=msid:-
...
       </section>
     </section>
+    <section title="Early Transport Warmup Example" anchor="sec.warmup-example" >
+
+      <t>This example demonstrates the early warmup technique described in
+      <xref target="sec.use-of-provisional-answer"/>. Here, Alice's endpoint
+      sends an offer to Bob's endpoint to start an audio/video call. Bob
+      immediately responds with an answer that accepts the audio/video
+      m= sections, but marks them as sendonly (from its perspective), meaning
+      that Alice will not yet send media. This allows the browser to start
```
ekr: s/browser/JSEP implementation/

draft-ietf-rtcweb-jsep.xml
```diff
@@ -4444,6 +4447,152 @@ a=msid:-
...
+      negotiating ICE and DTLS immediately. Bob's endpoint then prompts him
+      to answer the call, and when he does, his endpoint sends a second offer
```
ekr: s/endpoint/JSEP implementation/?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -4444,6 +4447,152 @@ a=msid:-
...
+
+//                  ICE and DTLS establish while call is ringing
+
+//                  Bob accepts call, starts media, and sends a new offer
+BobJS->BobUA:       transceiver.setTrack with audio and video tracks
+BobJS->BobUA:       transceiver.setDirection(sendrecv) for both transceivers
+BobJS->BobUA:       createOffer
+BobJS->BobUA:       setLocalDescription with offer
+BobUA->AliceUA:     media sent from Bob to Alice
+[TODO: ice restart]
```
ekr: Note this TODO. I'm assuming that's intentional

juberti: yes - trying to avoid a mammoth PR

draft-ietf-rtcweb-jsep.xml
juberti: fixed

draft-ietf-rtcweb-jsep.xml
juberti: I think endpoint is probably most appropriate here - only applications/endpoints can trigger the sending of an offer.

draft-ietf-rtcweb-jsep.xml
juberti: The text is treating the approach as hypothetical, i.e. the app *could* do this if one wanted to.

Alternatively I could make it concrete and say something like 'if the app wants to do warmup, it does the following..."

juberti: I reworded since Cullen had comments in this area as well.

juberti:
Cullen PTAL

PR #536, opened 2017-01-30
Add generator for example generator

juberti:
Avoids the hassle of manually generating random strings


tools/new_example.py
```diff
@@ -0,0 +1,58 @@
...
+    'mstv':  str(uuid.uuid4()),
+    'lp':    10000 + num * 100,
+    'sp':    11000 + num * 100,
+    'rp':    12000 + num * 100,
+  }
+
+def main():
+  letter = 'X'
+  if len(sys.argv) > 1:
+    letter = sys.argv[1]
```
ekr: Maybe an else here.

tools/new_example.py
```diff
@@ -0,0 +1,58 @@
...
+"""
+
+def make_obj(num):
+  return {
+    'num':   num,
+    'id':    random.getrandbits(63),
+    'ip':    num * 100,
+    'fp':    ':'.join('%02x' % ord(b) for b in os.urandom(16)).upper(),
+    'ufrag': base64.b64encode(os.urandom(3)),
+    'pwd':   base64.b64encode(os.urandom(18)),
```
ekr: It seems like it would be good to have these deterministic so when you rerun we don't get changes.

tools/new_example.py
```diff
@@ -0,0 +1,58 @@
...
+    'ddir':  ['passive', 'active'][num - 1],
+    'ms':    str(uuid.uuid4()),
+    'msta':  str(uuid.uuid4()),
+    'mstv':  str(uuid.uuid4()),
```
ekr: Same here.

tools/new_example.py
juberti: Not sure I understand. The goal here is to generate new random content for a new example. (As opposed to generating all these parameters by hand, which is tedious)

martinthomson: @ekr is suggesting that you use a deterministic PRNG so that you can regenerate the same examples.  A seeded PRNG achieves that.  That's more work though.

juberti:
Switched to a seedable RNG. PTAL.

PR #537, opened 2017-01-30
SDP for warmup example

juberti:
Builds on #534

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -4432,7 +4432,7 @@ a=max-message-size:65536
 m=video 12100 UDP/TLS/RTP/SAVPF 100 101
 c=IN IP4 192.0.2.100
 a=mid:v1
-a=sendrecv
+a=recvonly
 a=rtpmap:100 VP8/90000
 a=rtpmap:101 rtx/90000
 a=fmtp:101 apt=100
@@ -4440,12 +4440,12 @@ a=extmap:1 urn:ietf:params:rtp-hdrext:sdes:mid
 a=rtcp-fb:100 ccm fir
 a=rtcp-fb:100 nack
 a=rtcp-fb:100 nack pli
-a=msid:- -
+a=msid:- 08575c88-be1f-42d0-8717-5d7f77b57b20
 
 m=video 12100 UDP/TLS/RTP/SAVPF 100 101
 c=IN IP4 192.0.2.100
 a=mid:v2
-a=sendrecv
+a=recvonly
 a=rtpmap:100 VP8/90000
 a=rtpmap:101 rtx/90000
 a=fmtp:101 apt=100
@@ -4453,7 +4453,7 @@ a=extmap:1 urn:ietf:params:rtp-hdrext:sdes:mid
 a=rtcp-fb:100 ccm fir
 a=rtcp-fb:100 nack
 a=rtcp-fb:100 nack pli
-a=msid:- -
+a=msid:- 8dfffea7-c7d6-4036-86bd-98c59391f8a3
 ]]>
 </artwork>
           </figure>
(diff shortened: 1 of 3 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -4550,9 +4594,52 @@ placeholder
...
+a=mid:v1
+a=sendonly
+a=rtpmap:100 VP8/90000
+a=rtpmap:101 rtx/90000
+a=fmtp:101 apt=100
+a=extmap:1 urn:ietf:params:rtp-hdrext:sdes:mid
+a=rtcp-fb:100 ccm fir
+a=rtcp-fb:100 nack
+a=rtcp-fb:100 nack pli
+a=msid:3a0301bb-d0fa-4bf8-9ad4-86fbdf9f3aba
```
ekr: Should you be negotiating the media parameters? What happens if you have two cameras with different params?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1275,38 +1275,41 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+
+            <t>As an example, consider a typical JSEP application that wants to
+            set up audio and video. When the callee
+            receives an offer with these MediaStreamTracks, it could send an
+            answer accepting the audio and video tracks as sendonly (meaning
+            that the caller will not send the callee any media yet, and because
+            the callee has not yet added its own MediaStreamTracks, the callee
+            will not send any media either). It could
+            then ask the user to accept the call, acquire the needed local
+            tracks, and, upon acceptance, send a new offer to the remote side
```
fluffy: seems you might want one point added in here which is after acceptance, it can add the tracks which will case media to start being sent right away, and then it go on to send 2nd offer.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -4444,6 +4447,337 @@ a=msid:-
...
+
+//                  |answer-C1| (sendonly) arrives at Alice
+AliceJS->AliceUA:   setRemoteDescription with |answer-C1|
+AliceUA->AliceJS:   onaddtrack events for audio and video
+
+//                  ICE and DTLS establish while call is ringing
+
+//                  Bob accepts call, starts media, and sends a new offer
+BobJS->BobUA:       transceiver.setTrack with audio and video tracks
+BobJS->BobUA:       transceiver.setDirection(sendrecv) for both transceivers
```
fluffy: I wonder if we should move line 4519  up to here ...  it seems that t at this point, audio and video will start flowing from BobUA to Alice UA

draft-ietf-rtcweb-jsep.xml
juberti: Hmm, good point.

In this case I think we need to because we want to be able to send as soon as possible when the callee eventually answers. You could always change the parameters in the reoffer below.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -4444,6 +4447,337 @@ a=msid:-
...
+BobJS->BobUA:       createOffer
+BobJS->BobUA:       setLocalDescription with offer
+BobUA->AliceUA:     media sent from Bob to Alice
+[TODO: ice restart]
```
fluffy: do we need this TODO ice restart ?

draft-ietf-rtcweb-jsep.xml
fluffy: these parameters are functions of the transport more than the camera so seems like might be OK

draft-ietf-rtcweb-jsep.xml
juberti: I thought this example might be a good example of demoing ICE transport policy changes, so I left this here as a reminder. I will move this to an issue.

draft-ietf-rtcweb-jsep.xml
juberti: agreed, fixed

draft-ietf-rtcweb-jsep.xml
juberti: Addressed in #534

draft-ietf-rtcweb-jsep.xml
juberti: (in #534)

PR #538, opened 2017-01-30
Proposed updated of RTP/RTCP Muxing (Appendix B)

gloinul:
This updated proposal builds on the old Appendix B. And has mostly the
same behavior but describes things using RTP streams, and attempts to
makes no assumptions of how the forwarding of the information to the
relevant consumer or handler is located in relation to the rest of the
RTP/RTCP processing. It is clear on that PT demuxing can't be used other
than initial binding of a new SSRC. It also clarifies RTCP third party
reporting into a seperate category thus reducing overhead in duplicating
infromation form an algorithmic level. Handling of third party
monitoriing is now much more implementation specific.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -5600,6 +5600,44 @@ a=msid:-
 	<seriesInfo name='RFC' value='7656'/>
 	<seriesInfo name='DOI' value='10.17487/RFC7656'/>
       </reference>
+	  
+
+	  
+	  <reference anchor='RFC3611'>
+
+		<front>
+		<title>RTP Control Protocol Extended Reports (RTCP XR)</title>
+		<author initials='T.' surname='Friedman' fullname='T. Friedman'>
+		<organization /></author>
+		<author initials='R.' surname='Caceres' fullname='R. Caceres'>
+		<organization /></author>
+		<author initials='A.' surname='Clark' fullname='A. Clark'>
+		<organization /></author>
+		<date year='2003' month='November' />
+		<abstract>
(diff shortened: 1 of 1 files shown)


juberti:
Can someone generate a usable diff for this PR? (perhaps by running the htmltidy script against old and new?)

gloinul:
@juberti here is a rfcdiff side by side for the Appendix only: 
http://www.denstore.se/IETF/Diff-jsep-18-part.html

gloinul:
Updated with Bernard changes. Diff is also updated. 

Note on diff, it breaks down when you come down to the algorithmic description of handling RTP and RTCP. However, from that point forwards there is such significant rewrite, even if the intention is not to change the basics, the restructuring makes such a difference that you have to compare it on a functional level.

aboba:
@gloinul A question. How does this proposal handled the situation where a BYE is received for an SSRC then subsequently the SSRC is reused, potentially with a new MID and/or PT?

It seems to me that for this case to work correctly, SSRC entries that are "latched" need to be removed when a BYE is received.

gloinul:
@aboba this doesn't explicitly handle that situation any better than the previous proposal. I think there are some merit to explicitly discuss this situation in the text. To handle the re-order use cases, what needs to happen is that the endpoint should remove the RTP stream to m= line table entry 5*Td after having received the BYE. What might be an exception is the case when the SSRC has been explicitly indicated in SDP signalling, then it probably shouldn't be removed until the signalling information is updated.

fluffy:
sent email with comments to the list

fluffy:
I split this diff (with small changes but keeping algorithms fundamentally the same) into a bunch of PR starting from #539 to #556 to make it easier to deal with each issue one by one and see what changed in a more diff friendly way.

juberti:
The most useful parts of this PR were landed in #544 and related PRs. Closing.

PR #540, opened 2017-02-02
magnus update on PT demux intro

fluffy:


aboba:
@juberti Although it says "PT cannot be used" when it is not unique, it does not specifically say that no PT table entries are made in that case.  To ensure that different implementations interoperate, explicit instructions are better.

ekr:
I basically agree with bernard here. We should describe exactly how you handle PT mux.

ekr:
+1 to Justin and Bernard.

juberti:
@aboba, do you think this overall change would be useful with the PT muxing spelled out? Or is the PT muxing already handled sufficiently elsewhere?

juberti:
Deferring judgement on this one for now

aboba:
@juberti  I think the existing text describes PT multiplexing in a way that is easier to understand and implement.  This text seems less clear (though the intent seems to be the same).

juberti:
Based on feedback from @aboba, not planning to take this change.

juberti:
I'm not totally clear on when PT mux is supposed to be used, given that MID in RTP headers and RTCP SDES is mandatory. That said, this PR doesn't really pertain to that question, and as indicated above, is probably unnecessary.

aboba:
@juberti  PT mux is to ensure that existing applications continue to run, and that they can behave consistently across browsers.  Aside from conferencing applications which use SSRC mux today and are likely to migrate to MIDs, many applications use PT mux because it is simple.

PR #541, opened 2017-02-02
process stats line from magnus

fluffy:


aboba:
@fluffy I agree. Before the RTP packet is routed (and packet.PT is checked against the PTs for the m-line), you can't update statistics.

juberti:
I'm inclined to leave this out.

PR #542, opened 2017-02-02
magnus proposed delete outgoing SSRC table

fluffy:


aboba:
@fluffy Agree this should not be removed.  ssrc_table is needed for: 
a. RTCP routing
b. Signaled SSRCs when there is no MID
c. SSRCs latched from MID and PT matches (these "latched" entries may be removed via BYE  or timeout).

ekr:
+1 to Juberti and Fluffy

juberti:
Looks like this can be closed.

PR #543, opened 2017-02-03
Magnus proposed routing alg for rtp

fluffy:


aboba:
@fluffy I agree it is harder to understand.  I would not trust two developers coding from the proposed text to interoperate, whereas I do believe that the existing JSEP-18 text is specific enough to potentially achieve that result.

juberti:
Based on the comments above, not planning on accepting this. Leaving it open until I finish looking over the other PRs.

juberti:
Closing according to the previous comments.

PR #544, opened 2017-02-03
magnus update of SDES handling

fluffy:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -5844,6 +5857,10 @@ a=msid:-
       <t>
         <list>
 
+           <t>On reception of any compound RTCP packet, if there is an
+           RTCP SDES packet included that SHOULD be processed
+           first.</t>
```
fluffy: I think this point is a good think to add

draft-ietf-rtcweb-jsep.xml
```diff
@@ -5858,13 +5875,18 @@ a=msid:-
...
+          this can results in updates and additions to the incoming SSRC
+          to "m=" line mapping table. Thus each of the SDES MID items
+          are processed and the current table entries are checked if
+          the corresponding MID value matches the current SSRC
+          to "m=" line mapping, else the entry is updated. If there is
+          no RTP stream to "m=" line table mapping entry for the
+          received SDES item's SSRC, such an entry is created. Note,
+          that in the process of updating the table entries, update
+          flap suppression as discussed in Section 4.2.6 of <xref
+          target="RFC7941"/> needs to be considereds.</t>
```
fluffy: adding a reference to the flap issues seems OK but I this this actually needs some more text to discuss how this all works. I'd hate for the SSRC to change in a SFU, then the RTP sets the correct SSRC, then some sort of late RTCP packets messes up the mapping and we loose packets. This may actually point to not wanting to use RTCP for SDES.

fluffy:
Related to https://github.com/rtcweb-wg/jsep/issues/545

draft-ietf-rtcweb-jsep.xml
ekr: This should be rewritten to be imperative the same way that the rest of the text is.

ekr: "needs to be considered" is not clear enough to implement

juberti:
Agree this overall looks good but needs to be pounded into shape.

PR #550, opened 2017-02-03
Magnus changes for XR

fluffy:


fluffy:
I think the text in https://github.com/rtcweb-wg/jsep/pull/530 is a better way to go than this text (thought closely related).

draft-ietf-rtcweb-jsep.xml
```diff
@@ -5856,6 +5856,12 @@ a=msid:-
           deliver a copy of the RTCP packet to the "m=" line
           associated with that SSRC.</t>
 
+          <t> If the packet is fo type XR, as defined in RFC3611, it
```
ekr: fo -> of

Also, this needs a reference and a section #.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -5856,6 +5856,12 @@ a=msid:-
...
+          depends on the report. A large number of these report
```
ekr: What depends on the report?

aboba:
+1 to fluffy/ekr.

PR #551, opened 2017-02-03
magnus changes for BYE text

fluffy:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -5866,11 +5866,26 @@ a=msid:-
...
+          current RTP stream to "m=" line table. For this type of
+          pattern it is common that the RTCP packets and information
+          is repeated, either periodically (e.g. SDES items), or for a
+          duration (e.g. BYE), thus suppression of repetitions can be
+          considered. Examples of these are:<list style="hanging">
+
+		  <t hangText="Goodbye (BYE) RTCP Packet:">This <xref
+		  target="RFC3550">RTP/RTCP protocol</xref> mechanism
+		  indicates that a particular RTP stream is leaving the RTP
+		  session. Thus, a most relevant event to inform the handler
```
fluffy: Irrelevant side note but ... In the context of SDP offer / answer, it's pretty much baffling why this is "a most relevant" event. "irrelevant" I could buy

aboba: @fluffy  Have to agree that the original BYE text is easier to understand, though it is WRONG in an important respect, which is that a BYE (or a timeout) needs to result in removal of a "latched" entry in the ssrc table.  Magnus's replacement text doesn't fix the problem though.

ekr:
+1 to Fluffy, Bernard.

juberti:
Agree with the above - will try to rework and fix the problem indicated by @aboba

gloinul:
So, most of the text is discussing the "design pattern" the RTCP Bye falls under. Then the BYE event's "relevant" text may not be good writing. However, I fail to see why BYE becomes irrelevant in a O/A context. Yes, if one terminates the session, it doesn't matter. But for updates that changes the configuration and removes an RTP stream that might not be explicitly signalled in the SDP, then saying BYE matters.

draft-ietf-rtcweb-jsep.xml
juberti: Plan will be to simply address case of removing from SSRC table when receiving a BYE.

aboba: Removal from the "latching table" also needs to occur in event of a timeout.  Note that signalled SSRCs cannot be removed due to a BYE or timeout.

juberti:
@aboba, @fluffy, PTAL. If we need to handle something other than BYE, e.g. timeouts, let's handle that separately.

juberti:
@aboba, any further comments before I land this?

PR #552, opened 2017-02-03
Fix SDES chunk handeling

fluffy:


aboba:
LGTM as well.

fluffy:
@juberti - can you review this one.

PR #553, opened 2017-02-03
magnus update to SDES

fluffy:


aboba:
@fluffy I have to agree. The original text is more specific, though it doesn't describe how  "latched" SSRC table entries entered from MID/PT matches are handled differently from signaled SSRCs (the "latched" entries can be removed as the result of BYEs and timeouts).

ekr:
This seems to remove important material... Hard to tell if it got re-added elsewhere

juberti:
Agree - I think we should not accept this PR.

gloinul:
So, the update of the MID was part of the text on a high level, because it raised the potential issue that if you perform this algorithm without processing the SDES item first, you are routing the RTCP content to the wrong part. 

But, I agree there are information in the old text that probably should be retained.

juberti:
Agree - will include the part about processing MID first and leave the rest.

juberti:
This is covered in #544 now.

PR #548, opened 2017-02-03
magnus changes for FCI

fluffy:

PR #554, opened 2017-02-03
magnus update on RTCP feedback

fluffy:


aboba:
Would like to remove the word "pattern" if possible, and I'd like it to be more clear that we are matching against the outgoing SSRC table. But otherwise this snippet looks ok.

fluffy:
I'd like to cover the cases Magnus raised but would prefer specific things instead of treating it as a pattern that people follow. I'd rather try and figure out what we are missing from existing text and figure out how to add it.

ekr:
+1 to Fluffy/Bernard. This whole thing needs to be specifically algorithmic, not just examples

juberti:
I like the way this treats 4585 cases all at once, but agree we should retain the imperative style.

gloinul:
I wrote the text as refering to design pattern with known examples. Call it something else, but there is underlying issue here. This text considers the RTCP feedback types that RTCWeb has considered to use. It is not covering all RTCP messages that exist even today, and it can't cover any future ones. Thus, I tried to write it in a way that allows a person considering a non mentioned RTCP feedback packet type to match what they have to one of the design patterns and add it into that category.

gloinul:
Regarding @aboba comment on outgoing SSRC table. I mostly agree. Of the examples listed it is the outgoing table that are relevant. For any FB message that is a notification, i.e. in response to a request or command on another end-point's SSRC,  there exists the possibility that the used SSRC for sending the request/command is not in the outgoing table. For example if the RTP stack uses one SSRC for RTCP reporting only, not at all for sending any RTP stream. In that case, that SSRC would not be in the outgoing table, but would the addressed part in the RTCP Feedback message. Yes, corner case, but a possibility. 

Also in this case one could have thrid party targeted FB messages that would be addressed to an SSRC belonging to a peer end-point. They would then not be in the outgoing, but possibly in the incoming table.

juberti:
Obsoleted by #548.

PR #555, opened 2017-02-03
add a comment about 3rd party RTCP

fluffy:


fluffy:
I'm in favour of merging this.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -5886,6 +5886,12 @@ a=msid:-
 
           <t>If the packet is of type APP, deliver a copy of the packet
           to each "m=" line.</t>
+
+          <t> Some multiparty RTP topologies result in endoints
+          receiving RTCP reports for an SSRC that is not associated
+          with the endpoint that received the RTCP packet. These
+          packets MAY be discarded. </t>
```
ekr: Why MAY and not MUST?

fluffy: I don't know - perhaps some things (not browsers) might use them to keep stats on the whole conference or something.

juberti: For the purpose of the routing though, where should they be delivered? Nowhere, right?

juberti:
This seems worthwhile.

draft-ietf-rtcweb-jsep.xml
juberti: @fluffy, thoughts on what the right text is here?

gloinul: I wrote it that way so it allows the basic stupid implementation to just drop the data. However, a smart implementation that do want to monitor the current performance of the path between two session participants can do this. At least one of the old MBONE tools had a dashbord that presented also thrid party statistics.

juberti: OK, will include a sentence like this at the start of the section indicating that packets that don't meet any of the criteria stated here are simply not handled by this algorithm, and MAY be dropped, but really you can do whatever you want.

aboba: If an RTCP packet within a compound packet matches one or more RtpSender/RtpReceivers based on the routing algorithm, the compound packet will be delivered.  If there are no matches, it will be discarded.  So not sure that the "MAY" makes sense (aren't the processing rules mandatory-to-implement?) or how this addition changes the algorithm.

juberti:
Obsoleted by #618

juberti:
May still want to include a comment that this can occur in certain RTP topologies.

PR #556, opened 2017-02-03
magnus updates on 3rd party RTCP

fluffy:


fluffy:
I'd prefer to not merge this one but instead merge in #555

ekr:
+1 to Fluffy.

juberti:
Agree with both - rejecting this one.

PR #557, opened 2017-02-03
Rewrite the text around a=rtcp-mux-only. Fixes #502

### LINKED ISSUE #502: Describe rtcp-mux-only generation in offers/answers
We discuss how to parse it but not how to generate it. Also, I forget whether we should generate a=rtcp in offers when doing mux-only.

ekr:
I applied the following principles, which unfortunately are somewhat
inconsistent with the current mux-exclusive draft.

- You use a=rtcp-mux to negotiate mux
- a=rtcp-mux-only is used by the offerer to say "I really mean it"

To that end, offers contain a=rtcp-mux-only if the policy is require.
Answers don't contain it because you don't need it to negotiate mux.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1067,7 +1067,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             non-multiplexing endpoints.</t>
             <t hangText="require:">The JSEP implementation will only gather RTP
             candidates and will insert an "a=rtcp-mux-only" indication into
-            any offers or answers it generates.
+            any offers it generates.
             This halves the number of candidates that the
             offerer needs to gather. Applying a description with an
             m= section that does not contain an "a=rtcp-mux" attribute
@@ -2185,8 +2185,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               5.1.3. </t>
 
               <t>If the RTP/RTCP multiplexing policy is "require",
-              then an "a=rtcp-mux-only" line, as specified in
-              <xref target="I-D.ietf-mmusic-mux-exclusive" />.</t>
+              an "a=rtcp-mux-only" line, as specified in
+              <xref target="I-D.ietf-mmusic-mux-exclusive" />, Section 4.</t>
 
               <t>An "a=rtcp-rsize" line, as specified in
               <xref target="RFC5506"></xref>, Section 5.</t>
@@ -2443,8 +2443,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             the most recent answer.</t>
 
             <t>If the RTP/RTCP multiplexing policy is "require",
-            then an "a=rtcp-mux-only" line, as specified in
-            <xref target="I-D.ietf-mmusic-mux-exclusive" />.</t>
+            an "a=rtcp-mux-only" line, as specified in
+            <xref target="I-D.ietf-mmusic-mux-exclusive" />, Section 4.</t>
 
             <t>The "a=rtcp-rsize" line MUST only be added if present in
             the most recent answer.</t>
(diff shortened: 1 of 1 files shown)


juberti:
Those principles are consistent with bundle-only, although I wrote the mux-only examples as if mux-only should always be included whenever a=rtcp-mux is present.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1066,7 +1066,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             allowing for compatibility with either multiplexing or
             non-multiplexing endpoints.</t>
             <t hangText="require:">The JSEP implementation will only gather RTP
-            candidates. This halves the number of candidates that the
+            candidates and will insert an "a=rtcp-mux-only" indication into
+            any offers or answers it generates.
```
juberti: Did you mean to say answers here?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2182,6 +2184,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               target="RFC5761"></xref>, Section
               5.1.3. </t>
 
+              <t>If the RTP/RTCP multiplexing policy is "require",
+              then an "a=rtcp-mux-only" line, as specified in
```
juberti: remove 'then'

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2182,6 +2184,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              <xref target="I-D.ietf-mmusic-mux-exclusive" />.</t>
```
juberti: Add relevant section

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2436,8 +2442,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>The "a=rtcp-mux" line MUST only be added if present in
             the most recent answer.</t>
 
-            <t>The "a=rtcp-mux-only" line MUST only be added if present
-            in the most recent answer.</t>
+            <t>If the RTP/RTCP multiplexing policy is "require",
```
juberti: Is this paragraph needed? Or do you want to say that mux-only should not be included at this point?

draft-ietf-rtcweb-jsep.xml
fluffy: I'd say all of section 4 was prob the best ref

draft-ietf-rtcweb-jsep.xml
ekr: Try it now.

ekr:
@juberti: back at you

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3353,7 +3353,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             "ice-lite" attribute MUST NOT be specified.</t>
 
             <t>If the RTP/RTCP multiplexing policy is "require", each m=
-            section MUST contain an "a=rtcp-mux" attribute.</t>
+            section MUST contain an "a=rtcp-mux" attribute. If an
+            "m=" section contains the "a=rtcp-mux-only" attribute
```
juberti: This seems redundant. Did you mean to say rtcp-mux-only in the first sentence?

draft-ietf-rtcweb-jsep.xml
juberti: Is this right though? Do subsequent offers for existing m= sections contain mux-only?

ekr: The document actually says you use if if you *negotiated* mux-exclusive.

https://tools.ietf.org/html/draft-ietf-mmusic-mux-exclusive-10#section-4.5

draft-ietf-rtcweb-jsep.xml
ekr: Assuming I am reading this document correctly, our primary concern here is remote SDP.

So:
1. If we have set require then there better be a=rtcp-mux (or we will reject)
2. If the other side said a=rtcp-mux-only but not a=rtcp-mux then this is bogus SDP and we should reject.

Maybe #1 doesn't go here....

Local SDP doesn't need semantics verification because we do comparison and it's right by construction.

juberti: Right, I misunderstood the context. 

Minor grammar suggestion:
> ... "m=" section contains **an** "a=rtcp-mux-only" attribute, [snip] that section MUST also contain an "a=rtcp-mux" attribute.

draft-ietf-rtcweb-jsep.xml
juberti: Actually, according to your comment about -10, you can remove this para, since it basically just restates the earlier -only advice.

ekr: Removed

ekr:
@juberti: ping

PR #558, opened 2017-02-03
Allow codec preferences to affect subsequent offers/answers.

### LINKED ISSUE #424: Codec preferences - why is order of codecs prohibited from changing?
Section 4.2.4 setCodecPreferences - "However, the codec preferences
cannot change the order of the media formats after an answer containing
the transceiver has been applied". I don't understand why we can't
change the order in order to have the new order reflected on the next
offer/answer exchange - there would be no immediate effect, but the
current language locks in the order permanently, without a clear
justification for why it has to be that way.

The codec order shows the preference order that the receiver would prefer from the sender. The RFC says that the sender SHOULD obey that order, not that it MUST. I do not understand why reordering the codecs is a MUST NOT while deleting them (which is strictly more disruptive) is permitted.

taylor-b:
Fixes #424.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1690,10 +1690,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             The codec preferences of an RtpTransceiver can also determine
             the order of codecs in subsequent calls to createOffer and
             createAnswer, in which case the order of the media formats in
-            the associated m= section will match. However, the codec
-            preferences cannot change the order of the media formats after
-            an answer containing the transceiver has been applied. At this
-            point, codecs can only be removed, not reordered.
+            the associated m= section will match.
           </t>
         </section>
       </section>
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1690,10 +1690,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             The codec preferences of an RtpTransceiver can also determine
             the order of codecs in subsequent calls to createOffer and
             createAnswer, in which case the order of the media formats in
-            the associated m= section will match. However, the codec
-            preferences cannot change the order of the media formats after
-            an answer containing the transceiver has been applied. At this
-            point, codecs can only be removed, not reordered.
+            the associated m= section will match.
```
ekr: I think you should say "will match the codec preferences for the transceiver"

juberti: or, "will follow the specified preferences"

juberti:
Merging.

PR #579, opened 2017-02-04
Fix inconsistencies in example descriptions

### LINKED ISSUE #535: Remove mention of addStream/onaddstream/onaddtrack in examples
Replace with addTrack/ontrack

juberti:
Fixes #535
- onaddstream -> ontrack
- Complex example has 3 candidates for each side
- Warmup example uses trickle ICE


ekr:
@juberti: reminder that you need to write some ICE transports prose here.

Issue #509, opened 2017-01-18
Do subsequent offers contain a=bundle-only for existing m= lines?


juberti:

taylor-b:
If my interpretation is correct, the answer is no:

> Once the offerer BUNDLE address has been selected, the offerer does not need to include the 'bundle-only' attribute in subsequent offers.  By associating the offerer BUNDLE address with an "m=" line of a subsequent offer, the offerer will ensure that the answerer will either keep the "m=" line within the BUNDLE group, or the answerer will have to reject the "m=" line.

In other words: In a subsequent offer, you don't need to use "bundle-only" because you can force things to remain bundled by just not offering unique addresses for different "m=" sections. Or in the case of ICE, only putting the "ice-ufrag"/"ice-pwd" in one "m=" section.

juberti:
OK. We might need to reiterate this in JSEP.

juberti:
Waiting for @taylor-b 's email to the list to resolve.

fluffy:
talked to chairs and advice from chairs on this and similar things that depend on mmusic outcomes. We put in JSEP now what our best guess is of how mmusic will end up and publish JSEP. At the same time we log a bug to check if we need to make any changes to line up with what mmusic decides. We can take changes to line up with mmusic as Last Call comments later on.

fluffy:
I can't see why subsequent offer should not have the bundle-only. Agree that it is bundles problem but it seems bundle should change to have them.

ekr:
OK I have finally reloaded my state here.

The basic way that BUNDLE works is that you can say:

  a=group:BUNDLE A B
  m=video 9
  a=mid:A
  m=video 9
  a=mid:B

And the answerer can take any of A, B or both. However, because we don't
want to gather for both A and B, we might want to say "you can only take
B if you take A". You do this bu:

- A 0 port which which says "you shouldn't use this m= line"
- An a=bundle-only line which says "actually, you can use this m= line if
  you bundle it with someone else"


I.e.,

  a=group:BUNDLE A B
  m=video 9
  a=mid:A
  m=video 0
  a=mid:B
  a=bundle-only

This means that the answerer can now take A or (A, B) but not B alone.


I take Christer's point to be that once we have done that, if the answerer
took (A, B), if a subsequent offer contains:

  a=group:BUNDLE A B
  m=video 10000
  a=mid:A
  m=video 10000
  a=mid:B

Then the answerer can take any subset of (A, B) again, and this is free
[...]

PR #588, opened 2017-02-08
Don't send bundle-only when re-offering, as described in

ekr:
BUNDLE S 6.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2445,6 +2445,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
             <t>The "a=rtcp-rsize" line MUST only be added if present in
             the most recent answer.</t>
+
+            <t>An "a=bundle-only" line MUST NOT be added, as indicated
+            in <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation"/>,
+            Section 6. Instead, JSEP implementations simply omit ICE
+            parameters for m= lines which can only be used when bundled.
+          </t>
           </list></t>
 
           <t>The "a=group:BUNDLE" attribute MUST include the mid
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2448,8 +2448,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
             <t>An "a=bundle-only" line MUST NOT be added, as indicated
             in <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation"/>,
-            Section 6. Instead, JSEP implementations simply omit ICE
-            parameters for m= lines which can only be used when bundled.
+            Section 6. Instead, JSEP implementations MUST simply omit
+            parameters in the IDENTICAL and TRANSPORT categories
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2445,6 +2445,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
             <t>The "a=rtcp-rsize" line MUST only be added if present in
             the most recent answer.</t>
+
+            <t>An "a=bundle-only" line MUST NOT be added, as indicated
+            in <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation"/>,
+            Section 6. Instead, JSEP implementations simply omit ICE
+            parameters for m= lines which can only be used when bundled.
```
taylor-b: Should the second sentence be phrased as a MUST? Also, would "for m= sections" be more accurate than "for m= lines"?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2445,6 +2445,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
             <t>The "a=rtcp-rsize" line MUST only be added if present in
             the most recent answer.</t>
+
+            <t>An "a=bundle-only" line MUST NOT be added, as indicated
+            in <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation"/>,
+            Section 6. Instead, JSEP implementations simply omit ICE
```
taylor-b: More generally, it's TRANSPORT and IDENTICAL category attributes that are omitted

ekr:
I think we should hold this until Taylor's mmusic comments are resolved.

ekr:
cf. https://mailarchive.ietf.org/arch/msg/mmusic/-jENmu4iaapy7NdJ9_SLHpNjSDc

taylor-b:
It doesn't seem like BUNDLE is going to change at this point; the disadvantages of using "bundle-only" in subsequent offers are that non-BUNDLE-aware middleboxes may be unhappy with a port of "0", and also, the answerer can't reject the offerer BUNDLE-tag section without rejecting the whole group (https://github.com/cdh4u/draft-sdp-bundle/issues/22). I think the latter restriction is unnecessary (and it actually breaks some WebRTC max-bundle corner cases), but it doesn't sound like it's going to be changed.

Speaking of which, if we're being thorough, we should address those corner cases. In "creating an answer", we should specify that if the answer rejects all the offerered m= sections with nonzero ports, it must also reject the m= sections with "a=bundle-only". So, if you're offered transceivers A, B and C with "a=bundle-only", and you do `setRemoteDescription(sdp); A.stop();`, the answer will reject A _and_ B/C.

fluffy:
talked to chairs and advice from chairs on this and similar things that depend on mmusic outcomes. We put in JSEP now what our best gueses is of how mmusic will end up and publish JSEP. At the same time we log a bug to check if we need to make any changes to line up with what mmusic decides. We can take changes to line up with mmusic as Last Call comments later on.

ekr:
@juberti, @taylor-b: I think this should resolve it.

juberti:
Fixes #509

ekr:
@fluffy: this revised version needs your signoff + merge

PR #591, opened 2017-02-09
add ref to ICE TCP

### LINKED ISSUE #570: Do we need a ref to RFC6544
this is  TCP Candidates with Interactive Connectivity Establishment (ICE)

fluffy:
Fixes #570


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1847,7 +1847,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           implementations MUST support both the "UDP/DTLS/SCTP" and
           "TCP/DTLS/SCTP" profiles and MUST indicate one of these two
           profiles for each data m= line they produce in an offer.
-          Because ICE can select either TCP or UDP transport depending
+          Because ICE can select either TCP, specified in <xref
```
juberti: Reads kind of awkward with the 'specified in' here; I think this would be simplest as 

> Because ICE can select either UDP [RFC5245] or TCP [RFC6544] transport depending...

ekr:
@juberti: I cleaned up as you suggested.

PR #593, opened 2017-02-09
Add refs to opus and usedtx

### LINKED ISSUE #571: Do we need refs for opus sdp rfc6716 or rfc7587

fluffy:
Fixes #571


juberti:
Since the second ref is so close to the first, I decided it wasn't needed.

PR #594, opened 2017-02-09
smush down to 72 col

fluffy:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -4564,18 +4564,18 @@ AliceJS->AliceUA:   setRemoteDescription with |answer-C1|
...
 AliceJS->AliceUA:   addIceCandidate with |answer-B1-candidate-1|
 
 //                  ICE and DTLS establish while call is ringing
 
-//                  Bob accepts call, starts media, and sends a new offer
+//                 Bob accepts call, starts media, and sends a new offer
 BobJS->BobUA:       transceiver.setTrack with audio and video tracks
 BobUA->AliceUA:     media sent from Bob to Alice
-BobJS->BobUA:       transceiver.setDirection(sendrecv) for both transceivers
+BobJS->BobUA:       transceiver.setDirection(sendrecv) for both
```
ekr: This seems like it's two short

draft-ietf-rtcweb-jsep.xml
```diff
@@ -4065,11 +4065,11 @@ BobJS->BobUA:       setRemoteDescription with |offer-B1|
 BobUA->BobJS:       ontrack with audio track from Alice
 
 //                  candidates are sent to Bob
-AliceUA->AliceJS:   onicecandidate event with |offer-B1-candidate-1| (host)
```
juberti: I could live with just dropping the parentheticals (and preserving 'event'), but I could go either way.

PR #599, opened 2017-02-11
Codec reoffers

### LINKED ISSUE #266: When should it be possible to add previously discarded codecs in a subsequent offer?

juberti:
Fixes #266

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1911,11 +1911,6 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             corresponding order, and MUST exclude any codecs not present
             in the codec preferences.</t>
 
-            <t>The media formats in the answer MAY include codecs present
-            in the offer that were discarded in a previous offer/answer
-            exchange. This is necessary for compatibility with third-
-            party call control and SIP use cases.</t>
-
             <t>Unless excluded by the above restrictions, the media
             formats MUST include the mandatory audio/video codecs as
             specified in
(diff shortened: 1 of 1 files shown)


juberti:
@taylor-b @pthatcherg FYI

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2343,24 +2338,25 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            lines MUST only include media formats which have not been excluded
+            by the codec preferences of the associated transceiver, and
+            MUST include all currently available formats. Media formats that
+            were previously offered but are no longer available (e.g., a
+            shared hardware codec) MAY be excluded.</t>
 
             <t>Unless codec preferences have been set for the associated
             transceiver, the media formats on the m= line MUST be generated in
-            the same order as in the current local description.</t>
+            the same order as in the most recent answer. Any new media
```
taylor-b: nit: "new" isn't quite the right word, these codecs may have been in previous offers

ekr: agreed, I think you can just remove "new" and it will be fine.

PR #602, opened 2017-02-13
Fix "new" in codec reoffer text

juberti:


juberti:
Merging this, as this change was already discussed and approved in #599.

Issue #528, opened 2017-01-24
Unclear what to do about rtcp-mux/rtcp-rsize if first m= section is for data channels


juberti:
Say you have a session where only a data channel is created. SDP looks like

[session boilerplate]
a=group:BUNDLE d1
m=X application UDP/DTLS/SCTP webrtc-datachannel
a=mid:d1
...

Now, a media m= section is added (i.e. after first o/a). SDP becomes:

[session boilerplate]
a=group:BUNDLE d1 a1
m=X application UDP/DTLS/SCTP webrtc-datachannel
a=mid:d1
a=ice-ufrag:ABCD
...
m=X audio UDP/TLS/RTP/SAVPF 0
a=mid:a1
[no ice-ufrag or TRANSPORT attrs, since bundled]

Now, where does a=rtcp-mux go? It can't go into |d1| because the data section paragraph doesn't allow for it. And it can't go into |a1| because of the language that states

> [IDENTICAL/TRANSPORT] MUST appear only in "m=" sections which either have a unique address or which are associated with the bundle-tag. 

Thoughts on how to resolve?

taylor-b:
Could generating a session description be split into two parts? First, generate media-level attributes for each "m=" section, then transport-level attributes for the BUNDLE group and each non-bundled "m=" section.

juberti:
That would be my preference, i.e., jam it into the BUNDLE group m= section, as that's conceptually simplest, and easily handled here in JSEP.

Either way, I suspect this will require changes upstream in the bundle spec.

fluffy:
So ... is it really not allowed in the d1 section. Unknown attributes are allowed in the d1 section, they are just ignored if they have no meaning for the data channel. Agree this sounds like something that will hit bundle.

ekr:
Couldn't you do a=group:BUNDLE a1 d1

I would have thought this was prohibited, but is it really?
https://tools.ietf.org/html/draft-ietf-mmusic-sdp-bundle-negotiation-36#section-8.5.2

taylor-b:
I think that's legal, but that approach doesn't work in the general case, where "m=" section A has TRANSPORT attributes that only it uses, and section B has TRANSPORT attributes that only it uses.

BUNDLE just needs to allow an "m=" section to contain attributes that it normally wouldn't contain if it wasn't bundled. (If this isn't already allowed).

ekr:
@taylor-b: any action needed here?

taylor-b:
[...]

PR #607, opened 2017-02-15
Note that sometimes data m= sections may contain media appropriate

ekr:
attributes. As far as I can tell this can happen:

- On re-offers
- On any answer because nothing requires the offerer to generate
  bundle tags in the same order as m= lines (even though the
  data channel m= line is last)

See: https://github.com/cdh4u/draft-sdp-bundle/issues/27

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2378,6 +2378,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             for bundled m= sections, as
             described in <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation"/>,
             Section 8.1.
+            Note that if media m= sections are bundled into a
+            data m= section, then the TRANSPORT and IDENTICAL attributes
+            may appear in the data m= section even if they would otherwise
+            only be appropriate for a media m= section (e.g., "a=rtcp-mux").
+            This cannot happen in initial offers because all BUNDLE policies
+            result in at least one m= section without "a=bundle-only" on
+            such offers.
           </t>
           </list></t>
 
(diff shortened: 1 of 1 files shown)


fluffy:
Fixes #528

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2822,7 +2829,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           section MUST also be generated for data. The &lt;media&gt;
           field MUST be set to "application" and the &lt;proto&gt; and
           "fmt" fields MUST be set to exactly match the fields in the
-          offer.</t>
+          offer. Note that if media m= sections are bundled into a
+          data m= section, then the TRANSPORT and IDENTICAL attributes
```
juberti: As not all attributes are relevant here, suggest that you replace 'the' with 'certain'

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2378,6 +2378,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             for bundled m= sections, as
             described in <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation"/>,
             Section 8.1.
+            Note that if media m= sections are bundled into a
```
juberti: I would make this its own para, as it's only loosely connected to a=bundle-only.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2378,6 +2378,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            data m= section, then the TRANSPORT and IDENTICAL attributes
```
juberti: See below; suggest replacing 'the' with 'certain'.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2378,6 +2378,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            may appear in the data m= section even if they would otherwise
+            only be appropriate for a media m= section (e.g., "a=rtcp-mux").
+            This cannot happen in initial offers because all BUNDLE policies
```
juberti: This explanation is not obvious to me. Yes, at least one m= section must not have a=bundle-only, but that is a truism. The salient point is that initial offers always try to bundle into a non-data section (regardless of bundle policy), or have no media attribs (e.g. the data-only case). I think it would be good to expound on this a bit more.

Also, BUNDLE->bundle.

ekr:
@juberti: PTAL

PR #609, opened 2017-02-18
Data m= section clarification

### LINKED ISSUE #185: mediastream vs datachannels in "initial offers"
The "Initial Offers" section does not distinguish systematically between requirements that apply for any type of offer and the ones that apply only to mediastreams.

For instance:

> The next step is to generate m= sections, as specified in [RFC4566]  Section 5.14, for each MediaStreamTrack that has been added to the   PeerConnection via the addStream method

I think this should apply both to mediastreamtrack and to datachannels. 

Then many of the rules described under "Each m= section MUST include the following attribute lines:" seem specific to mediastreams, but not all; and that set of rules is then followed by a statement that a m= section must be created for data channels.

This could use some reorganization and clarification.

juberti:
Fixes #185

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2866,9 +2866,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             </list>
           </t>
 
-          <T>Note that if media m=
-          sections are bundled into a data m= section, then certain attributes
-          may appear in the data m= section even if they would otherwise
+          <t>Note that if media m= sections are bundled into a
+          data m= section, then certain TRANSPORT and IDENTICAL attributes
+          may also appear in the data m= section even if they would otherwise
           only be appropriate for a media m= section (e.g., "a=rtcp-mux").
           </t>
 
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2133,10 +2133,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           </t>
 
-          <t>As discussed above, the following attributes are included only
-          if the m= section either has a unique address or is associated
-          with the bundle-tag: "a=ice-ufrag",
-          "a=ice-pwd", "a=fingerprint", "a=dtls-id", and "a=setup".
+          <t>As discussed above, the following attributes of category IDENTICAL
+          or TRANSPORT are included only if the data m= section either has
+          a unique address or is associated with the bundle-tag (e.g., if
+          it is the only m= section):
```
ekr: I'm not sure this parenthetical is helping, because it seems like if it is really the only m= section then it's not going to be associated with the bundle tag, more likely you just wouldn't bundle

juberti: According to the current text, you insert a bundle group regardless of the number of m sections. Or did you mean something else?

juberti:
@ekr, PTAL at this. I realized the answer section also needed to be updated to match the new offer text, so I made some changes there which involved a bit of massaging of the text added in #607.

PR #614, opened 2017-02-20
Clarify what things are allowed to change from pranswer to next answer.

### LINKED ISSUE #600: Are there any uses of "pranswer" that aren't required to be supported?
Changing the codecs, header extensions, etc. from the provisional answer to the real answer is, in theory, pretty simple to implement. Other things, however - related to connection-oriented protocols - are more complex.

- Changing the ICE ufrag/password. Should this result in a non-standard "half" ICE restart (continuing to use the previously selected candidate pair), or should it result in ICE state being completely restored to the point before the "pranswer" was applied? Also, if this is supported, it means ICE candidates can't be deallocated until a final answer is applied. This may be already implied by "should not result in the freeing of allocated resources."
- Changing the DTLS fingerprint and/or "a=setup". Should this result in a new DTLS association? This involves some non-trivial interactions with ICE; see [dtls-sdp](https://tools.ietf.org/html/draft-ietf-mmusic-dtls-sdp-16#section-6). Also, if two "pranswers" that establish new DTLS associations are in-flight (due to slow signaling), it becomes hard to tell which association a given "Client Hello" is for, unless I'm not considering something. A robust implementation would need to have multiple DTLS associations going in parallel, one per remote ICE ufrag, and throw them away as answers arrive and it can be determined which one is the oldest.
- Changing the SCTP port. Does this result in a new SCTP association?

Are these things intended to be supported? If so, it might be worthwhile clarifying a bit how they should work.

taylor-b:
Fixes #600.

ICE credentials, DTLS fingerprint and SCTP port are allowed to change,
which is necessary to support sequential forking with pranswers.


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3785,11 +3793,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           <t>If the remote DTLS fingerprint has been changed or
           the dtls-id has changed, tear down the DTLS connection.
-          If a DTLS connection needs to be torn down but the answer
-          does not indicate an ICE restart, an error MUST be generated.
-          If an ICE restart is performed without a change in dtls-id
-          or fingerprint, then the same DTLS connection is continued over
-          the new ICE channel.</t>
+          This includes when the PeerConnection state is
+          "have-remote-pranswer". If a DTLS connection needs to be torn
+          down but the answer does not indicate an ICE restart, or new
```
juberti: I would write this as 

> "or, in the case of "have-remote-pranswer", new ICE credentials"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3882,6 +3892,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>If the m= section proto value indicates use of SCTP:
           <list style="symbols">
+            <t>If an SCTP association exists, and the remote SCTP port
+            has changed, discard the existing SCTP association. This
+            includes when the PeerConnection state is
```
juberti: "includes when" -> "includes the case when"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3882,6 +3892,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            "have-remote-pranswer"</t>
```
juberti: Missing period

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3785,11 +3793,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>If the remote DTLS fingerprint has been changed or
           the dtls-id has changed, tear down the DTLS connection.
-          If a DTLS connection needs to be torn down but the answer
-          does not indicate an ICE restart, an error MUST be generated.
-          If an ICE restart is performed without a change in dtls-id
-          or fingerprint, then the same DTLS connection is continued over
-          the new ICE channel.</t>
+          This includes when the PeerConnection state is
```
juberti: "includes when" -> "includes the case when"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3610,18 +3610,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          <xref target="RFC5245" />. An answer MUST change the ufrag
-          and password in an answer if and only if ICE is restarting,
-          as described in Section 9.2.1.1 of
-          <xref target="RFC5245" />.</t>
+          <t>If the PeerConnection state is "have-local-offer", and the
+          ICE ufrag or password changed from the previous remote
+          description, then an ICE restart is needed, as described in
+          Section 9.1.1.1 of <xref target="RFC5245" />. If the
+          description is of type "offer", mark that an ICE restart is
+          needed. If the description is of type "answer" or "pranswer"
```
taylor-b: But this only applies to a remote answer, so it's equally at home in "applying a remote description" or "applying an answer". And "applying an answer" doesn't currently say anything about ICE.

Since this isn't new text (all I did was add "if the state is have-local-offer") could this be handled in a separate PR?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3610,18 +3610,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          and the current local description is also an ICE restart, then
+          signal the ICE agent to begin checks as described in Section
+          9.3.1.1 of <xref target="RFC5245" />. An answer MUST change
```
taylor-b: It said "answer" twice before this PR. :)

I'll fix it, but see above comment about moving it.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3610,18 +3610,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          the ufrag and password in an answer if and only if ICE is
+          restarting, as described in Section 9.2.1.1 of <xref
+          target="RFC5245" />.</t>
+
+          <t>If the PeerConnection state is "have-remote-pranswer", and
+          the ICE ufrag or password changed from the previous
+          provisional answer, then signal the ICE agent to discard any
+          previous ICE checklist state for the m= section and begin
```

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml

fluffy:
I still need to read this all carefully but at first glance looks reasonable to me.

draft-ietf-rtcweb-jsep.xml
juberti: We do handle similar cases for DTLS and SCTP in the answer section, so I think ultimately this text may need to move there as well. But we can consider it out of scope for now.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3882,6 +3892,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>If the m= section proto value indicates use of SCTP:
           <list style="symbols">
+            <t>If an SCTP association exists, and the remote SCTP port
+            has changed, discard the existing SCTP association. This
+            includes the case when the PeerConnection state is
+            "have-remote-pranswer".</t>
 
             <t>If no SCTP association yet exists, prepare to initiate a
```
juberti: This sentence probably needs to change to indicate that the SCTP association needs to be created at this point if it was just destroyed above.

juberti: Possibly, use language similar to the DTLS section above, e.g. "If no valid SCTP association exists"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3610,18 +3610,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          restarting, as described in Section 9.2.1.1 of <xref
+          target="RFC5245" />.</t>
+
+          <t>If the PeerConnection state is "have-remote-pranswer", and
+          the ICE ufrag or password changed from the previous
+          provisional answer, then signal the ICE agent to discard any
+          previous ICE check list state for the m= section and begin
+          checks as if this were the first answer. However, such an
+          answer MAY only change the ICE ufrag or password if the local
+          offer is starting or restarting ICE for that m= section.</t>
```
juberti: "that" seems out of place here. "the m= section" makes it more clear which m= section we're talking about, I think.

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml

juberti:
LGTM. @fluffy, do you want to take another pass?
@ekr, do you want to take a look?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3610,18 +3610,26 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          to begin checks as described in Section 9.3.1.1 of
-          <xref target="RFC5245" />. An answer MUST change the ufrag
-          and password in an answer if and only if ICE is restarting,
-          as described in Section 9.2.1.1 of
-          <xref target="RFC5245" />.</t>
+          <t>If the PeerConnection state is "have-local-offer", and the
+          ICE ufrag or password changed from the previous remote
+          description, then an ICE restart is needed, as described in
+          Section 9.1.1.1 of <xref target="RFC5245" />. If the
+          description is of type "offer", mark that an ICE restart is
```
ekr: Maybe "note" that an ICE

taylor-b: Meaning, replace "mark" with "note"?

juberti: That was my interpretation.

PR #616, opened 2017-03-03
Add ref for CVO

### LINKED ISSUE #613: Do we need a reference for CVO?
We mention it in the imageattr section.

          This is
          required regardless of whether the receiver supports
          performing receive-side rotation (e.g., through CVO), as it
          significantly simplifies the matching logic.

fluffy:
Fixes #613

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -744,7 +744,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           that when considering a MediaStreamTrack that is producing
           rotated video, the unrotated resolution MUST be used. This is
           required regardless of whether the receiver supports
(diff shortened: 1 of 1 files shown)


fluffy:
Note that the use of CVO in webrtc is described in https://tools.ietf.org/html/rfc7742#section-4

fluffy:
@juberti, feel free to rewrite however and merge. I just wanted to get the xref in.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -5066,6 +5067,20 @@ a=msid:bbce3ba6-abfc-ac63-d00a-e15b286f8fce
   </middle>
   <back>
     <references title="Normative References">
+
+      <reference anchor="TS26.114" target="http://www.3gpp.org/DynaReport/26114.htm">
```
juberti: Does this need to be normative? I would have thought this was informative.

PR #617, opened 2017-03-03
update Acknowledgements

fluffy:
This adds people I thought were missing and sorts by last  name.


juberti:
@fluffy, @ekr, any further comments before I land this?

PR #620, opened 2017-03-10
Add explainer paragraph regarding why PT demux is part of the algorithm

### LINKED ISSUE #619: Is the PT table in the demux algo necessary?
BUNDLE mandates sending MID in RTP headers and RTCP messages. What's the rationale for having a second demux mechanism?

juberti:
Fixes #619

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -6437,6 +6437,14 @@ a=msid:bbce3ba6-abfc-ac63-d00a-e15b286f8fce
       routing that are updated each time an RTP/RTCP packet contains
       new information that affects how packets should be routed.</t>
 
+      <t>However, some pre-standard implementations of <xref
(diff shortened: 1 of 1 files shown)


fluffy:
This is not why we do this. If you add the MID for audio calls it ups the peak bandwidth which increases the amount of bandwidth you need to reserver for each call on a satellite link. So by using PT instead of MID, you get more calls on the sat link. Theses are calls where is tracks but few enough it is nowhere even close to running out PTs.

juberti:
OK, I got rid of the 'legacy' terminology. LMK if you want to include the bandwidth rationale (which seems like a modest improvement and may need a detailed explanation)

fluffy:
I like your new text. LGTM

juberti:
@ekr, LMK if you approve or have no opinion on this

ekr:
LGTM.

PR #625, opened 2017-03-21
remove a=ssrc-group requirements from media level parsing

fippo:
removes parsing of a=ssrc and a=ssrc-group. The createOffer parts were removed in the -17 draft.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3368,7 +3368,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             stored.</t>
             
             <t>Any "a=ssrc" attributes MUST be parsed
-            as specified in <xref target="RFC5576" />, Sections 4.1,
+            as specified in <xref target="RFC5576" />, Section 4.1,
             and their values stored.</t>
 
             <t>Any "a=extmap" attributes MUST be parsed as specified in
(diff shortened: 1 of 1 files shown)


nils-ohlmeier:
Wouldn't it make more sense to add the ssrc-group back into the draft to allow proper handling of RTX.

fluffy:
I'm fine with removing this

juberti:
Not sure why we need to remove this since this can be used to indicate how to init the demux table.

juberti:
I do think removing ssrc-group would be a good idea though since we don't plan to use it. We should also make clear that a=ssrc is only used for seeding the demux table.

juberti:
To be specific: 5.9 (remote desc) should be updated to indicate that any a=ssrc values are supplied to the demuxer, and we should remove discussion of a=ssrc-group here.

fluffy:
You arguments to keep this make sense. I'm fine with keeping it.

juberti:
Changed it to only remove a=ssrc-group, since IIUC the demux algorithm does not consider it. @nils-ohlmeier, does this seem reasonable to you?

Separately I will add text in 5.9 to say that the a=ssrc values should be supplied to the demuxer.

nils-ohlmeier:
It was non-obvious to me that RTX has to use the same MID RTP header extension values as the RTP stream it covers. Now the same argument for priming the demuxer could be made for the SSRC learned from ssrc-group. But if we assume that RTX packets will not be the first packets which you receive it's probably fine.

PR #627, opened 2017-03-24
Add delay to SSRC removal from an RTCP BYE.

pthatcherg:
This is based on the feedback from Magnus.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -6729,7 +6729,12 @@ a=msid:bbce3ba6-abfc-ac63-d00a-e15b286f8fce
           each SSRC indicated in the packet that is found in the
           incoming SSRC table, first deliver a copy of the packet to
           the "m=" line associated with that SSRC, but then remove the
-          entry for that SSRC from the incoming SSRC table.</t>
+          entry for that SSRC from the incoming SSRC table after an
+          appropriate delay (Similar to <xref target="RFC3550" />
+          section 6.2.1, some "straggler packets" might arrive after
+          the BYE).
+
+          </t>
 
           <t>If the packet is of type SR or RR, for each report block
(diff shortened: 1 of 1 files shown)


aboba:
LGTM.

gloinul:
Yes, works for me.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -6729,7 +6729,12 @@ a=msid:bbce3ba6-abfc-ac63-d00a-e15b286f8fce
           each SSRC indicated in the packet that is found in the
           incoming SSRC table, first deliver a copy of the packet to
           the "m=" line associated with that SSRC, but then remove the
-          entry for that SSRC from the incoming SSRC table.</t>
+          entry for that SSRC from the incoming SSRC table after an
+          appropriate delay (Similar to <xref target="RFC3550" />
```
juberti: I'm not a fan of normative language in parentheticals. Suggest rephrasing as:

    entry for that SSRC from the incoming SSRC table after an
    appropriate delay to account for "straggler packets", as 
    mentioned in the discussion of BYE handling in
    <xref target="RFC3550" />, Section 6.2.1.

juberti:
LGTM with comment

draft-ietf-rtcweb-jsep.xml
alvestrand: Should that be "as specified in 3550"? We're not adding new behavior here, we're reminding people what 3550 says on how to handle BYE.

fluffy: I think we should make change HTA proposes

pthatcherg:
I have addressed the comments.  PTAL

alvestrand:
LGTM

pthatcherg:
FYI, I updated https://github.com/cdh4u/draft-sdp-bundle/pull/29 to match.

PR #628, opened 2017-03-24
Make limitations of PT-based demux more clear, based on Magnus's feed…

pthatcherg:
This is based on the feedback from Magnus.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -6604,8 +6604,8 @@ a=msid:bbce3ba6-abfc-ac63-d00a-e15b286f8fce
       payload type values, in order for the payload type to be a
       reliable indicator of the relevant "m=" line for the RTP stream.
       Note that when using payload type based demuxing, an SSRC
-      will be mapped to an "m=" line by first packet with that
-      SSRC and the mapping will not be changed even if the same
+      will be mapped to an "m=" line by the first packet with that
+      SSRC, and the mapping will not be changed even if the same
       SSRC is received with a different payload type.  In other
       words, the SSRC cannot to "move" to a different "m=" line
       simply by changing the payload type.
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -6602,9 +6602,7 @@ a=msid:bbce3ba6-abfc-ac63-d00a-e15b286f8fce
       when using BUNDLE, and instead use a payload type based mechanism
       for demuxing. In this situation, each "m=" line MUST use unique
       payload type values, in order for the payload type to be a
-      reliable indicator of the relevant "m=" line for the RTP
-
-      stream.
+      reliable indicator of the relevant "m=" line for the RTP stream.
       Note that when using payload type based demuxing, an SSRC
       will be mapped to an "m=" line by first packet with that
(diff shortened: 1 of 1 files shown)


aboba:
SSRCs can move between m-lines in some situations, right?

If the application calls transceiver.stop(), a BYE will be sent for the SSRC and after negotiation, the m-line will have a zero port number. If a new O/A exchange were to put that former-SSRC on a new m-line, it will be signaled and installed into the SSRC table without issue, correct?

pthatcherg:
True.  I guess we'd have to clarify that by it can't move by simply sending RTP packets with the PT tied to another m= section.  Some kind of reconfiguration/renegotiation would be needed.

fluffy:
I don't really care if we add something this but this seems 100% clear from the existing text so I really don't see the point of adding it.

pthatcherg:
Bernard, I made it more clear that there are other mechanisms for "moving" an SSRC, such as the one you mentioned.  I also made the text more explicit.  I agree with Cullen that this isn't really needed, since it's clear from the algorithm.  But Magnus asked for it be extra super explicit, so I'm willing to add it.

gloinul:
Thanks, I would probably not have added as much. But, I think there is a point to make clear the limitations. I am not that happy about the added text in response to Bernard's comment. Yes, this could occur, but a re-added stream after having called stop (BYE) is a new RTP stream, and thus a new random SSRC, even if signalled explicit. I don't want to encourage behaviors with potential downsides. I think using MID is the obvious alternative that should be noted if anything. And that goes back to my initial motivation for this text, that MID is the better alternative.

pthatcherg:
Magnus, I made the text much shorter.  I also removed the "example" that would encourage people in the wrong direction.

gloinul:
This looks good to me.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -6602,8 +6602,14 @@ a=msid:bbce3ba6-abfc-ac63-d00a-e15b286f8fce
       when using BUNDLE, and instead use a payload type based mechanism
       for demuxing. In this situation, each "m=" line MUST use unique
       payload type values, in order for the payload type to be a
-      reliable indicator of the relevant "m=" line for the RTP
-      stream.</t>
+      reliable indicator of the relevant "m=" line for the RTP stream.
+      Note that when using payload type based demuxing, an SSRC
+      will be mapped to an "m=" line by first packet with that
```
juberti: "the first packet"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -6602,8 +6602,14 @@ a=msid:bbce3ba6-abfc-ac63-d00a-e15b286f8fce
...
+      SSRC and the mapping will not be changed even if the same
```
juberti: comma after 'SSRC'

fluffy:
LGTM

draft-ietf-rtcweb-jsep.xml
pthatcherg: fixed

draft-ietf-rtcweb-jsep.xml
pthatcherg: fixed

pthatcherg:
FYI, I updated https://github.com/cdh4u/draft-sdp-bundle/pull/29 to match.

PR #629, opened 2017-03-26
Clarify that when generating an answer the proto field value

gloinul:
Proposal for how to clarify that when JSEP implementation generates an answer it needs to have the proto value identical to what is in the offer.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1784,8 +1784,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <t>
             <list style="symbols">
 
-              <t>The profile in any "m=" line in any answer MUST
-              exactly match the profile provided in the offer.</t>
+              <t>The profile in any "m=" line in any answer generated 
+			  by the JSEP implementation MUST exactly match the 
+			  profile provided in the offer.</t>
 
               <t>Any profile matching the following patterns MUST be
               accepted: "RTP/[S]AVP[F]" and
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1779,19 +1779,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           "(UDP,TCP)/TLS/RTP/SAVPF". In order to simplify compatibility
           with such endpoints, JSEP implementations MUST follow the
           following rules when processing the media m= sections in an
-          offer:</t>
+          received offer:</t>
 
           <t>
             <list style="symbols">
 
+              <t>Any profile in the offer matching the following patterns
+			  MUST be accepted: "RTP/[S]AVP[F]" and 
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1784,8 +1784,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <t>
             <list style="symbols">
 
-              <t>The profile in any "m=" line in any answer MUST
-              exactly match the profile provided in the offer.</t>
+              <t>The profile in any "m=" line in any answer generated
```
juberti: The text above already makes it clear that this is about the JSEP implementation's processing of an offer, so this seems redundant.

However, I could live with "in any generated answer"

gloinul: The issue with the text "In order to simplify compatibility
          with such endpoints, JSEP implementations MUST follow the
          following rules when processing the media m= sections in an
          offer:"
is that is not clear if it is generating the offer, or processing a received offer to generate an answer. So it might be the quoted sentence that should be clarified by adding "... in a received offer:"?

juberti:
@fluffy, do you want to take a look at this one?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1778,19 +1778,20 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-              <t>The profile in any "m=" line in any answer MUST
-              exactly match the profile provided in the offer.</t>
-
-              <t>Any profile matching the following patterns MUST be
-              accepted: "RTP/[S]AVP[F]" and
-              "(UDP/TCP)/TLS/RTP/SAVP[F]"</t>
-
+              <t>Any profile in the offer matching the following patterns
+              MUST be accepted: "RTP/[S]AVP[F]" and 
+              "(UDP/TCP)/TLS/RTP/SAVP[F]".
```
martinthomson: this pattern could be read as matching "UDP/TCP/TLS/RTP/SAVPF"  I think that you want this "RTP/[S]AVP[F]" and "{UDP|TCP}/TLS/RTP/SAVP[F]" (that is, use braces for the first grouping and a vertical bar instead of a slash).

juberti: I agree with the vertical bar, but what's the rationale for braces vs parens?

martinthomson: Pick one. I probably had the font too small when I wrote the suggestion and thought that I had copied it.  You probably need to explain the significance of the special characters when you use them.  Or you could use a regex...

juberti: OK. Will tweak accordingly.

juberti:
Fixes #652

draft-ietf-rtcweb-jsep.xml
fluffy: How about we use posix basic regex as specified at https://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended

PR #685, opened 2017-04-29
Mention that candidate fields don't contain a=

### LINKED ISSUE #645: Clarify ICE candidate format
Since it has been the cause of actual interop issues in the field, it is probably worth mentioning, in section 3.5.2.1, that the IceCandidate strings explicitly *do not* start with "a=".

juberti:
Fixes #645

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -561,7 +561,9 @@ setRemote(ANSWER) |                                   |
             <t>The candidate details are specified in an IceCandidate
             field, using the same SDP syntax as the
             "candidate-attribute" field defined in
-            <xref target="RFC5245" />, Section 15.1. For example:</t>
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -561,7 +561,9 @@ setRemote(ANSWER) |                                   |
             <t>The candidate details are specified in an IceCandidate
             field, using the same SDP syntax as the
             "candidate-attribute" field defined in
-            <xref target="RFC5245" />, Section 15.1. For example:</t>
+            <xref target="RFC5245" />, Section 15.1. Note that this field
+            does not contain an "a=" prefix, as indicated in
+            the following example:</t>
```
ekr: Nit: "As indicated in the following example, this field does not contain an "a=" prefix"

PR #690, opened 2017-05-05
rephrase 'MUST only'

### LINKED ISSUE #660: MUST only is confusing
The bullets on page 45 use the construct "MUST only" three times. This is confusing, and likely to lead to implementors doing the wrong thing. Please rewrite in the form "The 'a=rtcp' line MUST NOT be added unless..."

fluffy:
Fixes #660 

Please review carefully


juberti:
I'm not sure this reformulation is an improvement - seems like a lot of double negatives. @ekr, wdyt?

ekr:
I think I agree with Justin. I guess I can take a pass at this if Fluffy feels like he has taken his best shot and we don't like it. @fluffy?

fluffy:
Yah - @ekr - see what you would do and either do it or just give me a template of how to eal with one without the double negatives

ekr:
Closing in favor of #706

PR #691, opened 2017-05-05
remove irrelevant error checking from old stuff

### LINKED ISSUE #664: Clarify parsing the local session description
Section 5.5, bullet 4 talks about parsing the local session description. This is legacy from when SDP munging was allowed, and should be removed (especially the normative statement about parsing errors: the check that the SDP has not changed will catch changes, and any implementation that cannot parse its own SDP is exhibiting an internal error, not a protocol error).

fluffy:
Fixes #664

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3098,6 +3098,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <xref target="sec.modifying-sdp" />; otherwise, processing
           MUST stop and an error MUST be returned.</t>
 
+          <t>Next, the SessionDescription is parsed into a data
+          structure, as described in the <xref
+          target="sec.parsing-a-desc" /> section below. </t>
+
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3098,12 +3098,6 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <xref target="sec.modifying-sdp" />; otherwise, processing
           MUST stop and an error MUST be returned.</t>
 
-          <t>Next, the SessionDescription is parsed into a data
```
juberti: I think we still want this sentence, but we can skip the second one.

ekr:
It seems like there are two implementation strategies here:

1. Keep a parsed version
2. Re-parse.

I think we probably could dispense with this sentence, but I don't object to it, because I think of "lookup in cache" as "parse".

juberti:
Knowing how things go in the real world, I'm a bit reluctant to make this change, but it should address Adam's concern.

PR #695, opened 2017-05-05
clarify when to add actpass for DTLS

### LINKED ISSUE #663: Editorial regarding a=setup
Page 52, bullet starting "An 'a=setup' line": add the words "when allowed" to the end of the bullet.

fluffy:
Fixes #663


ekr:
I think what he means is that this text actually covers both initial and subsequent offers and then you cannot use actpass. I am still double-checking if that's true.

fluffy:
So more discussion in the bug but seems like we should reject this PR and leave it as is

PR #696, opened 2017-05-05
Encourage parsing other attributes. Fixes #667

### LINKED ISSUE #667: Unknown attribute handling
Section 5.7.2 should include a statement (probably at the end) that implementations may know and process attributes other than those described in this section.

ekr:
Not sure I love what I have done here.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3268,9 +3268,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             stored.</t>
           </list></t>
 
-          <t>As required by
-          <xref target="RFC4566"></xref>, Section 5.13, unknown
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3265,7 +3265,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             stored.</t>
           </list></t>
 
-          <t>As required by
+          <t>This list is non-exhaustive and implementations SHOULD
```
juberti: "This list is non-exhaustive" makes it seem like we may have overlooked something. I would write this as:

> Other attributes that are not relevant to JSEP may also be present, and implementations SHOULD process any that they recognize. As required by...

Note also that this text should also be included in the handling of media-level attrs below (line 3436).

PR #702, opened 2017-05-05
clarify when to use active for DTLS role

### LINKED ISSUE #663: Editorial regarding a=setup

fluffy:
Fixes #663

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2841,7 +2841,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               for use in DTLS-SRTP scenarios in
               <xref target="RFC5763"></xref>, Section 5. The role value
               in the answer MUST be "active" or "passive"; the "active"
-              role is RECOMMENDED.</t>
+              role is RECOMMENDED if "actpass" was offered.</t>
 
               <t>An "a=tls-id" line, as specified in
               <xref target="I-D.ietf-mmusic-dtls-sdp" /> Section
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2841,7 +2841,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               for use in DTLS-SRTP scenarios in
               <xref target="RFC5763"></xref>, Section 5. The role value
               in the answer MUST be "active" or "passive". When
-              the the offer contains the "actpass" value, as
+              the offer contains the "actpass" value, as
               will always be the case with JSEP endpoints,
               the answerer SHOULD use the "active" role.
               Offers from non-JSEP endpoints MAY send other values for
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2840,8 +2840,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <xref target="RFC4145"></xref>, Section 4, and clarified
               for use in DTLS-SRTP scenarios in
               <xref target="RFC5763"></xref>, Section 5. The role value
-              in the answer MUST be "active" or "passive"; the "active"
-              role is RECOMMENDED if "actpass" was offered.</t>
+              in the answer MUST be "active" or "passive", and MUST
+              be consistent with the value in the offer. Where
+              possible, the "active" role is RECOMMENDED. This
+              will always be the case with JSEP endpoints, which
+              MUST send "actpass", as required <xref target="sec-create-offer"/>,
+              but non-JSEP endpoints MAY send other values for
+              "a=setup".</t>
 
               <t>An "a=tls-id" line, as specified in
               <xref target="I-D.ietf-mmusic-dtls-sdp" /> Section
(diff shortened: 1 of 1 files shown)


juberti:
Isn't actpass always offered? Not sure I understand the context here.

nils-ohlmeier:
My understanding is the JSEP compliant implementations are suppose to always use 'actpass' in offers. But you may interop with something which will only present it's current role in a subsequent offer.

juberti:
If that's the intent, I think that rationale should be spelled out in the text here.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2840,8 +2840,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <xref target="RFC4145"></xref>, Section 4, and clarified
               for use in DTLS-SRTP scenarios in
               <xref target="RFC5763"></xref>, Section 5. The role value
-              in the answer MUST be "active" or "passive"; the "active"
-              role is RECOMMENDED.</t>
+              in the answer MUST be "active" or "passive", and MUST
+              be consistent with the value in the offer. Where
```
juberti: I think the best thing to say would be that when "actpass" is presented, "active" is RECOMMENDED. Otherwise, replying with the correct value is a MUST.

ekr:
@juberti

fluffy:
LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2840,8 +2840,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <xref target="RFC4145"></xref>, Section 4, and clarified
               for use in DTLS-SRTP scenarios in
               <xref target="RFC5763"></xref>, Section 5. The role value
-              in the answer MUST be "active" or "passive"; the "active"
-              role is RECOMMENDED.</t>
+              in the answer MUST be "active" or "passive". When
+              the the offer contains the "actpass" value, as
+              will always be the case with JSEP endpoints,
+              the answerer SHOULD use the "active" role.
```
juberti: But, this isn't always the right guidance, since in a re-offer scenario the value may need to be either active or passive.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2840,8 +2840,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <xref target="RFC4145"></xref>, Section 4, and clarified
               for use in DTLS-SRTP scenarios in
               <xref target="RFC5763"></xref>, Section 5. The role value
-              in the answer MUST be "active" or "passive"; the "active"
-              role is RECOMMENDED.</t>
+              in the answer MUST be "active" or "passive". When
+              the the offer contains the "actpass" value, as
```
juberti: typo: the the

ekr:
OK, so this took a bit of research to get clear on.

First, let me state what I believe the rules are:
1. JSEP endpoints MUST always send actpass in offers. This is defined in http://rtcweb-wg.github.io/jsep/#rfc.section.5.2.1 and never modified in 5.2.2

2. In initial answers, JSEP endpoints 
  (a) MUST send active or passive.
  (b) if the other side sent something other than actpass, MUST send the complement
  (c) if the other side sent actpass, SHOULD send active.
http://rtcweb-wg.github.io/jsep/#sec.initial-answers

3. In subsquent answers, JSEP endpoints:
  (a) MUST send active or passive.
  (b) MUST send something consistent with the current state
http://rtcweb-wg.github.io/jsep/#sec.subsequent-answers

The text we are working on here is the initial answer text, therefore, I think that rule 3(b) does not apply, which, I believe, makes the current text correct.

ekr:
@juberti

fluffy:
The explanation from EKR above looks right to me

ekr:
Actually, 5763 requires it to always be actpass, so we can remove the non-JSEP section.

Ugh, but the revision does not require it:
https://tools.ietf.org/html/draft-ietf-mmusic-dtls-sdp-24

ekr:
I think we have two options here:

1. Get dtls-sdp modified (the offending text is in)
https://tools.ietf.org/html/draft-ietf-mmusic-dtls-sdp-24#section-5.2
https://tools.ietf.org/html/draft-ietf-mmusic-dtls-sdp-24#section-5.5

2. Make JSEP handle this condition, as above.

Does anyone know why we changed the rules in dtls-sdp? @cdh4u: ^^?

cdh4u:
I would have to check the archives in order to find out why we relaxed the must-use-actpass-in-initial-offer rule. Most likely people didn't see a need for not allowing other values.

Having said that, JSEP can always mandate sending "actpass" in initial offers. A JSEP implementation must of course still be prepared to receive something else.

juberti:
That's probably fine, as long as dtls-sdp ensures that everyone has to handle actpass in offers. I don't want people trying to avoid implementing part of the spec by only supporting passive mode or some such.

cdh4u:
If you think there is some text indicating that support of "actpass" is optional, that should be corrected.

ekr:
OK, after all that discussion, I think we need to keep the text that says only JSEP endpoints are guaranteed to do this, but they wwill always do so, which means, I believe that the current text is right, modulo the repeated "the" which the new commit fixes.

PR #703, opened 2017-05-05
Add redundant clarifying text to address review comment. Fixes #672

### LINKED ISSUE #672: Unsupported media formats
The bullet in the middle of page 66 (starting "For each specified media format...") states that unsupported media formats must be ignored. The following bullet specifies that "rtx" media formats that refer to "primary payload types [that] are not present" MUST result in an error. Read naïvely, implementors may very well discard a payload type for lack of understanding, and then discover an "rtx" for that discarded PT, and generate an error ("it's not present!"). These bullets need additional text regarding how they interact with each other.

ekr:
Fixes #672

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3729,7 +3729,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             primary payload type, as described in
             <xref target="RFC4588" />, Section 4. If any referenced
             primary payload types are not present, this MUST result in
-            an error.</t>
+            an error. Note that RTX payload types may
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3729,7 +3729,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             primary payload type, as described in
             <xref target="RFC4588" />, Section 4. If any referenced
             primary payload types are not present, this MUST result in
-            an error.</t>
+            an error. Note that RTX payload types may
+            refer to primary payload types which are not supported
+            by the local media implementation
```
juberti: comma after implementation

Issue #670, opened 2017-04-20
Editorial on RtpTransceiver


juberti:
Section 5.8 contains the sentence: "If there is no RtpTransceiver associated with this m= section (which will only happen when applying an offer)..." -- it is grammatically ambiguous whether this says that the *absence* of an RtpTranceiver can only happen with an offer, or if the *presence* of an RtpTranceiver can only happen with an offer. Suggest rephrasing as: "If applying an offer and there is no RtpTransceiver associated with this m= section, find one..."

juberti:
@pthatcherg, can you take a look at this?

juberti:
I'm not sold on this approach, since it then invites the question, what should one do if this condition happens when applying an answer?

My suggestion would be to still call this out explicitly, but try to avoid the ambiguity noted above by changing the parenthetical to its own sentence at the end of the paragraph, e.g.

"Note that this situation will only occur when applying an offer."

pthatcherg:
I changed it to that.

PR #707, opened 2017-05-05
Make RtpTransciever language less ambiguous

pthatcherg:
Fixes https://github.com/rtcweb-wg/jsep/issues/670.


juberti:
See comment in #670

pthatcherg:
Alright, try this version

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3532,9 +3532,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <list style="symbols">
 
             <t>If there is no RtpTransceiver associated with this m=
-            section (which will only happen when applying an offer),
-            find one and associate it with this m= section according to
-            the following steps:
+            section, find one and associate it with this m= section
+            according to the following steps.  Note that this should
```
juberti: 'should' seems problematic to me... when is it not the case?

Suggest
> Note that this **situation will only occur** when applying an offer.

PR #709, opened 2017-05-06
Bundle-only always gets port 0. Fixes #654

### LINKED ISSUE #654: Clarify statement regarding trickle ICE ports #wglc-punchlist
BLOCKER: The first bullet on page 37 says that port 9 "MUST be used," which contradicts the behavior required of BUNDLE-only sections (for which port 0 MUST be used). Clarify.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1910,7 +1910,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           itself, the following rules MUST be followed:
           <list style="symbols">
 
-            <t>The port value is set to the port of the default ICE
+            <t>If the m= section is marked as bundle-only, then
+            the port value is set to 0. Otherwise,
```
juberti: MUST be set to 0?

PR #710, opened 2017-05-06
Describe what can arrive before an answer

### LINKED ISSUE #639: Investigate potential problem in Section 3.2
Section 3.2 has a paragraph starting "Lastly, while the actual media parameters are only know after an offer..."; the entire contents of this paragraph appear to be actually false if both ICE and DTLS are in use. I know there's conversation underway on MMUSIC and the WebRTC mailing lists on this topic, but none of them have yet described a situation in which this paragraph is actually true. Please remove it or add a sequence diagram showing how it arises.

juberti:
Fixes #639

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -285,9 +285,10 @@ ipr="trust200902">
         contain the exact same number as the offer.</t>
 
         <t>Lastly, while the exact media parameters are only known only
-        after an offer and an answer have been exchanged, it is
-        possible for the offerer to receive media after they have sent
-        an offer and before they have received an answer. To properly
+        after an offer and an answer have been exchanged, the offerer may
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -285,9 +285,10 @@ ipr="trust200902">
         contain the exact same number as the offer.</t>
 
         <t>Lastly, while the exact media parameters are only known only
-        after an offer and an answer have been exchanged, it is
-        possible for the offerer to receive media after they have sent
-        an offer and before they have received an answer. To properly
+        after an offer and an answer have been exchanged, the offerer may
+        receive ICE checks, and possibly media (e.g., in the case of a
+        re-offer after a connection has been established)
+        before they receive an answer. To properly
```
ekr: they -> it would, I think be clear.

juberti: SGTM

PR #711, opened 2017-05-06
Clarify attribute duplication for bundle. Fixes #655

### LINKED ISSUE #655: Clarify how identical/transport attributes are handled
Page 37 contains a statement that "attributes of category IDENTICAL or TRANSPORT should not be repeated in bundled m= sections". Two issues: this reads as normative, so consider "SHOULD NOT" rather than "should not." Also, I *think* this is true of bundle-only sections, not bundled sections in general.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1952,8 +1952,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           SDP attributes into different categories. To avoid
           unnecessary duplication when bundling, Section 8.1 of
           <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation" /> specifies
-          that attributes of category IDENTICAL or TRANSPORT should not
-          be repeated in bundled m= sections.</t>
+          that attributes of category IDENTICAL or TRANSPORT MUST NOT
+          be repeated in bundled m= sections. In initial offers, this applies
+          solely to m= sections which are marked as bundle-only, but
+          in subsequent offers it also applies to m= sections
```
juberti: I think this would be slightly clearer as

"... in subsequent offers it also applies to m= sections for which bundling has been negotiated, and is still desired."

You might be even able to get away with just saying

"This includes m= sections for which bundling has been negotiated and is still desired, as well as m= sections marked as bundle-only."

PR #714, opened 2017-05-07
Clarify normative language for scaling

### LINKED ISSUE #647: Clarify "original resolution"
Section 3.6.2 talks about "original resolution" where it appears to want to reference the current resolution. This happens in three different places.

### LINKED ISSUE #648: Confusion on normative language regarding scaling
BLOCKER: Section 3.6.2, second paragraph, second sentence says a sender MUST scale to match imageattr. The same section, paragraph seven, says it SHOULD scale to match imageattr. There needs to be only one normative statement on this topic, so as to avoid confusion among implementors.

### LINKED ISSUE #715: Clarify handling of upscaling vs dynamic resolution adjustment
To meet bandwidth limits, the encoder may dynamically apply downscaling. This may conflict with upscaling performed to meet a=imageattr requirements.

This conflict needs to be resolved somehow. I suggest that we remove the language around upscaling and policy and instead state that JSEP implementations SHOULD have appropriate fallback to ensure that they can decode 'too-low' resolutions in software.

### LINKED ISSUE #741: Simulcast interaction with a=imageattr
If a sender is transmitting video at 180p, 360p, and 720p, and it receives an a=imageattr that limits received resolution to 480p, what should happen?

a) 180p and 360p are transmitted intact, 720p is not transmitted
b) 180p and 360p are transmitted intact; 720p is scaled to 480p
c) Everything is scaled down; transmission is 480p, 240p, 120p

I would argue that a) makes the most sense here, but this is somewhat at odds with the guidance we are giving for the non-simulcast case, which is more like b).

juberti:
Fixes #741
Fixes #715
Fixes #648
Fixes #647

- Specifically consider simulcast
- Remove upscaling
- Clean up some contradictory text

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -747,19 +747,20 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           to a RtpSender, which is encoding the track video
           at the same or lower resolution(s) (the "encoder resolutions"),
           and a remote description is applied that references the sender and
-          contains valid "a=imageattr recv" attributes, it will
+          contains valid "a=imageattr recv" attributes, it MUST
           follow the rules below to ensure the sender does not transmit a
           resolution that would exceed the size criteria specified in the
-          attributes. These rules are followed
+          attributes. These rules MUST be followed
           as long as the attributes remain present in the remote description,
-          including if the track changes its resolution,
-          or if the track is replaced with a different track.
+          including cases in which the track changes its resolution,
+          or is replaced with a different track.
           </t>
 
           <t>Depending on how the RtpSender is configured, it may be producing
-          a single resolution, or, if simulcast <xref target="sec.simulcast" />
-          has been negotiated, multiple specific resolutions. In addition,
-          depending on the configuration, each encoding may have the
+          a single encoding at a certain resolution, or, if
+          simulcast <xref target="sec.simulcast" /> has been negotiated,
+          multiple encodings, each at their own specific resolution. In
+          addition, depending on the configuration, each encoding may have the
           flexibility to reduce resolution when needed, or may be locked to
           a specific output resolution.
           </t>
@@ -768,20 +769,21 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           rules are applied to determine what should be transmitted:
             <list style="symbols">
               <t>First, the right "a=imageattr recv" attribute is selected.
-              While JSEP endpoints will not include more than one such
-              attribute per media format, non-JSEP endpoints may send
-              multiple relevant "a=imageattr recv" attributes in a m=
-              section. The correct attribute is selected by taking the
+              This is performed by taking the
               attribute with the highest "q=" value from the set of attributes
               that reference the media format that has been selected for the
               specified encoding. If multiple attributes have the same "q="
               value, the one that appears first in the m= section is used.
+              Note that while JSEP endpoints will include at most
+              one "a=imageattr recv" attribute per media format, JSEP
+              endpoints may receive session descriptions from non-JSEP
+              endpoints with m= sections that contain multiple such attributes.
               </t>
 
               <t>If there is an applicable "a=imageattr recv" attribute for
               the encoding, the limits from the attribute are then compared to
               the encoder resolution. Only the specific limits mentioned below
-              are considered; any other values, e.g. picture aspect ratio,
+              are considered; any other values, such as picture aspect ratio,
               MUST be ignored.
               Note that when considering a MediaStreamTrack that is producing
               rotated video, the unrotated resolution MUST be used for the
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -768,8 +768,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <t>For each encoding being produced by the RtpSender, the following
           rules are applied to determine what should be transmitted:
             <list style="symbols">
-              <t>First, the right "a=imageattr recv" attribute is selected.
-              This is performed by taking the
+              <t>First, the most suitable "a=imageattr recv" attribute is
+              selected. This is performed by taking the
               attribute with the highest "q=" value from the set of attributes
               that reference the media format that has been selected for the
               specified encoding. If multiple attributes have the same "q="
@@ -801,7 +801,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>If the encoder resolution exceeds the maximum size permitted by the
               attribute, and the encoder is allowed to adjust its resolution,
               the encoder SHOULD apply downscaling in order to satisfy the limits,
-              although the downscaling MUST NOT change the track aspect ratio.
+              although the downscaling MUST NOT change the picture aspect ratio
+              of the encoding. For example, if the encoder resolution is
+              1280x720, and the attribute specified a maximum of 640x480, the
+              expected output resolution would be 640x360.
               If downscaling cannot be applied, the encoding MUST NOT be
               transmitted, and an error SHOULD be surfaced to the application.</t>
 
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -744,11 +744,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <t>This specification prescribes more specific behavior. When
           a sender of a given MediaStreamTrack, which is producing
           video of a certain resolution, receives an "a=imageattr recv"
-          attribute, it MUST check to see if the original resolution
-          meets the size criteria specified in the attribute, and adapt
-          the resolution accordingly by scaling (if appropriate). Note
-          that when considering a MediaStreamTrack that is producing
-          rotated video, the unrotated resolution MUST be used. This is
+          attribute, it will follow the rules below to check if the original
+          resolution meets the size criteria specified in the attribute,
+          and, if appropriate, adjust the resolution.
+          Note that when considering a MediaStreamTrack that is producing
+          rotated video, the unrotated resolution MUST be used for the
+          checks. This is
           required regardless of whether the receiver supports
           performing receive-side rotation (e.g., through CVO
           <xref target="TS26.114" />), as it significantly simplifies
@@ -779,15 +780,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           Downscaling MUST NOT change the track aspect ratio.</t>
 
           <t>If the original resolution is less than the size limits in
-          the attribute, upscaling is needed, but this may not be
-          appropriate in all cases. To address this concern, the
+          the attribute, upscaling is needed, but this may not always be
+          appropriate. To address this concern, the
           application can set an upscaling policy for each sent track.
-          For this case, if upscaling is permitted by policy, the
+          If upscaling is permitted by policy, the
           sender SHOULD apply upscaling in order to provide the desired
           resolution. Otherwise, the sender MUST NOT apply upscaling.
-          The sender SHOULD NOT upscale in other cases, even if the
-          policy permits it. Upscaling MUST NOT change the track aspect
-          ratio.</t>
+          Upscaling MUST NOT change the track aspect ratio.</t>
 
           <t>If there is no appropriate and permitted scaling mechanism
           that allows the received size limits to be satisfied, the
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -741,65 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           constrain the video formats that the sender can use, but
           gives an indication of the preferred values.</t>
 
-          <t>This specification prescribes more specific behavior. When
-          a sender of a given MediaStreamTrack, which is producing
-          video of a certain resolution, receives an "a=imageattr recv"
-          attribute, it will follow the rules below to check if the original
-          resolution meets the size criteria specified in the attribute,
-          and, if appropriate, adjust the resolution.
-          Note that when considering a MediaStreamTrack that is producing
-          rotated video, the unrotated resolution MUST be used for the
-          checks. This is
-          required regardless of whether the receiver supports
-          performing receive-side rotation (e.g., through CVO
-          <xref target="TS26.114" />), as it significantly simplifies
-          the matching logic.</t>
+          <t>This specification prescribes more specific behavior.
+          When a MediaStreamTrack, which is producing
+          video of a certain resolution (the "track resolution"), is attached
+          to a RtpSender, which is encoding the track video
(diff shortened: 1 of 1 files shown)


juberti:
I was trying to understand what kind of emoji :GT< was until I realized it was a flubbed LGTM :-)

juberti:
I am going to rework this PR to deal with what we discuss regarding upscaling today.

juberti:
Note to self: target #715, #647 with this

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-
-          <t>If the attribute includes a "sar=" (sample aspect ratio)
-          value set to something other than "1.0", indicating the
-          receiver wants to receive non-square pixels, this cannot be
-          satisfied and the sender MUST NOT transmit the track.</t>
-
-          <t>In the special case of receiving a maximum resolution of
-          [0, 0], as described above, the sender MUST NOT transmit the
-          track.</t>
+          <t>This specification prescribes more specific behavior.
```
ekr: colon instead of .?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          When a MediaStreamTrack, which is producing
+          video of a certain resolution (the "track resolution"), is attached
+          to a RtpSender, which is encoding the track video
```
ekr: I don't think you want a comma here, because this is restrictive.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          at the same or lower resolution(s) (the "encoder resolutions"),
+          and a remote description is applied that references the sender and
+          contains valid "a=imageattr recv" attributes, it will
```
ekr: perhaps s/will/MUST/

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          follow the rules below to ensure the sender does not transmit a
+          resolution that would exceed the size criteria specified in the
+          attributes. These rules are followed
```
ekr: s/are/MUST be/?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          as long as the attributes remain present in the remote description,
+          including if the track changes its resolution,
```
ekr: "including if" is odd. Perhaps "even if" or "including cases in which"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          or if the track is replaced with a different track.
+          </t>
+
+          <t>Depending on how the RtpSender is configured, it may be producing
+          a single resolution, or, if simulcast <xref target="sec.simulcast" />
+          has been negotiated, multiple specific resolutions. In addition,
+          depending on the configuration, each encoding may have the
+          flexibility to reduce resolution when needed, or may be locked to
+          a specific output resolution.
+          </t>
```
ekr: I would try to use consistent terminology here, which seems to be:

- The source produces media
- The RtpSender encodes it into encodings

I don't think either produces "resolutions".

So perhaps "a single encoding at a given resolution or multiple encodings, each at their own resolution"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+
+          <t>For each encoding being produced by the RtpSender, the following
+          rules are applied to determine what should be transmitted:
+            <list style="symbols">
+              <t>First, the right "a=imageattr recv" attribute is selected.
+              While JSEP endpoints will not include more than one such
+              attribute per media format, non-JSEP endpoints may send
+              multiple relevant "a=imageattr recv" attributes in a m=
+              section. The correct attribute is selected by taking the
```
ekr: I would move this point about JSEP to the end of this graf and say:

"Note that JSEP endpoints will only include one "a=imageattr recv" attribute, but JSEP endpoints may receive SDP from non-JSEP endpoints which include  more than one such attribute."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              attribute with the highest "q=" value from the set of attributes
+              that reference the media format that has been selected for the
+              specified encoding. If multiple attributes have the same "q="
+              value, the one that appears first in the m= section is used.
+              </t>
+
+              <t>If there is an applicable "a=imageattr recv" attribute for
+              the encoding, the limits from the attribute are then compared to
+              the encoder resolution. Only the specific limits mentioned below
+              are considered; any other values, e.g. picture aspect ratio,
```
ekr: , after e.g.

also I think I would use "such as" in this case.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              MUST be ignored.
+              Note that when considering a MediaStreamTrack that is producing
+              rotated video, the unrotated resolution MUST be used for the
+              checks. This is
+              required regardless of whether the receiver supports
+              performing receive-side rotation (e.g., through CVO
+              <xref target="TS26.114" />), as it significantly simplifies
+              the matching logic.</t>
```
ekr: I'm not that familiar with this, so Cullen should take a careful look.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              </t>
+
+              <t>If there is an applicable "a=imageattr recv" attribute for
+              the encoding, the limits from the attribute are then compared to
+              the encoder resolution. Only the specific limits mentioned below
+              are considered; any other values, e.g. picture aspect ratio,
+              MUST be ignored.
+              Note that when considering a MediaStreamTrack that is producing
+              rotated video, the unrotated resolution MUST be used for the
+              checks. This is
```
ekr: Maybe "checks, even if" rather than making a second sentence.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,80 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              value set to something other than "1.0", indicating the
+              receiver wants to receive non-square pixels, this cannot be
+              satisfied and the sender MUST NOT transmit the encoding.</t>
+
+              <t>If the encoder resolution exceeds the maximum size permitted by the
+              attribute, and the encoder is allowed to adjust its resolution,
+              the encoder SHOULD apply downscaling in order to satisfy the limits,
+              although the downscaling MUST NOT change the track aspect ratio.
+              If downscaling cannot be applied, the encoding MUST NOT be
+              transmitted, and an error SHOULD be surfaced to the application.</t>
```
ekr: Is this saying that I must pan-and-scan rather than doing pillarboxing or letterboxing?

fluffy:
I like the new text - I'd like to do a final read before we merge but so far looks good.

draft-ietf-rtcweb-jsep.xml
juberti: I intended it to be illustrative, but perhaps I misunderstand your comment.

draft-ietf-rtcweb-jsep.xml
juberti: I chose not to use a colon because the rest of this paragraph is in sentence rather than list form.

draft-ietf-rtcweb-jsep.xml
juberti: OOC, what is your rule for e.g. vs such as?

draft-ietf-rtcweb-jsep.xml
juberti: This is holdover from before, so there is no normative change.

draft-ietf-rtcweb-jsep.xml
juberti: It is saying that you must scale down uniformly (which may result in some form of boxing), rather than crop/pan+scan.

draft-ietf-rtcweb-jsep.xml
juberti: I tried that but the sentence ended up feeling run-on:

              Note that when considering a MediaStreamTrack that is producing
              rotated video, the unrotated resolution MUST be used for the
              checks, regardless of whether the receiver supports
              performing receive-side rotation (e.g., through CVO
              [TS26.114]), as it significantly simplifies
              the matching logic.

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml

draft-ietf-rtcweb-jsep.xml

juberti:
New rev uploaded with editorial fixes from Eric's review.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,82 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+              one "a=imageattr recv" attribute per media format, JSEP
+              endpoints may receive session descriptions from non-JSEP
+              endpoints with m= sections that contain multiple such attributes.
+              </t>
+
+              <t>If there is an applicable "a=imageattr recv" attribute for
+              the encoding, the limits from the attribute are then compared to
+              the encoder resolution. Only the specific limits mentioned below
+              are considered; any other values, such as picture aspect ratio,
+              MUST be ignored.
```
gloinul: This appears to be a new requirement "Only the specific limits mentioned below are considered; any other values, such as picture aspect ratio, MUST be ignored."  that forces all JSEP implementations to ignore parameters not explicitly mentioned here. I don't like that for several reasons. First, it prevents implementation that have a more complete imageattr interpreter and handler to select more appropriately when it is possible. 
If the goal here is to prioritize distorted video over no video, that appear fine, but the minimal to allow this is to use MAY, rather than MUST. Can we also be clear that this can be the effect of following this?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,82 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          multiple encodings, each at their own specific resolution. In
+          addition, depending on the configuration, each encoding may have the
+          flexibility to reduce resolution when needed, or may be locked to
+          a specific output resolution.
+          </t>
+
+          <t>For each encoding being produced by the RtpSender, the following
+          rules are applied to determine what should be transmitted:
+            <list style="symbols">
+              <t>First, the right "a=imageattr recv" attribute is selected.
```
gloinul: I think the use of "right" here is strange. I think this should be formulated more like: "First, a "a=imageattr recv" attirbute is selected as specified below."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,82 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+
+              <t>If the attribute includes a "sar=" (sample aspect ratio)
+              value set to something other than "1.0", indicating the
+              receiver wants to receive non-square pixels, this cannot be
+              satisfied and the sender MUST NOT transmit the encoding.</t>
+
+              <t>If the encoder resolution exceeds the maximum size permitted by the
+              attribute, and the encoder is allowed to adjust its resolution,
+              the encoder SHOULD apply downscaling in order to satisfy the limits,
+              although the downscaling MUST NOT change the track aspect ratio.
```
gloinul: Do you mean the picture aspect ration for this particular track? "track aspect ratio" is not clear on what is referred to.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -741,66 +741,82 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          </t>
+
+          <t>For each encoding being produced by the RtpSender, the following
+          rules are applied to determine what should be transmitted:
+            <list style="symbols">
+              <t>First, the right "a=imageattr recv" attribute is selected.
+              This is performed by taking the
+              attribute with the highest "q=" value from the set of attributes
+              that reference the media format that has been selected for the
+              specified encoding. If multiple attributes have the same "q="
```
gloinul: I think this selection algorithm has a big weakness, that it can select an attribute that results in no encoding at all. I would strongly propose that the selection criteria is to pick the alternative with the highest q that allows for an encoding, i.e. not blocked by the rules preventing an encoding to be generated by the below rules, such as SAR!=1.0.

draft-ietf-rtcweb-jsep.xml
bubbelman: To try to amend to that, I think it would be appropriate to avoid rejecting to use media based on "any other values" and in that sense they "MUST be ignored". If, on the other hand, (for example) several formats are listed that differ only in "any other values", it should be allowed for a receiver that understand them to take them into account when selecting format. "MAY be ignored" is therefore probably also not entirely what we want, because it would then still be OK for someone that does not ignore them to reject a format based on "any other values". Would something like "...picture aspect ratio, MUST be ignored when comparing for match with encoder capabilities, but MAY be used to discriminate between otherwise comparable formats" be clear enough?

juberti: This was previously agreed at [IETF 95](https://docs.google.com/presentation/d/1pP4LDQQSyVDUTa20R3Zgmb9PXZJPPcXiSzFOYw64dKc/edit#slide=id.g1274a0be3c_0_123) , so if we are going to make any changes, it should be in a new bug.

draft-ietf-rtcweb-jsep.xml
juberti: What should happen then in the case of only a SAR!=1.0 imageattr?

draft-ietf-rtcweb-jsep.xml
juberti: Fixed.

draft-ietf-rtcweb-jsep.xml
gloinul: @juberti those slides does not indicate that it was inteded to be required to ignore the rest of the attributes. I think there is a big difference between mandating ignoring, and allowing for ignoring. Are there issues you see with an implementation actually taken the parameters into account?

PR #716, opened 2017-05-11
Make clear you don't have to aggregate. Fixes #635

### LINKED ISSUE #635: Resolve attribute aggregation
BLOCKER: Several of the SDP-building procedures in this document call out addition of "a=fingerprint" to individual media sections, while the canonical definition of "a=fingerprint" allows these to be added at either the session level or the media level. The same is true of ice-pwd and ice-ufrag. While the process of SDP building in this document appears to be prescriptive on purpose, one assumes that it is allowed some variation so as to accommodate deployment of future SDP extensions. Because of this state of affairs, it is unclear whether implementations that consolidate these attributes at the session level when they would otherwise be identical among all media sections is such an allowable variation. Ambiguity in specifications is never a good thing, so I believe that the draft needs to do one of: (a) specifically call out this kind of aggregation as allowed; (b) specifically call out this kind of aggregation as disallowed; or (c) include a blanket statement that implementations are allowed to emit SDP that is semantically equivalent to that which is generated by the procedures in this document, but which may vary in exact structure. I have a strong preference for (c) -- implementations will be required to deal with such variations anyway to interop with legacy SDP, so prohibiting JSEP clients from generating them serves no goal.

ekr:
I did add a MUST here, but I think it's just a restatement of what is already true.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2205,9 +2205,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           and debugging by making it easier to understand individual
           media sections, especially if one of a set of initially identical
           attributes is subsequently changed. However, implementations MAY
-          choose to aggregate attributes at the media level and JSEP
+          choose to aggregate attributes at the session level and JSEP
           implementations MUST be prepared to receive attributes
-          in both locations.</t>
+          in either location.</t>
 
           <t>Attributes other than the ones specified above MAY be
           included, except for the following attributes which are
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2204,7 +2204,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           level even if they are identical. This assists development
           and debugging by making it easier to understand individual
           media sections, especially if one of a set of initially identical
-          attributes is subsequently changed.</t>
+          attributes is subsequently changed. However, implementations MAY
+          choose to aggregate attributes at the media level and JSEP
```
juberti: should this say "session level"?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2204,7 +2204,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          implementations MUST be prepared to receive attributes
+          in both locations.</t>
```
juberti: They won't be in both simultaneously. Perhaps, 'in either location'?

PR #717, opened 2017-05-11
Explain "recycled" before its use in Section 3.5.1

### LINKED ISSUE #643: Add forward ref for m= line recycling
Section 3.5.1 talks about a "recycled m= section" without any context for what this means. Please provide a forward reference to the section that discusses m= section recycling.

juberti:
Fixes #643

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -453,10 +453,10 @@ setRemote(ANSWER) |                                   |
           <t>An RtpTransceiver is never associated with more than one m=
           section, and once a session description is applied, a m=
           section is always associated with exactly one
-          RtpTransceiver. However, once a RtpTransceiver is stopped,
-          and/or its associated m= section is rejected, that m= section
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -448,11 +448,15 @@ setRemote(ANSWER) |                                   |
...
-          RtpTransceiver is never associated with more than one m=
+          when creating an offer or applying a remote offer.</t>
+
+          <t>An RtpTransceiver is never associated with more than one m=
           section, and once a session description is applied, a m=
           section is always associated with exactly one
-          RtpTransceiver.</t>
+          RtpTransceiver. However, once a RtpTransceiver is stopped,
+          and/or its associated m= section is rejected, that m= section
+          may be "recycled" and associated with a new RtpTransceiver with
```
ekr: This "may" is making me a little sad, because we are very strict about what you can and cannot do.

Perhaps "will, under certain conditions" and then a cite to the relevant sections?

juberti:
PTAL

PR #718, opened 2017-05-11
Provide a reference for glare. Fixes #638

### LINKED ISSUE #638: Define "glare"
Section 1.2 uses the term "glare" without describing it. This term may be unfamiliar to readers, and should be explained or cited.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -174,7 +174,8 @@ ipr="trust200902">
         consume messages from this protocol. While providing a more
         high-level API, this put more control of signaling within the
         JSEP implementation, forcing it to have to understand and
-        handle concepts like signaling glare. In addition, it prevented
+        handle concepts like signaling glare (see
+        <xref target="RFC3264"/>, Section 4). In addition, it prevented
         the application from driving the state machine to a desired
         state, as is needed in the page reload case.</t>
```
juberti: I wonder if we should remove this last sentence, since we're no longer trying to support rehydration.

ekr: Yeah probably right. I can do that.

ekr:
@juberti: done. @fluffy, still waiting for you.

PR #733, opened 2017-06-01
rephrase ice-mismatch handling

### LINKED ISSUE #636: RFC 5245 contradiction
BLOCKER: The text in the third paragraph of section 5.7.1 intentionally contradicts language in RFC 5245; as such, this document needs to formally update that RFC. Note that updated RFCs are expected to be mentioned in the Abstract as well as the document metadata.

fluffy:
Fixes #636

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3186,9 +3186,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <list>
 
             <t>The "c=" line MUST be checked for syntax but its value
-            is only used for ICE mismatch detection. Some
-            implementations do not bother to implement ICE mismatch
-            detection as ICE is required in WebRTC.</t>
+            is only used for ICE mismatch detection, as defined in
+            <xref target="RFC5245" />, Section 6.1. Note that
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3186,11 +3186,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <list>
 
             <t>The "c=" line MUST be checked for syntax but its value
-            is not used. This supersedes the guidance in
-            <xref target="RFC5245" />, Section 6.1, to use
-            "ice-mismatch" to indicate mismatches between "c=" and the
-            candidate lines; because JSEP always uses ICE,
-            "ice-mismatch" is not useful in this context.</t>
+            is only used for ICE mismatch detection. Some
+            implementations do not bother to implement ICE mismatch
```
ekr: I would rewrite the second sentence as:

"Note that JSEP implementations should never encounter this condition because ICE is required for WebRTC."

juberti: Agreed. For the first sentence, I would suggest leaving the reference to 5245, Section 6.1.

ekr:
@juberti, @fluffy: PTAL

PR #736, opened 2017-06-02
Fixes #730. Remove the requirement that fingerprints match the

### LINKED ISSUE #730: Reference to fingerprint digest algorithm matching the certificate should be removed
Section 5.2.1 (creating an offer) still has this paragraph:

>      An "a=fingerprint" line for each of the endpoint's certificates,
>      as specified in [RFC4572], Section 5; the digest algorithm used
>      for the fingerprint MUST match that used in the certificate
>      signature.

In RFC4572-update, this requirement doesn't exist. So should this be changed to reference RFC4572-update, removing the second half of this paragraph?

ekr:
certificate.

This was updated in RFC 8122.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2117,7 +2117,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <xref target="RFC5245"></xref>, Section 15.4.</t>
 
               <t>For each desired digest algorithm, one or more
-              "a=fingerprint" line for each of the endpoint's
+              "a=fingerprint" lines for each of the endpoint's
               certificates, as specified in
               <xref target="RFC8122"></xref>, Section 5.</t>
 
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2118,9 +2118,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
               <t>An "a=fingerprint" line for each of the endpoint's
               certificates, as specified in
-              <xref target="RFC8122"></xref>, Section 5; the digest
-              algorithm used for the fingerprint MUST match that used
-              in the certificate signature.</t>
+              <xref target="RFC8122"></xref>, Section 5.</t>
```
juberti: Since we may need to support multiple hash algos for hash agility, we should make it clear that we could end up with more a=fingerprint lines than certs, i.e. we might need some sort of "for each desired digest algorithm".

ekr:
@juberti: PTAL

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2116,11 +2116,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>"a=ice-ufrag" and "a=ice-pwd" lines, as specified in
               <xref target="RFC5245"></xref>, Section 15.4.</t>
 
-              <t>An "a=fingerprint" line for each of the endpoint's
+              <t>For each desired digest algorithm, one or more
+              "a=fingerprint" line for each of the endpoint's
```
juberti: line -> lines

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2847,9 +2846,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
               <t>An "a=fingerprint" line for each of the endpoint's
```
juberti: Does this need the same treatment as above?

ekr:
@fluffy: PTAL

PR #739, opened 2017-06-09
update list RTP profiles to match IANA table

### LINKED ISSUE #738: Get the list of (UDP/TCP/...[s]AVP[f] to be match what IANA allows

fluffy:
Fixes #738

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1800,12 +1800,6 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                 <t> "RTP/SAVPF" </t>
                 <t> "TCP/DTLS/RTP/SAVP" </t>
                 <t> "TCP/DTLS/RTP/SAVPF" </t>
-                <t> "TCP/RTP/AVP" </t>
-                <t> "TCP/RTP/AVPF" </t>
-                <t> "TCP/RTP/SAVP" </t>
-                <t> "TCP/RTP/SAVPF" </t>
-                <t> "TCP/TLS/RTP/AVP" </t>
-                <t> "TCP/TLS/RTP/AVPF" </t>
                 <t> "UDP/TLS/RTP/SAVP" </t>
                 <t> "UDP/TLS/RTP/SAVPF" </t>
               </list>
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1794,14 +1794,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>Any profile in the offer matching one of the following
               MUST be accepted:
               <list style="symbols">
-                <t> "RTP/AVP" </t>
-                <t> "RTP/AVPF" </t>
-                <t> "RTP/SAVP" </t>
-                <t> "RTP/SAVPF" </t>
-                <t> "TCP/DTLS/RTP/SAVP" </t>
-                <t> "TCP/DTLS/RTP/SAVPF" </t>
-                <t> "UDP/TLS/RTP/SAVP" </t>
-                <t> "UDP/TLS/RTP/SAVPF" </t>
+                <t> "RTP/AVP" (Defined in <xref target="RFC4566"></xref>, Section 8.2.2)</t>
+                <t> "RTP/AVPF" (Defined in <xref target="RFC4585"></xref>, Section 9)</t>
+                <t> "RTP/SAVP" (Defined in <xref target="RFC3711"></xref>, Section 12)</t>
+                <t> "RTP/SAVPF" (Defined in <xref target="RFC5124"></xref>, Section 6)</t>
+                <t> "TCP/DTLS/RTP/SAVP" (Defined in <xref target="RFC7850"></xref>, Section 3.4)</t>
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1798,12 +1798,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-                <t> "UDP/TLS/RTP/AVPF" </t>
+                <t> "TCP/DTLS/RTP/SAVP" </t>
+                <t> "TCP/DTLS/RTP/SAVPF" </t>
+                <t> "TCP/RTP/AVP" </t>
+                <t> "TCP/RTP/AVPF" </t>
+                <t> "TCP/RTP/SAVP" </t>
+                <t> "TCP/RTP/SAVPF" </t>
+                <t> "TCP/TLS/RTP/AVP" </t>
+                <t> "TCP/TLS/RTP/AVPF" </t>
                 <t> "UDP/TLS/RTP/SAVP" </t>
```
juberti: This seems a bit inconsistent. Why do we have TCP/TLS/RTP/AVP, but not UDP/TLS/RTP/AVP?

fluffy: The only draft or RFC that uses "UDP/TLS/RTP/AVP" is one typo in https://tools.ietf.org/html/rfc5939. I just reported an RFC Errata on that. I think the argument here is anyone who was advertising TLS in the label, also put the S in the SAVP so we don't need to worry about any of the cases where we have TLS. 

The case with TCP it is a bit different. You can take RTP (not SRTP) and send it as unencrypted RTP over a normal TLS (not DTLS-SRTP) tunnel so for that case the label is TCP/TLS/RTP/AVPF. However, this is RTP over 4571 framing so treating it as  SRTP is going to not work. Because of this, I think we need to remove all the options from this list that are RTP (not SRTP) framed over 4571. That means remove from the list "TCP/RTP/AVP", "TCP/RTP/AVPF" ,  "TCP/TLS/RTP/AVP" , "TCP/TLS/RTP/AVPF".

For "TCP/RTP/SAVP" and "TCP/RTP/SAVPF", this is SRTP with 4571 framing over TCP ( See https://tools.ietf.org/html/rfc7850#section-3.2 ). Since this is not DTLS-SRTP, it seems we should also remove theses two labels as well. 

So what I am proposing is we move the list to be just 

               <t> "RTP/AVP" </t>
                <t> "RTP/AVPF" </t>
                <t> "RTP/SAVP" </t>
                <t> "RTP/SAVPF" </t>
                <t> "TCP/DTLS/RTP/SAVP" </t>
                <t> "TCP/DTLS/RTP/SAVPF" </t>
                <t> "UDP/TLS/RTP/SAVP" </t>
                <t> "UDP/TLS/RTP/SAVPF" </t>

Backing up to first principals on this, the problem we are tying to solve is an offer that is offering RTP or DTLS-SRTP. It is going to have RTP/AVP[F] and a fingerprint so it seems as long as we have that on the list it is good. 

Of some one, RTP/SAVP is defined in https://tools.ietf.org/html/rfc3711#section-12 (and the RTP/SAVFP version in https://tools.ietf.org/html/rfc5124#section-6) while the UDP/TLS/RTP/SAVP[F] is in https://tools.ietf.org/html/rfc5764#section-9 )

fluffy:
The only draft or RFC that uses "UDP/TLS/RTP/AVP" is one typo in https://tools.ietf.org/html/rfc5939. I just reported an RFC Errata on that. I think the argument here is anyone who was advertising TLS in the label, also put the S in the SAVP so we don't need to worry about any of the cases where we have TLS.

The case with TCP it is a bit different. You can take RTP (not SRTP) and send it as unencrypted RTP over a normal TLS (not DTLS-SRTP) tunnel so for that case the label is TCP/TLS/RTP/AVPF. However, this is RTP over 4571 framing so treating it as SRTP is going to not work. Because of this, I think we need to remove all the options from this list that are RTP (not SRTP) framed over 4571. That means remove from the list "TCP/RTP/AVP", "TCP/RTP/AVPF" , "TCP/TLS/RTP/AVP" , "TCP/TLS/RTP/AVPF".

For "TCP/RTP/SAVP" and "TCP/RTP/SAVPF", this is SRTP with 4571 framing over TCP ( See https://tools.ietf.org/html/rfc7850#section-3.2 ). Since this is not DTLS-SRTP, it seems we should also remove theses two labels as well.

So what I am proposing is we move the list to be just
           <t> "RTP/AVP" </t>
            <t> "RTP/AVPF" </t>
            <t> "RTP/SAVP" </t>
            <t> "RTP/SAVPF" </t>
            <t> "TCP/DTLS/RTP/SAVP" </t>
            <t> "TCP/DTLS/RTP/SAVPF" </t>
            <t> "UDP/TLS/RTP/SAVP" </t>
            <t> "UDP/TLS/RTP/SAVPF" </t>

Backing up to first principals on this, the problem we are tying to solve is an offer that is offering RTP or DTLS-SRTP. It is going to have RTP/AVP[F] and a fingerprint so it seems as long as we have that on the list it is good.

Of some one, RTP/SAVP is defined in https://tools.ietf.org/html/rfc3711#section-12(and the RTP/SAVFP version in https://tools.ietf.org/html/rfc5124#section-6) while the UDP/TLS/RTP/SAVP[F] is in https://tools.ietf.org/html/rfc5764#section-9 )

fluffy:
I added the refs for where these are defined. If that look OK, I think this is ready to merge.

PR #744, opened 2017-06-21
cleaned up language around section refs

### LINKED ISSUE #743: For consistency, prefer '[REF], Section X' instead of 'Section X of [REF]'
Worth taking an editorial pass over this if time permits

fluffy:
Fixes #743

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -943,8 +943,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           call, and multiple simultaneous active signaling sessions can
           be established as a result. If multiple callees send media at
           the same time, the possibilities for handling this are
-          described in Section 3.1 of
-          <xref target="RFC3960"></xref>. Most SIP devices today only
+          described in
+          <xref target="RFC3960"></xref>, Section 3.1.
+          Most SIP devices today only
           support exchanging media with a single device at a time, and
           do not try to mix multiple early media audio sources, as that
           could result in a confusing situation. For example, consider
@@ -1907,7 +1908,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           </list></t>
 
           <t>The next step is to generate m= sections, as specified in
-          <xref target="RFC4566" /> Section 5.14. An m= section is
+          <xref target="RFC4566" />, Section 5.14. An m= section is
           generated for each RtpTransceiver that has been added to the
           PeerConnection, excluding any stopped RtpTransceivers. This
           is done in the order the RtpTransceivers were added to the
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1977,8 +1978,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <t>
           <xref target="I-D.ietf-mmusic-sdp-mux-attributes" /> groups
           SDP attributes into different categories. To avoid
-          unnecessary duplication when bundling, Section 8.1 of
-          <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation" /> specifies
+          unnecessary duplication when bundling
+          <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation" />,
```
juberti: This one might be better as-is.

juberti: or, change it to 

> "attributes of category IDENTICAL or TRANSPORT MUST NOT be repeated in bundled m= sections, as specified in [ref], Section 8.1."

juberti: @ekr, I think you wrote this part, interested in your thoughts.

fluffy:
I like the text in your comment and changed to that.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1976,11 +1977,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          unnecessary duplication when bundling, Section 8.1 of
-          <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation" /> specifies
-          that attributes of category IDENTICAL or TRANSPORT MUST NOT
-          be repeated in bundled m= sections. This includes m=
+          SDP attributes into different categories.
+          To avoid unnecessary duplication when bundling, attributes
+          of category IDENTICAL or TRANSPORT MUST NOT be repeated in
+          bundled m= sections, as specified in <xref
+          target="I-D.ietf-mmusic-sdp-bundle-negotiation" />, Section
+          8.1.
```
ekr: This one looks a bit odd because you might read it as contradicting BUNDLE. I.e., it's not-repeating which is specified, not the repeating that we are now prohibiting.

Perhaps "This restates the requirements of...."

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3965,7 +3968,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <xref target="RFC4585" />, Section 4.2. When sending RTCP
             feedback, follow the rules and recommendations from
             <xref target="RFC8108"></xref>,
-            Section 5.4.1 to select which SSRC to use.</t>
+            Section 5.4.1, to select which SSRC to use.</t>
```
ekr: I think this second comma should probably go.

juberti: agree

draft-ietf-rtcweb-jsep.xml
juberti: how about 'repeating the guidance from [REF]'

PR #746, opened 2017-06-23
add pointers for rollback to apply local/remote sections

### LINKED ISSUE #650: Rollback missing from type enum
Section 4.1.9 paragraph one includes an enumeration of things that can appear in the "type" field, but omits "rollback" from the list.

fluffy:
Fixes #650

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1433,7 +1433,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           PeerConnection to apply the supplied session description as
           its local configuration. The type field indicates whether the
           description should be processed as an offer, provisional
-          answer, or final answer; offers and answers are checked
+          answer, rollback, or final answer; offers and answers are checked
           differently, using the various rules that exist for each SDP
           line.</t>
 
@@ -1477,7 +1477,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>This API changes the local media state; among other
           things, it sets up local resources for sending and encoding
-          media.</t>
+          media.
+          A rollback will revert to the current local description.</t>
 
           <t>If setLocalDescription was previously called with an
           offer, and setRemoteDescription is called with an answer
@@ -3565,6 +3566,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         session MUST be restored to the state it was in before
         performing these steps.</t>
 
+        <t>If the description is of type "rollback", follow the
+        processing defined in the <xref target="sec.rollback" /> and
+        skip the processing described in the rest of this
+        section.</t>
+
         <t>Next, m= sections are processed. For each m= section, the
         following steps MUST be performed; if any parameters are out of
         bounds, or cannot be applied, processing MUST stop and an error
@@ -3657,6 +3663,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         description. If an error is returned, the session MUST be
         restored to the state it was in before performing these
         steps.</t>
+
+        <t>If the description is of type "rollback", follow the
+        processing defined in the <xref target="sec.rollback" /> and
+        skip the processing described in the rest of this
+        section.</t>
 
         <t>If the answer contains any "a=ice-options" attributes where
         "trickle" is listed as an attribute, update the PeerConnection
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1696,8 +1696,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <t>When creating offers, the transceiver direction is
           directly reflected in the output, even for re-offers. When
           creating answers, the transceiver direction is intersected
-          with the offered direction, as explained in the
-          <xref target="sec.generating-an-answer" /> section below.</t>
+          with the offered direction, as explained in
+          <xref target="sec.generating-an-answer" /> below.</t>
 
           <t>Note that while setDirection sets the direction property
           of the transceiver immediately
@@ -3160,12 +3160,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           MUST stop and an error MUST be returned.</t>
 
           <t>Next, the SessionDescription is parsed into a data
-          structure, as described in the <xref
-          target="sec.parsing-a-desc" /> section below. </t>
+          structure, as described in <xref
+          target="sec.parsing-a-desc" /> below. </t>
 
           <t>Finally, the parsed SessionDescription is applied as
-          described in the
-          <xref target="sec.applying-a-local-desc" /> section below.</t>
+          described in
+          <xref target="sec.applying-a-local-desc" /> below.</t>
         </list></t>
       </section>
       <section title="Processing a Remote Description"
@@ -3196,14 +3196,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           processing MUST stop and an error MUST be returned.</t>
 
           <t>Next, the SessionDescription is parsed into a data
-          structure, as described in the
-          <xref target="sec.parsing-a-desc" /> section below. If parsing
+          structure, as described in
+          <xref target="sec.parsing-a-desc" /> below. If parsing
           fails for any reason, processing MUST stop and an error MUST
           be returned.</t>
 
           <t>Finally, the parsed SessionDescription is applied as
-          described in the
-          <xref target="sec.applying-a-remote-desc" /> section
+          described in
+          <xref target="sec.applying-a-remote-desc" />
           below.</t>
         </list></t>
       </section>
@@ -3400,8 +3400,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           </list></t>
 
           <t>If the "m=" proto value indicates use of RTP, as described
-          in the
-          <xref target="sec.profile-names" /> section above, the
+          in
+          <xref target="sec.profile-names" /> above, the
           following attribute lines MUST be processed:
           <list style="symbols">
 
@@ -3669,8 +3669,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         </list></t>
 
         <t>Finally, if this description is of type "pranswer" or
-        "answer", follow the processing defined in the
-        <xref target="sec.applying-an-answer" /> section below.</t>
+        "answer", follow the processing defined in
+        <xref target="sec.applying-an-answer" /> below.</t>
       </section>
       <section title="Applying a Remote Description"
       anchor="sec.applying-a-remote-desc">
@@ -3896,8 +3896,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         </list></t>
 
         <t>Finally, if this description is of type "pranswer" or
-        "answer", follow the processing defined in the
-        <xref target="sec.applying-an-answer" /> section below.</t>
+        "answer", follow the processing defined in
+        <xref target="sec.applying-an-answer" /> below.</t>
       </section>
       <section title="Applying an Answer"
       anchor="sec.applying-an-answer">
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -399,8 +399,8 @@ setRemote(ANSWER) |                                   |
         operation, described in <xref target="sec.rollback"/>, that is
         can take the system back to the last stable state. It is
         typically used to recovering from errors that happened on the
-        remote side. </t>
-        
+        remote side.</t>
+
         <t>Aside from these state transitions there is no other
         difference between the handling of provisional ("pranswer") and
         final ("answer") answers.</t>
@@ -3136,7 +3136,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           processing defined in <xref target="sec.rollback" /> and
           skip the processing described in the rest of this
           section.</t>
-           
+
           <t>The type of the SessionDescription is checked
           against the current state of the PeerConnection:
           <list style="symbols">
@@ -3179,7 +3179,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           processing defined in <xref target="sec.rollback" /> and
           skip the processing described in the rest of this
           section.</t>
-              
+
           <t>The type of the SessionDescription is checked
           against the current state of the PeerConnection:
           <list style="symbols">
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -257,7 +257,7 @@ ipr="trust200902">
         </figure>
       </section>
       <section title="Session Descriptions and State Machine"
-      anchor="sec.session-descriptions-and-">
+      anchor="sec.session-descriptions-and-state-machine">
 
         <t>In order to establish the media plane, the JSEP implementation needs
         specific parameters to indicate what to transmit to the remote
@@ -395,6 +395,12 @@ setRemote(ANSWER) |                                   |
           </figure>
         </t>
 
+        <t> In addition to theses state transitions, there is a rollback
+        operation, described in <xref target="sec.rollback"/>, that is
+        can take the system back to the last stable state. It is
+        typically used to recovering from errors that happened on the
+        remote side. </t>
+        
         <t>Aside from these state transitions there is no other
         difference between the handling of provisional ("pranswer") and
         final ("answer") answers.</t>
@@ -1378,10 +1384,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             setLocalDescription. However, the remote side, either
             before or after setRemoteDescription, decides it does not
             want to accept the new parameters, and sends a reject
-            message back to the offerer. Now, the offerer, and possibly
-            the answerer as well, need to return to a stable state and
-            the previous local/remote description. To support this, we
-            introduce the concept of "rollback".</t>
+            message back to the offerer.
+            Now, the offerer, and possibly the answerer as well, need
+            to return to the state machine described in <xref
+            target="sec.session-descriptions-and-state-machine"/> to
+            the "Stable" state and the previous local/remote
+            description. To support this, we introduce the concept of
+            "rollback".</t>
 
             <t>A rollback discards any proposed changes to the session,
             returning the state machine to the stable state, and
@@ -1433,7 +1442,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           PeerConnection to apply the supplied session description as
           its local configuration. The type field indicates whether the
           description should be processed as an offer, provisional
-          answer, rollback, or final answer; offers and answers are checked
+          answer, final answer, or rollback; offers and answers are checked
           differently, using the various rules that exist for each SDP
           line.</t>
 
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1433,7 +1433,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           PeerConnection to apply the supplied session description as
           its local configuration. The type field indicates whether the
           description should be processed as an offer, provisional
-          answer, or final answer; offers and answers are checked
+          answer, rollback, or final answer; offers and answers are checked
```
juberti: I think I would put 'rollback' at the end, and then say 'the SDP for offers and answers is checked differently, using ... that exist for each line.'

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1433,7 +1433,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           differently, using the various rules that exist for each SDP
           line.</t>
```
ekr: Odd lack of parallel structure here between offer/answer and rollback

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1477,7 +1477,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>This API changes the local media state; among other
           things, it sets up local resources for sending and encoding
-          media.</t>
+          media.
+          A rollback will revert to the current local description.</t>
```
ekr: From the pending one, right?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3565,6 +3566,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         session MUST be restored to the state it was in before
         performing these steps.</t>
 
+        <t>If the description is of type "rollback", follow the
+        processing defined in the <xref target="sec.rollback" /> and
```
juberti: I think you want to remove 'the'.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3658,6 +3664,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         restored to the state it was in before performing these
         steps.</t>
 
+        <t>If the description is of type "rollback", follow the
```
juberti: See above (both comments).

draft-ietf-rtcweb-jsep.xml
juberti: 'Current remote description'. But not sure this is adding a lot here, may be simpler to just remove.

draft-ietf-rtcweb-jsep.xml
juberti: Also, I think this should be hoisted to the 'Processing a Local Description' section, since you will also need to discuss the states where this is permitted (and the fact that the desc doesn't need to be parsed).

draft-ietf-rtcweb-jsep.xml
```diff
@@ -395,6 +395,12 @@ setRemote(ANSWER) |                                   |
           </figure>
         </t>
 
+        <t> In addition to theses state transitions, there is a rollback
+        operation, described in <xref target="sec.rollback"/>, that is
+        can take the system back to the last stable state. It is
```
ekr: "that returns the system to the stable state. Rollback is"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -395,6 +395,12 @@ setRemote(ANSWER) |                                   |
...
+        typically used to recovering from errors that happened on the
```
ekr: "to recover from errors produced by"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1378,10 +1384,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-            the answerer as well, need to return to a stable state and
-            the previous local/remote description. To support this, we
-            introduce the concept of "rollback".</t>
+            message back to the offerer.
+            Now, the offerer, and possibly the answerer as well, need
+            to return to the state machine described in <xref
+            target="sec.session-descriptions-and-state-machine"/> to
+            the "Stable" state and the previous local/remote
+            description. To support this, we introduce the concept of
+            "rollback".</t>
```
ekr: I actually wouldn't make this change. I think the previous text was clearer.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1449,7 +1458,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           and continues until setRemoteDescription is called with
           either a final answer, at which point the PeerConnection can
           fully adopt the pending local description, or a rollback,
-          which results in a revert to the current local
+          which results in a revert back from the pending local
+          description to the current local
```
ekr: This seems like awkward language. I think I might say
"which causes the JSEP implementation to revert back from the pending local description to the current locatl description"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1477,7 +1487,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>This API changes the local media state; among other
           things, it sets up local resources for sending and encoding
-          media.</t>
+          media.
+          A rollback will revert from the pending remote description
+          to the current remote description.</t>
```
ekr: See above.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -395,6 +395,12 @@ setRemote(ANSWER) |                                   |
           </figure>
         </t>
 
+        <t> In addition to theses state transitions, there is a
```
juberti: theses -> these

draft-ietf-rtcweb-jsep.xml
```diff
@@ -395,6 +395,12 @@ setRemote(ANSWER) |                                   |
...
+        rollback operation, described in <xref
+        target="sec.rollback"/>, that returns the system to the stable
+        state. Rollback is typically used to recover from errors
+        produced by the remote side.</t>
```
juberti: This isn't always true though. Maybe, 'typically used to recover when an offer is rejected'?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1378,10 +1384,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             setLocalDescription. However, the remote side, either
             before or after setRemoteDescription, decides it does not
             want to accept the new parameters, and sends a reject
-            message back to the offerer. Now, the offerer, and possibly
-            the answerer as well, need to return to a stable state and
-            the previous local/remote description. To support this, we
-            introduce the concept of "rollback".</t>
+            message back to the offerer.
```
juberti: Is there any change to this section?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1449,7 +1457,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           and continues until setRemoteDescription is called with
           either a final answer, at which point the PeerConnection can
           fully adopt the pending local description, or a rollback,
-          which results in a revert to the current local
+          which causes the JSEP implementation to revert back from the
```
juberti: This seems less clear than before.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1477,7 +1486,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>This API changes the local media state; among other
           things, it sets up local resources for sending and encoding
-          media.</t>
+          media.
+          A rollback will causes the JSEP implementation to revert
```
juberti: Not sure this sentence is needed; it's covered in more detail elsewhere.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3120,7 +3132,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         setLocalDescription, the following steps MUST be performed:
         <list style="symbols">
 
-          <t>First, the type of the SessionDescription is checked
+          <t>If the description is of type "rollback", follow the
+          processing defined in <xref target="sec.rollback" /> and
+          skip the processing described in the rest of this
+          section.</t>
+
+          <t>The type of the SessionDescription is checked
```
juberti: cheked -> checked

juberti: I think we probably also need to document in this section the PC states in which a setLD(rollback) is valid. It's not totally clear from the rollback section.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3158,7 +3175,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         setRemoteDescription, the following steps MUST be performed:
         <list style="symbols">
 
-          <t>First, the type of the SessionDescription is checked
+          <t>If the description is of type "rollback", follow the
```
juberti: I think we probably also need to document in this section the PC states in which a setRD(rollback) is valid. It's not totally clear from the rollback section.

draft-ietf-rtcweb-jsep.xml
fluffy: no - change - it was changed , but EKR thought orig was better so I changed it back. Next rer will try and line up formatting to make it clear there is not change.

fluffy:
So trying to outline some thoughts on what we write in the new empty section title "Processing a Rollback".

 
        if state is stable, then do nothing

        there are 4 other possible states

        have-local-offer
        have-remote-pranswer
        have-remote-offer
        have-local-pranswer

        and the rollback can arrive via a setLocal or setRemote

        The common case where an offer was sent, a pranswer might be
        received, then an error received: This is going to be the state
        is have-local-offer or have-remote-pranswer and a setRemote
        gets the rollback.

        The case where an offer was received, a pranswer might have
        been sent, then error occurred:  State is  have-remote-offer or
        have-local-pranswer and a setLocal gets the rollback.

        The other 4 cases are a bit harder to image where they happen
        but it seems reasonable that what happens is exactly the same
        as the above two case situations.

        I'm not sure what to say about what happens but it seems to be
        the same in all cases. The handling of API objects is
        described in <xref target="sec.rollback"/>. The handling of the
        SDP is that the things the media stack simply reverts to state
        it was on the last Stable state.

        The only really weird case I think we should ignore is A sends
        offer to B and A is in in have-local-offer state, B sends
        answer and now B is in stable state. But something goes
        sideways and the web server sends a rollback to A and B that
        overlaps with the answer coming from B to the web server. At
        this point, A and B, roll back to different states.  For this
        to work, the website should not be doing third party rollback -
        just don't do it.

draft-ietf-rtcweb-jsep.xml
fluffy: I did not fix this one as EKR was looking for this but I can go eitherway

draft-ietf-rtcweb-jsep.xml
fluffy: I did not fix this one as EKR was looking for this but I can go eitherway

juberti:
Not sure I followed your comments:

> The common case where an offer was sent, a pranswer might be
received, then an error received: This is going to be the state
is have-local-offer or have-remote-pranswer and a setRemote
gets the rollback.

I think you mean, it is have-local-offer, but setLocal gets the rollback.

> The case where an offer was received, a pranswer might have
been sent, then error occurred:  State is  have-remote-offer or
have-local-pranswer and a setLocal gets the rollback.

Again, I think you mean, it is have-remote-offer, but setRemote gets the rollback. 

Handling rollback of pranswer is more complicated, of course (which method does it get supplied to?) I think such rollback situations are true corner cases and we should avoid them entirely.

> The other 4 cases are a bit harder to image where they happen but it seems reasonable that what happens is exactly the same as the above two case situations.

Not quite sure what this means. Do you mean have-remote-offer, but setLocal gets the rollback?

>     The only really weird case I think we should ignore is A sends
>     offer to B and A is in in have-local-offer state, B sends
>     answer and now B is in stable state. But something goes
>     sideways and the web server sends a rollback to A and B that
>     overlaps with the answer coming from B to the web server. At
>     this point, A and B, roll back to different states.  For this
>     to work, the website should not be doing third party rollback -
>     just don't do it.

B won't be able to rollback since B already applied their answer, so this should result in an error and a call failure. Generally I think any place where a 3rd party generates signaling messages in addition to the endpoints is ripe for race conditions.

fluffy:
OK - update the actually text to get what I was trying to get at. Regardless of what state you are in or if the rollback happens on setLocal or setRemote, exactly the same thing happens which is you go back go previous stable state.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3174,17 +3164,63 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+
+        <t>  The rollback is used to recover from possible errors and
+        return to stable state the system was in before a new offer /
+        answer negotiation started. </t>
+
+        <t> When either setLocalDescription or setRemoteDescription
+        method is called with a a session description of type
+        "rollback" a rollback occurs. </t>
+
+        <t> If the system is currently in the "stable" state, no
```
juberti: Should this be a no-op or an error? I can see arguments either way; rolling back from a stable state to the same state is a no-op, but if the app was really trying to roll back to the previous state, it should trigger an error.

Generally it seems like this would most likely happen if you were confused about the current state, so an error seems better to me.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3174,17 +3164,63 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        further processing happens. </t>
+
+        <t>A rollback MUST discard any proposed changes to the session,
+        returning the state machine to the stable state, and setting
+        the pending local and/or remote description (see <xref
+        target="sec.pendinglocaldescription" /> and <xref
+        target="sec.pendingremotedescription" />) to null.  Any
+        resources or candidates that were allocated by the abandoned
+        local description are discarded; any media that is received
+        will be processed according to the previous local and remote
```
juberti: What about media that was sent, e.g. in response to a pranswer? (This text never fully tried to cover rollback from a pranswer.)
I think this section would need to be expanded a fair amount to cover this.

draft-ietf-rtcweb-jsep.xml
juberti: Asking @ekr to weigh in

draft-ietf-rtcweb-jsep.xml
juberti: Asking @ekr to weigh in

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1378,52 +1383,23 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+             <t>A rollback is performed by supplying a session
+             description of type "rollback" with empty contents to
+             either setLocalDescription or setRemoteDescription. It is
+             further described in <xref
+             target="sec.processing-a-rollback"/>.  Rollback can only
+             be used to cancel proposed changes; there is no support
+             for rolling back from a stable state to a previous stable
+             state. Note that this implies that once the answerer has
+             performed setLocalDescription with his answer, this
+             cannot be rolled back and neither the endpoint offerer or
```
juberti: This sentence is awkward, and may need to change if we allow pranswer rollbacks.

ekr: As Justin says

draft-ietf-rtcweb-jsep.xml
ekr: Well, the reason I suggested it is that "current" is a term of art here, and you're actually using the pending description

draft-ietf-rtcweb-jsep.xml
ekr: +1

ekr: In fact, it should go in the state machine picture

draft-ietf-rtcweb-jsep.xml
ekr: Agreed.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3174,17 +3164,63 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        <t> When either setLocalDescription or setRemoteDescription
+        method is called with a a session description of type
+        "rollback" a rollback occurs. </t>
+
+        <t> If the system is currently in the "stable" state, no
+        further processing happens. </t>
+
+        <t>A rollback MUST discard any proposed changes to the session,
+        returning the state machine to the stable state, and setting
+        the pending local and/or remote description (see <xref
```
ekr: maybe description(s)

draft-ietf-rtcweb-jsep.xml
ekr: previous -> current

ekr: Note, the text about whether a pranswer is the current description is vague

"   The pendingLocalDescription method returns a copy of the local
   description currently in negotiation - i.e., a local offer set
   without any corresponding remote answer - in addition to any local
   candidates that have been generated by the ICE agent since the local
   description was set.
"

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3174,17 +3164,63 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+        local description are discarded; any media that is received
+        will be processed according to the previous local and remote
+        descriptions.</t>
+
+        <t>A rollback will disassociate any RtpTransceivers that were
+        associated with m= sections by the application of the
+        rolled-back session description (see <xref
+        target="sec.applying-a-remote-desc" /> and <xref
+        target="sec.applying-a-local-desc" />). This means that some
+        RtpTransceivers that were previously associated will no longer
```
ekr: associated with...?

draft-ietf-rtcweb-jsep.xml
juberti: I considered this implicit from the surrounding text, i.e., the previous clause refers to the pending description, and this clause is positioned as an alternative. IOW, seems a bit redundant to spell this out, although I don't feel super strongly.

draft-ietf-rtcweb-jsep.xml
juberti: indeed

fluffy:
OK - this has been rebased to current master.

draft-ietf-rtcweb-jsep.xml
fluffy: We all agree roll back from stable state should cause an error

fluffy:
so just if you are trying to figure out what happened with this, i think EKR made a separate PR that had his fixed up version of this text thus all the conflicts with this and the base.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -395,6 +395,11 @@ setRemote(ANSWER) |                                   |
           </figure>
         </t>
 
+        <t> In addition to these state transitions, there is a
+        rollback operation, described in <xref
+        target="sec.rollback"/>, that returns the system to the stable
+        state. Rollback is typically used to recover when an offer is rejected.</t>
```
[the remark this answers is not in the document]
ekr: note to self: this is covered.

draft-ietf-rtcweb-jsep.xml
ekr: Note to self: this is covered.

juberti:
Agree with plan.

PR #748, opened 2017-06-23
new media formats and hdr extensions do not cause an error

### LINKED ISSUE #675: Clarify RFC 3264-related sub-bullet
BLOCKER: Page 68 contains a sub-bullet that says "If the m= section references any media formats, RTP header extensions, or RTCP feedback mechanisms that were not present in the corresponding m= section in the offer, this indicates a negotiation problem and MUST result in an error." No. This is a violation of RFC 3264. Section 6.1 of that document, when talking about generating answers, specifically stipulates:
   The stream MAY indicate additional media formats, not listed in the    corresponding stream in the offer, that the answerer is willing to    send or receive (of course, it will not be able to send them at this    time, since it was not listed in the offer).  
Note, for example, that this is *exactly* the behavior that the CN handling I describe above relies on. General handling for header extensions and (I believe) feedback mechanisms is intended to operate the same way. This sub-bullet needs to be remove in its entirety.

fluffy:
Fixes #675

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3915,11 +3915,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>If the m= section references RTCP feedback mechanisms
             that were not present in the corresponding m= section in
             the offer, this indicates a negotiation problem and MUST
-            result in an error.  New RTP header extensions values are
-            allowed as described in <xref target="RFC5285" />, Section
-            6.  New media types in the answer can be used as described
-            in <xref target="RFC3264" />, Section 7.</t>
-
+            result in an error. However, new media formats and new 
+            RTP header extension values are permitted in the answer, as
+            described in <xref target="RFC3264" />, Section 7, and   
+            <xref target="RFC5285" />, Section 6.</t>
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3912,11 +3912,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-            header extensions, or RTCP feedback mechanisms that were
-            not present in the corresponding m= section in the offer,
-            this indicates a negotiation problem and MUST result in an
-            error.</t>
+            <t>If the m= section references RTCP feedback mechanisms
+            that were not present in the corresponding m= section in
+            the offer, this indicates a negotiation problem and MUST
+            result in an error.  New RTP header extensions values are
+            allowed as described in <xref target="RFC5285" />, Section
+            6.</t>
```
ekr: Why are you specifically calling out header extensions? What about media types.

fluffy: It seemed a point of confusion ... then again so was media types so glad to just add media types here too

fluffy: added a sentence for this

PR #750, opened 2017-06-27
Clarify interaction of VoiceActivityDetection with DTX. Fixes #661.

### LINKED ISSUE #661: Clarify VoiceActivityDetection handling
The paragraph that spans pages 46 and 47 claims that the "VoiceActivityDetection" parameter has no impact on whether the local endpoint does silence suppression for the audio it sends. The paragraph immediately preceding it says that JSEP implementations MUST NOT emit "CN" codecs if this parameter is set to "false." This only makes sense because RFC 3264 allows SDP to specify codecs that you want to receive by do not plan to generate (i.e., you can send CN even if your offer doesn't have CN in it, as long as the answer *does*). This is slightly counter-intuitive to many implementors, so it probably bears mention.

ekr:
Fixes #661

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2542,21 +2542,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             parameters for that codec MUST be specified to indicate
             that silence suppression for received audio is not desired.
             For example, when using the Opus codec, the "usedtx=0"
-            parameter would be specified in the offer.</t>
-
-            <t>Depending on the selected codec, setting the
-            "VoiceActivityDetection" option may or may not impact
-            whether the JSEP implementation performs silence suppression
-            for the audio it sends. In particular, for Opus and for
-            codecs without internal silence suppression which therefore use
-            CN, the SDP indication is unilateral and therefore only
-            the peer's indication matters for whether the JSEP
-            implementation performs silence suppression. For G.729,
-            the "dtx" parameter
-            as specified in <xref target="RFC5459"/>, Section 5.2.1,
-            is bidirectional and therefore
-            setting "VoiceActivityDetection" to "false" disables
-            silence suppression in both directions.</t>
+            parameter would be specified in the offer. In addition,
+            the implementation MUST NOT use silence suppression for
+            media it generates, regardless of whether the "CN"
+            codecs or DTX fmtp parameters appear in the peer's
+            description. The impact of these rules is that silence
+            suppression depends on mutual agreement of both sides,
+            which is consistent with the current behavior of G.729,
+            as specified in as specified in <xref target="RFC5459"/>, Section 5.2.1,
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3077,7 +3081,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             on a per-codec basis (e.g., if the offerer somehow offered
             support for CN but set "usedtx=0" for Opus, setting
             VoiceActivityDetection to true would result in an answer
-            with CN codecs and "usedtx=0").</t>
+            with CN codecs and "usedtx=0"). The impact of this rule is that
+            an answerer will not offer silence suppression with any
```
juberti: The use of the word 'offer' here in an answer context is somewhat confusing.

Perhaps 'offer' -> 'try to use'

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2542,13 +2542,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-
-            <t>Note that setting the "VoiceActivityDetection" parameter
-            when generating an offer is a request to receive audio with
-            silence suppression. It has no impact on whether the local
-            endpoint does silence suppression for the audio it
-            sends.</t>
+            parameter would be specified in the offer. In addition,
+            the implementation MUST NOT use silence suppression for
+            media it generates, regardless of whether the "CN"
+            codecs or DTX fmtp parameters appear in the peer's
```
juberti: We don't mention 'DTX' fmtp params anywhere else in this section. Perhaps 'DTX' -> 'related'

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2542,13 +2542,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            description. The impact of these rules is that silence
+            suppression depends on mutual agreement of both sides,
+            which is consistent with the current behavior of G.729,
```
juberti: I thought an illustration would be useful here, but it kind of seems like inside baseball. 

Perhaps just say 'silence suppression in JSEP depends ... both sides, which ensures consistent handling regardless of which codec is used.'

draft-ietf-rtcweb-jsep.xml
```diff
@@ -5302,6 +5306,7 @@ a=msid:bbce3ba6-abfc-ac63-d00a-e15b286f8fce
       <?rfc include='reference.RFC.6464.xml'?>
       <?rfc include='reference.RFC.6544.xml'?>
       <?rfc include='reference.RFC.3556.xml'?>
+      <?rfc include='reference.RFC.5459.xml'?>
```
juberti: This ref is no longer needed.

fluffy:
LGTM

juberti:
I am going to land this and then make a followup PR for S 5.9.

PR #751, opened 2017-06-27
Clarify codec preference handling

### LINKED ISSUE #651: Codec preference discussion unclear
Section 4.2.6, second paragraph says that codec preferences cannot add formats that were negotiated away in previous offer/answer exchanges. My recollection is that this conflicts with the decision taken by the WebRTC working group; and a quick read of the description of setCodecPreferences() in their current spec seems to confirm this (at least, by omitting this restriction). Unless I'm mistaken about the state of play, this paragraph should be modified to indicate that codec preferences *can* add back in codecs were not negotiated in previous offer/answer exchanges.

juberti:
Fixes #651

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3923,16 +3923,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             that the JSEP implementation should be sending media
             ("sendonly" for local answers, "recvonly" for remote
             answers, or "sendrecv" for either type of answer),
-            choose the media format to send; for the offerer, this will be
-            the most preferred media format from the answer that is also
-            locally supported, as described in <xref target="RFC3264" />,
-            Section 7; for the answerer, this will be
-            the most preferred media format from the offer that is also present
-            in the answer, as
-            described in <xref target="RFC3264" />, Sections 6.1. Once the
-            underlying transport layers have been established, start
-            transmitting RTP media using this
-            media format. If an SSRC has not already been
+            choose the media format to send
+            as the most preferred media format from the remote
+            description that is also locally supported, as
+            discussed in
+            <xref target="RFC3264" />, Sections 6.1 and 7, and start
+            transmitting RTP media using that format once the underlying
+            transport layers have been established.
+            If an SSRC has not already been
             chosen for this outgoing RTP stream, choose a random one.
             If media is already being transmitted, the same SSRC SHOULD
             be used unless the clockrate of the new codec is different,
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1722,9 +1722,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           createAnswer, in which case the corresponding media formats
           in the associated m= section will be excluded. The codec
           preferences cannot add media formats that would otherwise not
-          be present. This includes codecs that were not negotiated in
-          a previous offer/answer exchange that included the
-          transceiver.</t>
+          be present. TODO?</t>
 
           <t>The codec preferences of an RtpTransceiver can also
           determine the order of codecs in subsequent calls to
@@ -1944,11 +1942,6 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             transceiver, media formats MUST be generated in the
             corresponding order, and MUST exclude any codecs not
             present in the codec preferences.</t>
-
-            <t>The media formats in the answer MAY include codecs
-            present in the offer that were discarded in a previous
-            offer/answer exchange. This is necessary for compatibility
-            with third- party call control and SIP use cases.</t>
 
             <t>Unless excluded by the above restrictions, the media
             formats MUST include the mandatory audio/video codecs as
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2746,22 +2742,22 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            preferences.</t>
+
+            <t>Otherwise, the media formats on the m= line
+            MUST be generated in the same order as those offered in the
+            current remote description, excluding any currently unsupported
+            formats. Any currently available media formats that are
+            not present in the current remote description MUST be added after
+            all existing formats.</t>
+
+            <t>In either case, the media formats in the answer MUST include
```
fluffy: Do we need to talk about what happens if the offer does not have any codec supported by the answer ?

juberti: We mention it on line 2715, but let me try to tie that into the text here a bit better.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2746,22 +2742,22 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            at least one format that is present in the offer, but MAY include
+            formats that are locally supported but not present in the offer,
+            as mentioned in <xref target="RFC3264" />, Section 6.1.
```
ekr: So just to be clear, this means that if the offer has A, B and the answerer wants C, A, B, he must do A, B, C.

However, if he wants C, B, A, then he can do B, A, C.

fluffy:
LGTM and if Justin wants to tie in the stuff from my comment a bit more that is great

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3929,13 +3920,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-            as the most preferred media format from the remote
-            description that is also present in the answer, as
-            described in
-            <xref target="RFC3264" />, Sections 6.1 and 7, and start
-            transmitting RTP media once the underlying transport layers
-            have been established. If an SSRC has not already been
+            choose the media format to send; for the offerer, this will be
+            the most preferred media format from the answer that is also
+            locally supported, as described in <xref target="RFC3264" />,
+            Section 7; for the answerer, this will be
```
ekr: Kind of run-on.... Maybe make a list.

ekr: Or could you do:
"this will be the most preferred media format from the peer that is also supported"?

draft-ietf-rtcweb-jsep.xml
juberti: If the answerer has an explicit preference, that wins, and so it would be C, A, B or C, B, A.

But if the preference is implicit, then the first one is correct (A, B, C) but not the second (which also must be A, B, C).

draft-ietf-rtcweb-jsep.xml
juberti: Well, this gets at an interesting point. 3264 says that the offerer has to pick something from the answer, but the answerer has to pick something that is in both. So there is an edge case where the offerer offers A B due to codec prefs, and the answerer replies with C A B; is it legal for the offerer to send with C?

It seems illegal for you to send with something you haven't offered, and that would clean up a lot of things (including CN handling), but I'm not positive this is always true. @adamroach, any thoughts on this?

taylor-b: > So there is an edge case where the offerer offers A B due to codec prefs, and the answerer replies with C A B; is it legal for the offerer to send with C?

That's what we currently do, and I can't find anything in JSEP that disallows it.

juberti: The ambiguity comes from 6.1, which states that

   The answerer MUST send using a media format in the offer
    that is also listed in the answer,

fluffy: This is a non realistic case but I hope it helps illustrate ...  Lets say Alice can only receive 711 and 729 (prefers 729) but can also send opus (but can't receive it). Alice offers 729, 711 in that order. Bob can do all all of the above and prefers in order opus, 729, 711. Bob relies in answer with, assuming a preferences override, SDP that says opus,  729 , 711. Alice sends opus as that was bobs top choice. Bob sends 729 as that was Alice top choice.  Does that outcome seem right? (Does the spec allow an app to cause that to happen? I think it does if caller prefs are explicitly set they do, but if not set, then Bob's answer will only have 729, 711, opus and Alice will send 729 which seems like a reasonable default.

fluffy: Of 3264 has the constraint that  the answer can not send with a codec it is not capable of receiving. I don't think we should try and change that in JSEP.

juberti: I think the thing that is strange is that the offerer can send with a codec that wasn't in the offer, but the answerer cannot send with a codec that wasn't in the answer.

I agree we shouldn't try to allow the answerer to send whatever (that seems like a heavy lift), but we could ensure uniformity and force the offerer to only send things it offered.

draft-ietf-rtcweb-jsep.xml
juberti: Added some text for this - PTAL

draft-ietf-rtcweb-jsep.xml
juberti: @taylor-b , wdyt?

taylor-b: > Does that outcome seem right? (Does the spec allow an app to cause that to happen?)

RFC3264 allows it to happen. But if Bob is a JSEP implementation, he can't answer with opus, because of additional restrictions JSEP adds.

> Of 3264 has the constraint that the answer can not send with a codec it is not capable of receiving.

... If it's a "sendrecv" m= section. If it's "sendonly" (in the answer) this restriction doesn't apply.

> I think the thing that is strange is that the offerer can send with a codec that wasn't in the offer, but the answerer cannot send with a codec that wasn't in the answer.

SDP is indeed strange.

> we could ensure uniformity and force the offerer to only send things it offered.

I don't see why we should add even more additional restrictions to JSEP. Plus we've been over this several times before (at IETF meetings and editor meetings) and agreed it should be legal.

juberti: OK. I am going to make the text reflect this (that you can send with a codec you didn't put in your SDP), and that will address Eric's concern as well.

juberti:
@ekr, PTAL

PR #752, opened 2017-06-30
Add text regarding when to send using silence suppression

juberti:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3956,6 +3956,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             the outgoing RTP streams, as indicated in
             <xref target="I-D.ietf-mmusic-rid"></xref>, Section 4.</t>
 
+            <t>If the m= section is of type audio, and silence suppression
+            was configured for the send media format as a result of
```
ekr: send -> sending?

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3956,6 +3956,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            processing the remote description, and is also enabled for that
+            format in the local description, use silence suppression
+            for outgoing media, in accordance with the guidance in
+            <xref target="sec.voiceactivitydetection1" />. If these conditions
+            are not met, silence suppression MUST NOT be used.</t>
```
ekr: MUST NOT be used for sending, right? Because the other side might send it and you'll still need to process it.

draft-ietf-rtcweb-jsep.xml
juberti: I reused the term introduced on line 3950.

draft-ietf-rtcweb-jsep.xml
juberti: Good point. Will change accordingly.

juberti:
Fixes #661

draft-ietf-rtcweb-jsep.xml

PR #753, opened 2017-06-30
Don't use x=0/y=0 imageattrs

### LINKED ISSUE #646: Clarify x=0,y=0 imageattr text
The second paragraph of section 3.6.1 seems odd when talking about creating answers, and downright incorrect when talking about offers. Minimally, it needs text to explain why  x=0/y=0 would ever make more sense than using "a=sendonly/recvonly/inactive."

juberti:
Fixes #646


ekr:
This seems like a normative change. I'm not saying I'm against this, but what's the rationale?

juberti:
Adam's argument in #646, which I found to be convincing.

PR #755, opened 2017-07-03
Land partial fix for #650

juberti:
This addresses Adam's actual comment - and adds the first references to rollback in the processing section - but leaves the details to be filled in by a future PR.


ekr:
LGTM

PR #758, opened 2017-07-10
Fixing some text that deals with currentDirection.

### LINKED ISSUE #757: removeTrack should change direction based on "transceiver.direction", not "transceiver.currentDirection"
Section 4.1.3, removeTrack, states:

> Future calls to createOffer will mark the m= section associated with the sender as recvonly (if transceiver.currentDirection is sendrecv) or as inactive (if transceiver.currentDirection is sendonly).

This should say "transceiver.direction". Using "currentDirection" doesn't work because it may be null, and it also may result in un-intuitive behavior, where `removeTrack` ends up not having any effect and results in a "sendrecv" or "sendonly" direction being negotiated.

taylor-b:
Fixes #757.

`currentDirection` is meant to represent "direction in effect", whereas
`direction` represents "direction the application *wants* to negotiate".
So, the direction that `removeTrack` both inspects and modifies should
be `direction`; its purpose is to change the application's desired
direction for a transceiver, regardless of the current state of
negotiation.

Also clarifying that `currentDirection` also reflects the direction
negotiated in a provisional answer, which was the intent (I believe)
but wasn't explicitly stated. This is consistent with it being the
direction "in effect", since applying a provisional answer allows the
endpoint to start sending/receiving media.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1146,9 +1146,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           which sender should have its track removed. The sender's
           track is cleared, and the sender stops sending. Future calls
           to createOffer will mark the m= section associated with the
-          sender as recvonly (if transceiver.currentDirection is
-          sendrecv) or as inactive (if transceiver.currentDirection is
-          sendonly).</t>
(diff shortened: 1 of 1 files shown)


juberti:
@ekr, do you want to look at this? I believe this is straightforward.

PR #761, opened 2017-07-20
Document what PeerConnection states allow rollbacks.

### LINKED ISSUE #760: Should be more clear about what states/methods allow a "rollback"
We have these rules in `setLocalDescription` (and similar rules in `setRemoteDescription`).

>       *  If the type is "offer", the PeerConnection state MUST be either
>          "stable" or "have-local-offer".
> 
>       *  If the type is "pranswer" or "answer", the PeerConnection state
>          MUST be either "have-remote-offer" or "have-local-pranswer".

But they don't mention "rollback". They should. Related WebRTC issue: https://github.com/w3c/webrtc-pc/issues/1470

taylor-b:
Fixes #760.

This raises a question I haven't seen discussed before: can you roll
back a provisional answer? I'd been assuming "no" but I could be wrong.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1417,10 +1417,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
             <t>A rollback is performed by supplying a session
             description of type "rollback" with empty contents to
-            either setLocalDescription or setRemoteDescription,
-            depending on which was most recently used (i.e. if the new
-            offer was supplied to setLocalDescription, the rollback
-            should be done using setLocalDescription as well).</t>
+            either setLocalDescription or setRemoteDescription.
+            The effects MUST be the same regardless of whether
+            setLocalDescription or setRemoteDescription is called. Note
+            that the only reason rollbacks use setLocalDescription
+            and setRemoteDescription rather than an independent method
+            is for compatibility with existing implementations.</t>
+
+            <t>A rollback may be performed if the PeerConnection is in
+            any state except for "stable". This means that both offers
+            and provisional answers can be rolled back. If a rollback is
+            attempted in the "stable" state, processing MUST stop and an
+            error MUST be returned.</t>
           </section>
         </section>
         <section title="setLocalDescription"
@@ -3128,9 +3136,6 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>If the type is "pranswer" or "answer", the
             PeerConnection state MUST be either "have-remote-offer" or
             "have-local-pranswer".</t>
-
-            <t>If the type is "rollback", the PeerConnection state MUST
-            be "have-local-offer".</t>
           </list></t>
 
           <t>If the type is not correct for the current state,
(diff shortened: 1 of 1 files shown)


aboba:
In PR https://github.com/w3c/webrtc-pc/pull/1487, it is also assumed that a provisional answer can't be rolled back (either in "have-local-pranswer" or "have-remote-pranswer" states).

taylor-b:
Updated PR based on editors' discussion. PTAL.

adamroach:
As the sponsoring area director for this document, I would like to remind you that it is in IETF last call.
Any changes to the document at this point need to be the result of discussion on the ietf@ietf.org mailing list. Please ensure that this topic is discussed there, and that any changes made to the document ultimately reflect the result of those conversations.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1417,10 +1417,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-            either setLocalDescription or setRemoteDescription,
-            depending on which was most recently used (i.e. if the new
-            offer was supplied to setLocalDescription, the rollback
-            should be done using setLocalDescription as well).</t>
+            either setLocalDescription or setRemoteDescription.
+            The effects MUST be the same regardless of whether
+            setLocalDescription or setRemoteDescription is called. Note
+            that the only reason rollbacks use setLocalDescription
+            and setRemoteDescription rather than an independent method
+            is for compatibility with existing implementations.</t>
```
ekr: I would remove this "Note". I don't think it assists comprehension.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1417,10 +1417,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+
+            <t>A rollback may be performed if the PeerConnection is in
+            any state except for "stable". This means that both offers
+            and provisional answers can be rolled back. If a rollback is
+            attempted in the "stable" state, processing MUST stop and an
+            error MUST be returned.</t>
```
ekr: I am fine with this, but IIRC Cullen wanted to allow it in stable as well.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1417,10 +1417,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-            should be done using setLocalDescription as well).</t>
+            either setLocalDescription or setRemoteDescription.
+            The effects MUST be the same regardless of whether
+            setLocalDescription or setRemoteDescription is called. Note
+            that the only reason rollbacks use setLocalDescription
+            and setRemoteDescription rather than an independent method
+            is for compatibility with existing implementations.</t>
+
+            <t>A rollback may be performed if the PeerConnection is in
+            any state except for "stable". This means that both offers
```
fluffy: Should we could add "closed" to the set of states that cause error ?

draft-ietf-rtcweb-jsep.xml
taylor-b: If I were reading this document for the first time, I would be confused by why rollbacks use `setLocalDescription`/`setRemoteDescription`, and would look for an explanation, so personally I think this note assists the reader and doesn't hurt anything. Do you still want it removed? Or would you suggest some alternative wording?

taylor-b: Nevermind, I didn't notice this was already revised and committed in https://github.com/rtcweb-wg/jsep/commit/6f5dc414265cab8b76b25e5cd381a53bcb9cd5c5.

PR #794, opened 2017-08-17
Robert suggested change to intro

### LINKED ISSUE #786: Rephrase intro sentence
Suggested change to the first sentence of 1.1

  OLD:

    The thinking behind WebRTC call setup has been to fully specify and
    control the media plane, but to leave the signaling plane up to the
    application as much as possible.

  New:

    WebRTC call setup has been designed to focus on controlling the media
    plane, leaving signaling plane behavior up to the application as much as
    possible.

fluffy:
Fix #786


juberti:
Fixed my own comment, now LGTM

fluffy:
I think this got all the text from Robert, it was just not a change so it does not show up in diff.

PR #798, opened 2017-08-17
merge two bullet points

### LINKED ISSUE #782: Redundant bullets?
The 4th bullet on page 64 is ambiguous at "this m= section". It looks like
this bullet should just be merged with the previous bullet.

fluffy:
Fixes #782


ekr:
I agree with justin here. my bad for missing that.

PR #801, opened 2017-08-17
clean up session id text

### LINKED ISSUE #769: sess-id wonkiness
The fourth sentence of the second bullet of section 5.2.1 is confusing. (It
starts "It is RECOMMENDED that the <sess-id>".) RFC3265 requires the
initial sess-id high-order bit to be zero. As written, this can be read to
say that setting that bit to zero is only RECOMMENDED, and that it's
possible to have sess-ids that are something other than 64 bits. You mean
to say "Do what 3265 says to do and make the lower 63 bits random". Here's
one possible replacement for the sentence: "RFC 3265 requires that the
high-order bit of the initial <sess-id> be zero. It is RECOMMENDED that the
remaining 63 bits be initially set to a cryptographically random value."

fluffy:
Fixes #769

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1883,7 +1883,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             The sess-id MUST be representable by a 64-bit signed
             integer, and the initial value MUST be less than
             (2**62)-1, as required by <xref target="RFC3264" />. It is
-            RECOMMENDED that this be constructed by generating a
+            RECOMMENDED that the sess-id be constructed by generating a
             64-bit quantity with the two highest bits being set to
             zero and the remaining 62 bits being cryptographically
             random. The
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1880,13 +1880,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>The second SDP line MUST be an "o=" line, as specified
             in <xref target="RFC4566"></xref>, Section 5.2. The value
             of the &lt;username&gt; field SHOULD be "-".
-            <xref target="RFC3264" /> requires that the inital value
-            of the &lt;sess-id&gt; be representable as a 64-bit signed
-            integer less than (2**62)-1.  <xref target="RFC3264" />
-            requires that the two high-order bits of the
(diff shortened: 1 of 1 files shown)


fluffy:
Not sure this is correct yet. Have a read of text in bugs.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1878,14 +1878,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-            the remaining 63 bits being cryptographically random. The
+            in <xref target="RFC4566"></xref>, Section 5.2. The value
+            of the &lt;username&gt; field SHOULD be "-".
+            <xref target="RFC3264" /> requires that the inital value
+            of the &lt;sess-id&gt; be representable as a 64-bit signed
+            integer less than (2**62)-1.  <xref target="RFC3264" />
+            requires that the two high-order bits of the
+            &lt;sess-id&gt; initially be set to zero. It is
+            RECOMMENDED that the remaining 62 bits are initially set
+            to a cryptographically random value. The
```
ekr: I think that this is semantically correct, but in the style of JSEP, I would say.

"The sess-id MUST be representable by a 64-bit signed integer, and the initial value MUST be less than (2**62)-1, as required by RFC 3264. It is RECOMMENDED that this be constructed by generating a 64-bit quantity with the two highest bits being set to zero and the remaining 62 bits being cryptographically random."

juberti: That works for me.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1878,14 +1878,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-            be representable as a 64-bit signed integer. It is
-            RECOMMENDED that the &lt;sess-id&gt; be generated as a
-            64-bit quantity with the high bit being sent to zero and
-            the remaining 63 bits being cryptographically random. The
+            in <xref target="RFC4566"></xref>, Section 5.2. The value
+            of the &lt;username&gt; field SHOULD be "-".
+            The sess-id MUST be representable by a 64-bit signed
+            integer, and the initial value MUST be less than
+            (2**62)-1, as required by <xref target="RFC3264" />. It is
+            RECOMMENDED that this be constructed by generating a
```
ekr: I would say "that the sess-id be constructed by..."

ekr: ..

PR #802, opened 2017-08-17
update fields

### LINKED ISSUE #766: Last para of 3.5.2.1
Last paragraph of 3.5.2.1 - "all of these fields" is ambiguous. Please
state precisely which fields you mean.

fluffy:
Fixes #766

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -607,10 +607,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             used instead.</t>
 
             <t>When creating an IceCandidate object, JSEP
-            implementations MUST populate the all of the
+            implementations MUST populate each of the
             candidate, ufrag, m= section index, and MID fields.
             Implementations MUST also be prepared to receive
-            candidates with some of the fields missing, as mentioned
+            objects with some of the fields missing, as mentioned
             above.</t>
 
           </section>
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -607,7 +607,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             used instead.</t>
 
             <t>When creating an IceCandidate object, JSEP
-            implementations MUST populate all of these fields.</t>
+            implementations MUST populate both the 
+            m=section index field and the MID field.</t>
+
```
ekr: This doesn't seem quite right. Aren't you required to also populate ufrag and candidate

draft-ietf-rtcweb-jsep.xml
```diff
@@ -607,7 +607,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             used instead.</t>
 
             <t>When creating an IceCandidate object, JSEP
-            implementations MUST populate all of these fields.</t>
+            implementations MUST populate the aforementioned
```
ekr: perhaps "all of the"

ekr: actually, I think this may require us to also say, "though MUST also be prepared to receive candidates with some of the fields missing, as mentioned above".

PR #812, opened 2017-08-18
add lip sync info

### LINKED ISSUE #776: LS needs a reference
"LS" (lipsync) is first used at 4.1.2, with no context for what it is.
Please add a reference.

fluffy:
Fixes #776

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1132,10 +1132,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           the track with other tracks in the same MediaStream, so that
           they can be added to the same "LS" group when creating an
           offer or answer.
-          Adding tracks to the same "LS" group allows the play back of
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1131,7 +1131,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           PeerConnection, using the MediaStream argument to associate
           the track with other tracks in the same MediaStream, so that
           they can be added to the same "LS" group when creating an
-          offer or answer. addTrack attempts to minimize the number of
+          offer or answer.
+          Adding tracks to the same "LS" group allows the play back of
+          theses tracks to be sycronized so there is lip sync between
+          audio and video as described in <xref
+          target="RFC5888"></xref>.
```
ekr: I think you could shorten this considerably. How about:
"thus allowing these tracks to be synchronized when played out, as described in [RFC5888]"

juberti:
@ekr, PTAL at the reformulation I added.

PR #813, opened 2017-08-18
remote ice candiate parsing error 

### LINKED ISSUE #779: Why mention 'if parsed successfully'
At 4.1.17's first paragraph: Why is "if parsed successfully" necessary to
call out? I suggest removing the phrase.

fluffy:
Fixes #779

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1627,7 +1627,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         <section title="addIceCandidate" anchor="sec.addicecandidate">
 
           <t>The addIceCandidate method provides a remote candidate to
-          the ICE agent, which, if parsed successfully, will be added
+          the ICE agent which will be added
           to the current and/or pending remote description according to
           the rules defined for Trickle ICE. The pair of MID and ufrag
           is used to determine the m= section and ICE candidate
@@ -1635,7 +1635,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           present, the m= section index is used to look up the locally
           generated MID (see
           <xref target="sec.applying-a-remote-desc" />), which is used
-          in place of a supplied MID. If these values or the candidate
+          in place of a supplied MID. If the remote candidate is not
+          succesfullf parsed, or these values or the candidate
           string are invalid, an error is generated.</t>
 
           <t>The purpose of the ufrag is to resolve ambiguities when
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1635,9 +1635,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           present, the m= section index is used to look up the locally
           generated MID (see
           <xref target="sec.applying-a-remote-desc" />), which is used
-          in place of a supplied MID. If the remote candidate is not
-          successfully parsed, or these values or the candidate
-          string are invalid, an error is generated.</t>
+          in place of a supplied MID.
+          If the remote candidate is not successfully parsed, or these
+          values or the candidate string are invalid, the JSEP
+          implementation will generate an error without adding the
+          candidate.</t>
 
           <t>The purpose of the ufrag is to resolve ambiguities when
           trickle ICE is in progress during an ICE restart. If the
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1626,39 +1626,53 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         </section>
         <section title="addIceCandidate" anchor="sec.addicecandidate">
 
-          <t>The addIceCandidate method provides a remote candidate to
-          the ICE agent, which will be added
+          <t>The addIceCandidate method provides an update to
+          the ICE agent via an IceCandidate object 
+          <xref target="sec.ice-candidate-format"/>. If the
+          IceCandidate's candidate field is filled in, the IceCandidate
+          is treated as a new remote ICE candidate, which will be added
           to the current and/or pending remote description according to
-          the rules defined for Trickle ICE. The pair of MID and ufrag
-          is used to determine the m= section and ICE candidate
-          generation to which the candidate belongs. If the MID is not
-          present, the m= section index is used to look up the locally
-          generated MID (see
-          <xref target="sec.applying-a-remote-desc" />), which is used
-          in place of a supplied MID.</t>
-
-          <t>The purpose of the ufrag is to resolve ambiguities when
-          trickle ICE is in progress during an ICE restart. If the
-          ufrag is absent, the candidate MUST be assumed to belong to
-          the most recently applied remote description. Connectivity
-          checks will be sent to the new candidate.</t>
-
-          <t>If any of the the m= section index, MID, ufrag, or the
-          candidate string are not valid, then the JSEP implementation
-          will generate an error without adding the candidate.</t>
-
-          <t>This method can also be used to provide an
-          end-of-candidates indication to the ICE agent, as defined in
-          <xref target="I-D.ietf-ice-trickle" />). The MID and ufrag
-          are used as described above to determine the m= section and
-          ICE generation for which candidate gathering is complete. If
-          the ufrag is not present, then the end-of-candidates
-          indication MUST be assumed to apply to the relevant m=
-          section in the most recently applied remote description. If
-          neither the MID nor the m= index is present, then the
+          the rules defined for Trickle ICE. Otherwise, the IceCandidate
+          is treated as an end-of-candidates indication, as defined in
+          <xref target="I-D.ietf-ice-trickle" />).</t>
+          
+          <t>In either case, the MID and ufrag fields from the supplied 
+          IceCandidate are used to
+          determine which m= section and ICE candidate generation the
+          IceCandidate is for. In particular, the ufrag is used to resolve
+          ambiguities when trickle ICE is in progress during an ICE restart.</t>
+          
+          <t>Specific processing for each case is outlined below. If any
+          IceCandidate fields contain invalid values, or an error
+          occurs during the processing, the supplied IceCandidate MUST be ignored
+          and an error MUST be returned.</t>
+      
+          <t>If the IceCandidate is a new remote candidate:
+            <list style="symbols">
+              <t>The m= section index field must be present.</t>
+              
+              <t>If the MID field is not filled in, the m=
+              section index MUST be used to look up the locally generated
+              MID (see <xref target="sec.applying-a-remote-desc" />), which
+              is used in place of a supplied MID.</t>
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1627,15 +1627,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           to the current and/or pending remote description according to
           the rules defined for Trickle ICE. The pair of MID and ufrag
           is used to determine the m= section and ICE candidate
           generation to which the candidate belongs. If the MID is not
           present, the m= section index is used to look up the locally
           generated MID (see
           <xref target="sec.applying-a-remote-desc" />), which is used
-          in place of a supplied MID. If these values or the candidate
+          in place of a supplied MID. If the remote candidate is not
+          successfully parsed, or these values or the candidate
```
ekr: I don't think you need to expand on the errors but rather on the impact:
"If these values or the candidate string are invalid, the JSEP implementation will generate an error without adding the candidate"

fluffy: so I left in the syntax error as that was what we wanted to move forward but not exactly what you asked for. Have a look

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1627,16 +1627,19 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           the rules defined for Trickle ICE. The pair of MID and ufrag
           is used to determine the m= section and ICE candidate
           generation to which the candidate belongs. If the MID is not
           present, the m= section index is used to look up the locally
           generated MID (see
           <xref target="sec.applying-a-remote-desc" />), which is used
-          in place of a supplied MID. If these values or the candidate
-          string are invalid, an error is generated.</t>
+          in place of a supplied MID.
+          If the remote candidate is not successfully parsed, or these
```
ekr: How is "not successfully parseD" different from the candidate string being invalid

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1627,16 +1627,19 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         <section title="addIceCandidate" anchor="sec.addicecandidate">
 
           <t>The addIceCandidate method provides a remote candidate to
-          the ICE agent, which, if parsed successfully, will be added
+          the ICE agent which will be added
```
juberti: Comma after 'agent'

draft-ietf-rtcweb-jsep.xml
juberti: I think this para needs to come after the para below regarding ufrag.

juberti:
Upon review, it became apparent that this section needed more work to handle the end-of-candidates case properly. I took a first cut at reworking it, but more polish is needed.

@ekr, @fluffy, PTAL to see if you see anything egregiously wrong before I finish polishing it up.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1626,36 +1626,53 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          ICE generation for which candidate gathering is complete. If
-          the ufrag is not present, then the end-of-candidates
-          indication MUST be assumed to apply to the relevant m=
-          section in the most recently applied remote description. If
-          neither the MID nor the m= index is present, then the
+          the rules defined for Trickle ICE. Otherwise, the IceCandidate
+          is treated as an end-of-candidates indication, as defined in
+          <xref target="I-D.ietf-ice-trickle" />).</t>
+          
+          <t>In either case, the MID and ufrag fields from the supplied
```
fluffy: I think this sentence needs some tweaking to take care of case there is an m= index but no MID

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1626,36 +1626,53 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          determine which m= section and ICE candidate generation the
+          IceCandidate is for. In particular, the ufrag is used to resolve
+          ambiguities when trickle ICE is in progress during an ICE restart.</t>
+          
+          <t>Specific processing for each case is outlined below. If any
+          IceCandidate fields contain invalid values, or an error
+          occurs during the processing, the supplied IceCandidate MUST be ignored
+          and an error MUST be returned.</t>
+      
+          <t>If the IceCandidate is a new remote candidate:
```
fluffy: I'd phrase this as "candidate field is not empty" as reduce the level of indirection to the the "new remote" term but happy to go as you have it if you feel that is better.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1626,36 +1626,53 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            <list style="symbols">
+              <t>The m= section index field must be present.</t>
+              
+              <t>If the MID field is not filled in, the m=
+              section index MUST be used to look up the locally generated
+              MID (see <xref target="sec.applying-a-remote-desc" />), which
+              is used in place of a supplied MID.</t>
```
fluffy: I don't typically care about edge error cases but if the m= index has a MID that does not the match the MID here, we could have an error but I would tend towards, if there is a MID, you use the MID and ignore index. Trying to say I am fine with text as is but consider if you want to do something with a mismatch error.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1626,36 +1626,53 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          
+              <t>If the ufrag field is not filled in, the candidate object MUST be
+              assumed to refer to the most recently applied remote description.</t>
+              
+              <t>If the candidate is valid, it is added to the ICE Agent as a new
+              remote candidate. Connectivity checks will be sent to the new candidate.</t>
```
fluffy: Say something about we get an error if it is not valid.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1626,36 +1626,53 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            </list>
+          </t>
+          
+          <t>If the IceCandidate is an end-of-candidates indication:
```
fluffy: again here, I would tend to say "candidate field is empty" instead of end-of-candiates

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1626,36 +1626,53 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+            <list style="symbols">
+                <t> 
+          If neither the MID nor the m= index is present, then the
```
fluffy: This sentence now seems very out of place in the sort of pseudo code algorithm style that all the above is in.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1626,36 +1626,53 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          <xref target="I-D.ietf-ice-trickle" />). The MID and ufrag
-          are used as described above to determine the m= section and
-          ICE generation for which candidate gathering is complete. If
-          the ufrag is not present, then the end-of-candidates
-          indication MUST be assumed to apply to the relevant m=
-          section in the most recently applied remote description. If
-          neither the MID nor the m= index is present, then the
+          the rules defined for Trickle ICE. Otherwise, the IceCandidate
+          is treated as an end-of-candidates indication, as defined in
+          <xref target="I-D.ietf-ice-trickle" />).</t>
```
ekr: Unnecessary )

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1626,36 +1626,53 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          </t>
+          
+          <t>If the IceCandidate is an end-of-candidates indication:
+            <list style="symbols">
+                <t> 
+          If neither the MID nor the m= index is present, then the
           indication MUST be assumed to apply to all m= sections in the
           most recently applied remote description.</t>
-
+            </list>
```
ekr: You probably need to talk about the index case.

ekr: Don't we already have text about how to identify the right m-line in 3.5.2.1?

draft-ietf-rtcweb-jsep.xml
juberti: Done

draft-ietf-rtcweb-jsep.xml
juberti: Done

draft-ietf-rtcweb-jsep.xml
juberti: Done

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1626,39 +1626,32 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          <t>This call will result in a change to the state of the ICE
-          Agent, and may result in a change to media state if it
-          results in connectivity being established.</t>
+          the rules defined for Trickle ICE. Otherwise, the IceCandidate
+          is treated as an end-of-candidates indication, as defined in
+          <xref target="I-D.ietf-ice-trickle" />.</t>
+
+          <t>In either case, the m= section index, MID, and ufrag fields from
+          the supplied IceCandidate are used to determine which m= section
+          and ICE candidate generation the IceCandidate is for, as described
```
ekr: "is for" seems a bit informal. perhaps "the corresponding m= section"?

draft-ietf-rtcweb-jsep.xml
juberti: Mentioned this a few sentences above, but I decided to rewrite this section to make it clearer.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1626,39 +1626,32 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
+          in <xref target="sec.ice-candidate-format" /> above. In the
+          specific case of an end-of-candidates indication, both the m= section
+          index and MID may be omitted to mean that the indication applies
+          to all m= sections in the specified ICE candidate generation.</t>
```
ekr: You've switched from the receiver's perspective to the senders here. Perhaps
"If the object indicates end-of-candidates and both the m= section index and MID re omitted, that means that the candidate applies. Otherwise, if both are omitted, then the JSEP implementation MUST ignore the IceCandidate and return ann error"

draft-ietf-rtcweb-jsep.xml
juberti: Went with "belongs to"

draft-ietf-rtcweb-jsep.xml
juberti: Agreed; rewrote to be clear this is from the receiver's perspective

PR #814, opened 2017-08-18
Fix reference. Try to clarify. Fixes #768

### LINKED ISSUE #768: Profile string references
Slightly more than minor maybe: The first paragraph of 5.1.2 has at least a
reference error. UDP/TLS/RTP/SAVPF is not specified in RFC7850. It's in
RFC5764. Assuming that this is only a reference error, then the last
sentence seems like a stray. It doesn't seem to be in the right context
("this advertisement" doesn't cast back to the discussion of the two
profiles above it, which are both UDP only). Please clarify the sentence,
giving it the right context, or remove it.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1799,16 +1799,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
-          each data m= line they produce in an offer. Because ICE can
-          select either UDP
+          each data m= line they produce in an offer. Although these
+          profiles are formally associated with UDP,
+          ICE can select either UDP
           <xref target="RFC5245" /> or TCP
           <xref target='RFC6544' /> transport depending on network
-          conditions, this advertisement is consistent with ICE
-          eventually selecting either either UDP or TCP.</t>
+          conditions, even while advertising a UDP profile.</t>
```
juberti: I would have used 'when' instead of 'while', but otherwise LGTM.

PR #816, opened 2017-08-18
Fixes #771.

### LINKED ISSUE #771: Para in S 5.4 is confusing
The second paragraph of section 5.4 is confusing. A slight restructuring of
the sentences would remove much of the potential confusion, I think. I
suggest replacing the first sentence (which is very complex) with this:
"After calling setLocalDescription, the application MAY modify the SDP to
reduce the capabilities in the offer before sending it to the far side, as
long as it remains a valid SDP offer and specifies a subset of what was in
the original offer. Likewise, an application that has received an offer
from a peer MAY modify the received SDP, subject to those same constraints,
before calling setRemoteDescription."

ekr:
I took Robert's text and removed the line about "This is safe because
it's not exactly right (you can add new codecs, etc. as we discussed)
and it didn't seem to be needed. If people think it is, I can rewrite.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3096,14 +3096,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
         createOffer/createAnswer options or RtpTransceiver APIs MUST be
         used.</t>
 
-        <t>Note that the application MAY modify the SDP to reduce the
-        capabilities in the offer it sends to the far side
-        (post-setLocalDescription) or the offer that it installs from
-        the far side (pre-setRemoteDescription), as long as it remains
-        a valid SDP offer and specifies a subset of what was in the
-        original offer. This is safe because the answer is not
-        permitted to expand capabilities, and therefore will just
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3096,14 +3096,12 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
 
-        <t>Note that the application MAY modify the SDP to reduce the
-        capabilities in the offer it sends to the far side
-        (post-setLocalDescription) or the offer that it installs from
-        the far side (pre-setRemoteDescription), as long as it remains
-        a valid SDP offer and specifies a subset of what was in the
-        original offer. This is safe because the answer is not
-        permitted to expand capabilities, and therefore will just
-        respond to what is present in the offer.</t>
+        <t>After calling setLocalDescription, the application MAY modify the SDP to
```
juberti: Do we want to explicitly mention offers here, i.e.

> After calling setLocalDescription with an offer, ...

Or do we want to explicitly allow the same munging on answers (which almost certainly should be legal)?

PR #823, opened 2017-08-18
Change wording of DTMF instructions

### LINKED ISSUE #773: setRD section shouldn't talk about sending DTMF
On page 69, at "If the application attempt to transmit DTMF when using a
media format that does not have a corresponding telephone-event format,
this MUST result in an error". This sentence is out-of-place. The section
is about applying a remote description. If the document needs to retain the
sentence, it belongs off in some "processing media" section. In particular,
the current placement of text is confusing about what and when an error
could be returned, and it could be misread to say that an error should be
returned to the setRemoteDescription call.

juberti:
Fixes #773


juberti:
@fluffy, PTAL

PR #825, opened 2017-08-18
Minor clarifications to security considerations. Fixes #792

### LINKED ISSUE #792: Clarify sentence in security seciton
From "Opsdir last call review of draft-ietf-rtcweb-jsep-21"

A minor observation, not even a nit, is that the text in the Security
Considerations "While formally the JSEP interface is an API, it is better to
think of it is an Internet protocol, with the JS being untrustworthy from the
perspective of the endpoint. " could be more clear.

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -5223,28 +5223,28 @@ a=msid:bbce3ba6-abfc-ac63-d00a-e15b286f8fce
       describes security considerations for this document.</t>
 
       <t>While formally the JSEP interface is an API, it is better to
-      think of it is an Internet protocol, with the JS being
-      untrustworthy from the perspective of the endpoint. Thus, the
+      think of it is an Internet protocol, with the JavaScript being
+      untrustworthy from the perspective of the JSEP implementation. Thus, the
       threat model of
-      <xref target="RFC3552" /> applies. In particular, JS can call the
+      <xref target="RFC3552" /> applies. In particular, application JavaScript can call the
```
juberti: I was thinking the first JavaScript (line 5226) was the best fit for "application JavaScript", but open to other ideas on this.

PR #831, opened 2017-08-26
Use first-fit algorithm, rather than highest-q

### LINKED ISSUE #745: If we receive an imageattr with q=1.0 but sar!=1.0, should we use the next best imageattr?
@gloinul sayeth:
> I think this selection algorithm has a big weakness, that it can select an attribute that results in no encoding at all. I would strongly propose that the selection criteria is to pick the alternative with the highest q that allows for an encoding, i.e. not blocked by the rules preventing an encoding to be generated by the below rules, such as SAR!=1.0.

@juberti sayeth:
> What should happen then in the case of only a SAR!=1.0 imageattr?

juberti:
Fixes #745

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -10,7 +10,7 @@
 <?rfc rfcedstyle="no" ?>
 <?rfc docmapping="yes" ?>
 <?rfc tocdepth="4"?>
-<rfc category="std" docName="draft-ietf-rtcweb-jsep-22"
+<rfc category="std" docName="draft-ietf-rtcweb-jsep-latest"
 ipr="trust200902">
   <front>
     <title abbrev="JSEP">JavaScript Session Establishment
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -837,7 +837,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>If the attribute includes a "sar=" (sample aspect ratio)
               value set to something other than "1.0", indicating the
               receiver wants to receive non-square pixels, this cannot be
-              satisfied and MUST be treated as an error.</t>
+              satisfied and the attribute MUST NOT be used.</t>
 
               <t>If the encoder resolution exceeds the maximum size permitted by the
               attribute, and the encoder is allowed to adjust its resolution,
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -841,15 +846,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
               <t>If the encoder resolution is less than the minimum size permitted
-              by the attribute, the encoding MUST NOT be transmitted, and an
-              error SHOULD be raised to the application; the encoder
+              by the attribute, this MUST be treated as an error; the encoder
               MUST NOT apply upscaling. JSEP implementations SHOULD avoid this
               situation by allowing receipt of arbitrarily small resolutions,
               perhaps via fallback to a software decoder.</t>
+
+              <t>If the encoder resolution is within the maximum and minimum
+              sizes, no action is needed.</t>
```
ekr: I am generally good with this, but I think "MUST be treated as an error" is confusing. Perhaps "this set of attributes MUST NOT be used"?

juberti: I think you mean 'this attribute MUST NOT be used', but I think this is a good suggestion; I wasn't completely happy with the current text either.

juberti: Updated accordingly

ekr:
Still LGTM

draft-ietf-rtcweb-jsep.xml
```diff
@@ -832,7 +837,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
               <t>If the attribute includes a "sar=" (sample aspect ratio)
               value set to something other than "1.0", indicating the
               receiver wants to receive non-square pixels, this cannot be
-              satisfied and the sender MUST NOT transmit the encoding.</t>
+              satisfied and the attribute MUST NOT be used.</t>
```
fluffy: I think we need to talk about par as well, I would suggest adding:

If the attribute includes a "par=" (picture aspect ratio) value set to something other than the encoder produces, this cannot be satisfied and the attribute MUST NOT be used.

juberti:
@gloinul, we have consensus on this from the editor team, but please comment here if you see anything else and I'll take a look.

gloinul:
Thanks, I think it looks good in general. 

Regarding the "par" question @fluffy raised the current text promotes some picture over correct picture aspect ratio. Having a bullet that one selects one that has PAR that the sending side supports to send can result in more optimal choice but also may result in a slight increase in blocked video where a stretched video may be shown instead.

PR #833, opened 2017-08-31
Add text to handle no m= section case

### LINKED ISSUE #832: Offer without m-section allowed
I failed to find any text in webrtc-pc or JSEP which indicates that the createOffer() call should fail if no RtpTransceiver or data channel has been added to a PeerConnection.

Webrtc-pc contains the following:

> 5. If this inspection failed for any reason, reject p with a newly created OperationError and abort these steps.

But section `5.2.1. Initial Offers` does not seem to contain any wording indicating to treat the absence of any RtpTransceiver and DataChannel as such an failure.

I think such a scenario should be treated as a failure, as I fail to see any value in generating an empty offer.

juberti:
Fixes #832

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1706,7 +1706,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           associated m= section on future calls to createOffer and
           createAnswer.
           The permitted values for direction are "recvonly",
-          "sendrecv", "sendonly", and "inactive", mirroring the 
+          "sendrecv", "sendonly", and "inactive", mirroring the
           identically-named directional attributes defined in <xref
           target="RFC4566"></xref>, Section 6.
           </t>
@@ -1962,9 +1962,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <t>The next step is to generate m= sections, as specified in
           <xref target="RFC4566" />, Section 5.14. An m= section is
           generated for each RtpTransceiver that has been added to the
-          PeerConnection, excluding any stopped RtpTransceivers. This
+          PeerConnection, excluding any stopped RtpTransceivers; this
           is done in the order the RtpTransceivers were added to the
-          PeerConnection.</t>
+          PeerConnection. If there are no such RtpTransceivers, no
+          m= sections are generated, as discussed in
+          <xref target="RFC3264" />, Section 5.</t>
 
           <t>For each m= section generated for an RtpTransceiver,
           establish a mapping between the transceiver and the index of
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1962,9 +1962,11 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           <t>The next step is to generate m= sections, as specified in
           <xref target="RFC4566" />, Section 5.14. An m= section is
           generated for each RtpTransceiver that has been added to the
-          PeerConnection, excluding any stopped RtpTransceivers. This
+          PeerConnection, excluding any stopped RtpTransceivers; this
           is done in the order the RtpTransceivers were added to the
-          PeerConnection.</t>
+          PeerConnection. If there are no such RtpTransceivers, no
+          m= sections are generated, as discussed in
+          <xref target="RFC3264" />, Section 5.</t>
```
ekr: Do you want to say the semantics of this, e.g., 
"generated, allowing more streams to be added later, as discussed"

juberti: I figured that was sort of implicit in the context of this document, but wdyt?

ekr: It would have made it a bit easier for me to read, but I don't feel strongly about it.

ekr:
@fluffy: as Justin points out, 3264 explicitly authorizes this.

draft-ietf-rtcweb-jsep.xml
juberti: Took a shot at this, PTAL

fippo:
i'm a bit late but... without this, the following sequence of events is always going to happen:
* createOffer
* setLocalDescription
* onicecandidate(null) / icegatheringstatechange(complete)

Executing this snippet in Chrome
```
var pc = new RTCPeerConnection();
pc.onicecandidate = (e) => console.log(e.candidate);
pc.createOffer({offerToReceiveAudio: false}).then(offer => pc.setLocalDescription(offer));
```
never shows a candidate. Is that the intended behaviour?

Also the ice gathering state stays "new"? My confusion about this is probably related to [set theory](https://math.stackexchange.com/questions/202452/why-is-predicate-all-as-in-allset-true-if-the-set-is-empty) ;-)

nils-ohlmeier:
I guess if we need to allow empty SPDs (no m-sections) then ICE related events should fire right away as they know that nothing is going to happen any way.

taylor-b:
@fippo I don't see why any events need to fire, since ICE gathering never starts in the first place and there are no `RTCIceTransports` in that case.

fippo:
Also that set theory thing might be more real than I thought at first:
http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnectionstate says things like `All RTCIceTransport s and RTCDtlsTransport s are in the "connected", "completed" or "closed" state and at least of them is in the "connected" or "completed" state.`
This particular example seems to be guarded by a state change but having an empty set of ice transports makes things complicated.
Try `[].every(t => t.state === "connected")` for your amusement.

PR #839, opened 2017-09-22
Clarify aspect ratio and rounding

### LINKED ISSUE #838: Clarify apect ratio preservation with rounding
In spite of what was said at the W3C WebRTC WG virtual meeting yesterday I think the aspect ratio preservation can be an issue also in context of ```imageattr```. Just to take as example that the video resolution of the track is 595 by 753 (e.g. capturing a part of the screen) and ```imageattr``` has limited what can be transmitted to 480 by 640. Scaling according to JSEP section 3.6 (if I read right) gives a transmission resolution of 480 by 607.4621.. but since we can't transmit fractional pixels we would transmit 480 by 607 and have a slight aspect ratio mismatch.

The spec should clarify what to do in situations like this.

juberti:
Fixes #838


stefhak:
LGTM.

PR #849, opened 2018-06-19
Add bullet about rejecting all m= sections in a BUNDLE group.

taylor-b:
If an answerer rejects the "offerer tagged" m= section (the one
listed first in "a=group:BUNDLE"), the BUNDLE spec requires rejecting
*all* m= sections in the BUNDLE group.

This means that stopping one transceiver between setting an offer and
creating an answer may potentially cause multiple m= sections to be
rejected.

See: https://github.com/w3c/webrtc-pc/issues/1858

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2735,6 +2735,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>The bundle policy is "balanced", and this is not the
             first m= section for this media type or in the same bundle
             group as the first m= section for this media type.</t>
+
+            <t>This m= section is in a bundle group, and the group's
+            offerer tagged m= section is being rejected due to one of
+            the above reasons. This requires all m= sections in the
+            bundle group to be rejected, as specified in
+            <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation" />,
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2735,6 +2735,13 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>The bundle policy is "balanced", and this is not the
             first m= section for this media type or in the same bundle
             group as the first m= section for this media type.</t>
+
+            <t>This m= section is in a bundle group, and the group's
+            offerer tagged m= section is being rejected due to one of
+            the above reasons. This requires all m= sections in the
+            bundle group to be rejected, as specified in
+            <xref target="I-D.ietf-mmusic-sdp-bundle-negotiation" />,
+            Section 7.3.3.</t>
```
henbos: [BUNDLE section 7.3.3](https://tools.ietf.org/html/draft-ietf-mmusic-sdp-bundle-negotiation-52#section-7.3.3) says that if an m= section is tagged the answer cannot reject this m= section without rejecting all bundled m= sections.

It is possible to disable an offerer tagged m= section but that requires accepting it first and do another O/A.

Anyway this says that if we reject the offerer tagged m= section we reject all of them, like we have to. LGTM

PR #850, opened 2018-06-19
Stop using the "appdata" field in "a=msid".

### LINKED ISSUE #842: removeTrack(track) then addTrack(track) yields bogus SDP
Once a track has been negotiated, calling removeTrack on it does not remove the a=msid attribute from the SDP. A subsequent addTrack(track) will result in a new transceiver being created, resulting in an additional m-section with the same a=msid line. It is even possible for both m-sections to be negotiated to be sendrecv or sendonly if you give the old transceiver a track with replaceTrack.

taylor-b:
Fixes #842.

And also https://github.com/w3c/webrtc-pc/issues/1718.

Since introducing transceivers, replaceTrack, early media, etc., the
definition of a MediaStreamTrack has changed and track ID signaling has
become somewhat pointless. In many cases, "sender.track.id" on one side
will not match "receiver.track.id" on the other side; endpoints must use
MIDs or m= section indices or some other means to identify which track
is which.

Thus this PR just removes track ID signaling altogether and the
lingering problems it causes.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2093,19 +2093,8 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
                 <t>For each MediaStream that was associated with the
                 transceiver when it was created via addTrack or
                 addTransceiver, an "a=msid" line, as specified in
-                <xref target="I-D.ietf-mmusic-msid"></xref>, Section 2.
-                If a MediaStreamTrack is attached to the transceiver's
-                RtpSender, the "a=msid" lines MUST use that track's ID.
-                If no MediaStreamTrack is attached, a valid ID MUST be
-                generated, in the same way that the implementation
-                generates IDs for local tracks.</t>
-
-                <t>If no MediaStream is associated with the
-                transceiver, a single "a=msid" line with the special
-                value "-" in place of the MediaStream ID, as specified
-                in
-                <xref target="I-D.ietf-mmusic-msid"></xref>, Section 3.
-                The track ID MUST be selected as described above.</t>
+                <xref target="I-D.ietf-mmusic-msid"></xref>, Section 2,
+                but omitting the "appdata" field.</t>
               </list></t>
 
               <t>If the RtpTransceiver has a sendrecv or sendonly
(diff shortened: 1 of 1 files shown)


henbos:
Now that we only use the stream ID, "a=msid:{streamId}", I just want to verify that I understand this correctly:
- If I do addTrack(track) without any streams, "a=msid" is completely omitted.
- If I do addTrack(track, stream1, stream2, ...) with multiple streams, I get one "a=msid" line per stream.

Is this correct?

henbos:
Or is "a=msid" mandatory in which case I would have to do "a=msid:-" in the case of no stream?

henbos:
Before this change, "a=msid" was broken e.g. https://github.com/w3c/webrtc-pc/issues/1718 and https://github.com/w3c/webrtc-pc/issues/2005. The track ID is only surfaced to JavaScript if the answerer does not have any transceivers to match with the offer. Relying on "mid" or MediaStream IDs are clearly the better alternative, because they have a reliable behavior that works not just in edge-cases and are not misleading in all other cases.

An argument can be made for keeping track IDs, but I suspect the argument would be for backwards compatibility reasons or for people who want to parse SDP to figure out ID information (information which, by the way, would have been easier to carry through streams). In either case, JSEP and MSID specs [contradict themselves](https://github.com/w3c/webrtc-pc/issues/2005#issue-369203604).

As I recall from testing, while Chrome surfaces track IDs in some edge-cases, Firefox does not. If we can survive the switch to Unified Plan without track ID related breakages, as Firefox did, then I see little reason to try to save the sinking ship that is "a=msid". This is a remnant of the track/stream way of thinking.

To be discuss at TPAC. There is a slide for the issues referenced.

fippo:
a=msid is still relied on to create streams with synchronized tracks  -- LS groups haven't made it to any implementation afaik.

henbos:
Anyone who is sad to not be able to rely on MediaStreamTrack.id could though, because of "a=msid", do this :)
```
pc1.addTrack(track, new MediaStream(track.id));
...
pc2.ontrack = e => {
  remoteTrackIdsMap.set(e.track, e.streams[0].id);
}
```

juberti:
@henbos a=msid is not mandatory, and can be omitted if there are no streams.

alvestrand:
discoveries when updating -msid to allow for this:

- omitting a=msid will lead to a default stream (legacy behavior)
- the way to signal that a track is no longer present is by omitting msid after it has been present
- so the only way to signal "one track no stream" is to send a=msid:-
- -msid used to allow for sending one track into multiple recipient streams.With a=msid:- it can no longer do so.

PR is up. https://github.com/alvestrand/rtcweb-msid/pull/17

PR #851, opened 2018-10-12
Change references from RFC 5245 to RFC 8445 and I-D.ietf-mmusic-ice-sip-sdp.

### LINKED ISSUE #841: Wrong refs for end-of-candidates SDP
Agree, the references in Sections 5.2.2 and 5.3.2 should be updated accordingly. 

> On Wed, Oct 11, 2017 at 12:23 AM, Christer Holmberg <christer.holmberg@ericsson.com> wrote:
> Hi,
> 
> JSEP references draft-ietf-ice-trickle for the SDP ‘end-of-candidates’ attribute.
> 
> But, draft-ice-trickle only talks about a generic end-of-candidates indication. For SDP, the actual SDP attribute is defined in draft-ietf-mmusic-trickle-ice-sip.
>  Shouldn’t JSEP reference that spec instead?
>

### LINKED ISSUE #852: Update ICE reference to 8445

pthatcherg:
Fixes #841 
Fixes #852


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1932,6 +1934,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             be added, as specified in
             <xref target="I-D.ietf-ice-trickle"></xref>, Section 4.</t>
 
+            <t>An "a=ice-options" line with the "ice2" option MUST
```
juberti: ice2 should be combined into the previous paragraph that discusses ice-options.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2391,7 +2400,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             the most recent gathering phase (see
             <xref target="sec.ice-gather-overview"></xref>), an
             "a=candidate" line MUST be added, as defined in
-            <xref target="RFC5245"></xref>, Section 4.3., paragraph 3.
+            <xref target="8445"></xref>, Section 4.1.
```
juberti: This reference doesn't seem right - there is no S 4.1 in 8445.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2595,7 +2605,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
           identical to that indicated in the initial offers section
           above, except that the "a=ice-options" line, with the
           "trickle" option as specified in
-          <xref target="I-D.ietf-ice-trickle"></xref>, Section 4, is
+          <xref target="I-D.ietf-ice-trickle"></xref>, Section 4,
+          and the "ice2" option as specified in
+          <xref target="RFC8445"></xref>, Section 10, is
```
juberti: is->are

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3751,13 +3767,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>If the ICE ufrag or password changed from the previous
           remote description:
-          <xref target="RFC5245" />.
+          <xref target="RFC8445" />.
```
juberti: This line is a typo and should just be removed.

draft-ietf-rtcweb-jsep.xml
juberti: Seems like this should be a ref to sip-sdp

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3567,7 +3584,7 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
               <t>ICE ufrag and password values, which MUST comply with
               the size limits specified in
-              <xref target="RFC5245" />, Section 15.4.</t>
+              <xref target="RFC8445" />, Section 5.3.</t>
```
juberti: These limits don't seem clear enough - is the max length specified somewhere?

draft-ietf-rtcweb-jsep.xml
pthatcherg: I think the subject is 'the "a=ice-options" line', which is singular.

draft-ietf-rtcweb-jsep.xml
pthatcherg: Done

draft-ietf-rtcweb-jsep.xml
pthatcherg: Ah, it's in ietf-mmusic-ice-sip-sdp 4.4.  I've fixed the ref.

draft-ietf-rtcweb-jsep.xml
pthatcherg: Yep; fixed

draft-ietf-rtcweb-jsep.xml
pthatcherg: Whoops, duh. Fixed.

pthatcherg:
OK, I think I fixed everything

juberti:
@fluffy, that makes sense. That probably needs to be new text in Section 5.11, I think that could be separate from this PR (which doesn't talk about which ICE version is used at all).

fluffy:
I don't care if it is one PR to two and the rest of this PR looks good to me.

pthatcherg:
I added a note about RFC 5245 (based on Adam's email) and I added text like what Cullen was asking for around accepting offers and answers without "ice2".  I excluded the part of Adam's not about previous versions of trickle ICE because it didn't seem relevant to this PR and it had a missing ref and I didn't know what the ref should be (so I remove it).

TL;DR: I think this is closer to what we want eventually.  I'd suggest we land something and then tweak if further in a followup PR.  But I'm willing to delete or modify the note if that would get the PR landed.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -148,7 +148,17 @@ ipr="trust200902">
         startup, since ICE checks can start as soon as any transport
         information is available rather than waiting for all of it.
         JSEP's decoupling of the ICE and signaling state machines
-        allows it to accommodate either model.</t>
+        allows it to accommodate either model.
+
+        While this specification formally relies on <xref
+        target="RFC8445"></xref>, at the time of its publication, the
+        majority of WebRTC implementations support the version of ICE
+        described in <xref target="RFC5245"></xref>, the use of the
```
adamroach: This is a comma splice.
```suggestion
        described in <xref target="RFC5245"></xref>. The use of the
```

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3639,18 +3664,17 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
           m= section is marked bundle-only, or it is an answer and the
           m= section is bundled into into another m= section.)</t>
 
           <t>Or, if the ICE ufrag and password values have changed,
-          trigger the ICE agent to start an ICE restart, and begin
-          gathering new candidates for the m= section as described in
-          <xref target="RFC5245" />, Section 9.1.1.1. If this
+          trigger the ICE agent to start an ICE restart as described in
+          <xref target="RFC8445" />, Section 9, and begin
+          gathering new candidates for the m= section . If this
```
juberti: extra space

PR #857, opened 2018-11-08
Always use the JSEP proto fields. Fixes #854

### LINKED ISSUE #854: JSEP says that subsequent offers need to specify the ICE protocol
This contradicts the guidance in 5.1.2 Profile Names, which says impls MUST indicate
the UDP/TLS/RTP/SAVPF profile for each media m= line they produce in an offer.

Introduced in https://github.com/rtcweb-wg/jsep/commit/e8fd5c67e7fa09c6c0c171739449a7271de58fc7, which predates our decision to not use active ICE protocols in the m= line. 

Note that sip-sdp, S 3.3.4 still says that the active ICE protocol needs to be included.

ekr:
Fixes #854

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -2359,11 +2359,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             be generated according to the same rules as for an initial
             offer.</t>
 
-            <t>Each "m=" and c=" line MUST be filled in with the port,
-            protocol, and address of the default candidate for the m=
+            <t>Each "m=" and c=" line MUST be filled in with the port
+            and address of the default candidate for the m=
             section, as described in
             <xref target="I-D.ietf-mmusic-ice-sip-sdp"></xref>,
-            Section 3.2.1.2. If ICE
+            Section 3.2.1.2. The profile MUST be as specified
+            in <xref target="sec.profile-names"/>.
+            Note that
+            this differs from the guidance in
+            <xref target="I-D.ietf-mmusic-ice-sip-sdp" />, Section
+            3.3.4, which requires that the profile match the
+            in-use transport protocol.
+            If ICE
             checking has already completed for one or more candidate
             pairs and a candidate pair is in active use, then that pair
             MUST be used, even if ICE has not yet completed. Note that
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1799,16 +1799,16 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
 
           <t>For media m= sections, JSEP implementations MUST support
           the "UDP/TLS/RTP/SAVPF" profile specified in
-          <xref target="RFC5764"></xref>, and MUST indicate this
-          profile for each media m= line they produce in an offer. For
-          data m= sections, implementations MUST support the
-          "UDP/DTLS/SCTP" profile and MUST indicate this profile for
-          each data m= line they produce in an offer. Although these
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -2359,11 +2359,18 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             be generated according to the same rules as for an initial
             offer.</t>
 
-            <t>Each "m=" and c=" line MUST be filled in with the port,
-            protocol, and address of the default candidate for the m=
+            <t>Each "m=" and c=" line MUST be filled in with the port
+            and address of the default candidate for the m=
```
rshpount: This should be address and port of the default candidate if it matches the protocol in the m= line or IN IP4 0.0.0.0 and port 9, if protocol does not match the default candidate. Otherwise it adds an additional candidate or causes the mismatch.

juberti:
Being litigated on the list right now...

juberti:
Will sync with @seanturner next week and figure out how to put this to bed.

draft-ietf-rtcweb-jsep.xml
fluffy: So just to make sure I understand, let me make an example. Say the offer came from a server with a public IP and only does ice light and it offers an IP of 1.2.3.4 with port 2000 for TCP and 3000 for UDP. And lets say UDP is blocked and does not work. You are concerned that the m line in answer could end up IP and port mapped to the TCP flow, but the protocol would say UDP not TCP and that would be really weird? I'm just not following the concern here so if you could spell it out a bit more, that would be great. Thanks

rshpount: My concern here is that if server on public IP which implements *full* ICE sends an offer with IP of 1.2.3.4 with port 2000 for TCP and 3000 for UDP. Let's say UDP is blocked or disabled on the client, so server get's back an answer with only TCP candidate with IP of 2.3.4.5 and port 4000 but m and c lines would have UDP protocol with IP 2.3.4.5 and port 4000. According to ice-sip-sdp this will be treated as an additional candidate by the server, with server sending STUN message over UDP to IP 2.3.4.5 port 4000. With older ICE implementations this will be treated as ICE mismatch. Finally, this is just plain weird and misleading.

What I am suggesting is that if none of the candidates in the answer match the protocol in the offer, send an answer with protocol from the offer, IN IP4 0.0.0.0 and port 9. So, in our example client answer will have only a single TCP candidate, IN IP4 0.0.0.0 in the c= line and port 9 in the m line. The server, when it receives this answer will know that information is a place holder and will not treat it as an additional candidate or mismatch.

fluffy: Hmm - you could be right but can you point me at where in  ice-sip-sdp it takes something that arrives in the m= line and forms a new candidate out of it? That would seem wrong in that when doing ice, only candidates that are in candidate lines should be used. I get the m= line would indicate the default candidate and how that might cause some confusion here but I don't think ICE should change for it to form a new candidate as that seems like it will have bad backwards compatible interoperability problems.

rshpount: This is coming from ice-sip-sdp section 3.2.5 (https://tools.ietf.org/html/draft-ietf-mmusic-ice-sip-sdp-24#section-3.2.5).

The idea was that if firewall is modifying the address in the c= and m= line to an address that does not match any of the candidates, it is a better to run a connectivity check against the firewall provided address together with all the ICE candidates vs declaring the mismatch,  ignoring all the ICE candidates and only using the address in c= and m= line (as specified in RFC5245).

juberti:
Closing in favor of #863

PR #858, opened 2018-11-08
Fix sess-id text

### LINKED ISSUE #855: JSEP restriction on session ID is 2^62, should be 2^63 or 2^64
JSEP -24 currently says:

      The sess-id MUST be representable by a 64-bit signed
      integer, and the initial value MUST be less than (2**62)-1, as
      required by [RFC3264].  It is RECOMMENDED that the sess-id be
      constructed by generating a 64-bit quantity with the two highest
      bits being set to zero and the remaining 62 bits being
      cryptographically random.

There are two problems here:

- RFC 3264 actually doesn't place that restriction on the sess-id, that's the restriction on the version field.

- Deployed code (libwebrtc) generates sess-ids that are larger than 2^62-1

Quote from RFC 3264:

   The offer (and answer) MUST be a valid SDP message, as defined by RFC
   2327 [1], with one exception.  RFC 2327 mandates that either an e or
   a p line is present in the SDP message.  This specification relaxes
   that constraint; an SDP formulated for an offer/answer application
   MAY omit both the e and p lines.  The numeric value of the session id
   and version in the o line MUST be representable with a 64 bit signed
   integer.  The initial value of the version MUST be less than
   (2**62)-1, to avoid rollovers.  Although the SDP specification allows
   for multiple session descriptions to be concatenated together into a
   large SDP message, an SDP message used in the offer/answer model MUST
   contain exactly one session description.

So what's easiest to change, deployed code or JSEP?

juberti:
Fixes #855
Also add ip-handling reference

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1906,15 +1906,15 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <xref target="RFC4566"></xref>, Section 5.2. The value of
             the &lt;username&gt; field SHOULD be "-". The sess-id MUST
             be representable by a 64-bit signed integer, and the
-            initial value MUST be less than (2**62)-1, as required by
-            <xref target="RFC3264" />. It is RECOMMENDED that the
+            value MUST be less than (2**63)-1. It is RECOMMENDED that the
             sess-id be constructed by generating a 64-bit quantity with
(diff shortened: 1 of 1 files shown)


juberti:
ping @fluffy, @ekr

seanturner:
@fluffy @ekr PTAL - Barring any objections this one ought to get merged.

juberti:
Also think this is uncontroversial, planning to land this next week if there are no objections.

PR #859, opened 2018-11-08
Add text about default msid

### LINKED ISSUE #856: backward compability with endpoint that do not send a=msid and only a single audio/video stream
Take this simplified SDP offer from a remote endpoint (in parts taken from [here](https://tools.ietf.org/html/draft-ietf-rtcweb-sdp-11#section-5.2.2.1):
```
v=0
o=- 20518 0 IN IP4 0.0.0.0
s=-
t=0 0
m=audio 9 UDP/TLS/RTP/SAVPF 109
c=IN IP4 203.0.113.141
a=sendrecv
a=rtpmap:109 opus/48000/2
a=ice-ufrag:074c6550
a=ice-pwd:a28a397a4c3f31747d1ee3474af08a068
a=fingerprint:sha-256 19:E2:1C:3B:4B:9F:81:E6:B8:5C:F4:A5:A8:D8:73:04:BB:05:2F:70:9F:04:A9:0E:05:E9:26:33:E8:70:88:A2
a=setup:actpass
m=video 9 UDP/TLS/RTP/SAVPF 99
c=IN IP4 203.0.113.141
a=sendrecv
a=rtpmap:99 H264/90000
a=fmtp:99 profile-level-id=4d0028;packetization-mode=1
a=ice-ufrag:074c6550
a=ice-pwd:a28a397a4c3f31747d1ee3474af08a068
a=fingerprint:sha-256 19:E2:1C:3B:4B:9F:81:E6:B8:5C:F4:A5:A8:D8:73:04:BB:05:2F:70:9F:04:A9:0E:05:E9:26:33:E8:70:88:A2
a=setup:actpass
```
A single audio and video stream. No a=mid (there is quite some text taking care of that) but no a=msid line either. https://jsfiddle.net/5jd8Lz97/1/ shows a live example.

Both Chrome (without unified-plan) and Firefox will fire the ontrack event (or legacy onaddstream) with a single mediastream containing an audio and video track. This has been like that since... 2012-ish?

Unified plan support in Chrome changes this behaviour: https://jsfiddle.net/5jd8Lz97/
ontrack is fired with an empty set of media streams. onaddstream is not fired.

This is quite a change in behaviour. And unexpected, I didn't expect any issues for this trivial case of a single audio/video track.

As @henbos [pointed out](https://bugs.chromium.org/p/chromium/issues/detail?id=901711) webrtc-pc assumes/requires a=msid to be present[here](https://w3c.github.io/webrtc-pc/#processing-remote-mediastreamtracks) and [here](https://w3c.github.io/webrtc-pc/#set-the-rtcsessiondescription). In particular this says
```
If direction is "sendrecv" or "recvonly", let msids be a list of the MSIDs that the media 
description indicates transceiver.[[Receiver]].[[ReceiverTrack]] is to be associated with.
Otherwise, let msids be an empty list.
```
which means firing with no streams is technically correct.

JSEP seems to assume it is present as well, but has this text about a=msid in [5.8.2](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-24#section-5.8.2)
```
 If present, "a=msid" attributes MUST be parsed as specified in
      [I-D.ietf-mmusic-msid], Section 3.2, and their values stored.
```
Note the "If present". What if it is not present?

The closest I could find in terms of guidance was this from the [msid draft](https://tools.ietf.org/html/draft-ietf-mmusic-msid-16)
```
   It follows from the above that MediaStreamTracks in the "default"
   MediaStream cannot be closed by removing the msid attribute;
```
and its section about handling [non-signalled tracks](https://tools.ietf.org/html/draft-ietf-mmusic-msid-16#section-3.1)

a=msid is supposed to be present in initial offer so we are talking about dealing with non-webrtc entities send an offer. Given that there is plenty of support for entities that do not support a=mid (an RFC published in 2010) in JSEP do we need backward compability for the case of a single audio/video track each but no msid?

Interop between browsers is probably not affected, due to Chrome still interpreting and sending a=ssrc:123 msid:...

juberti:
Fixed #856

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3519,7 +3519,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>If present, "a=msid" attributes MUST be parsed as
             specified in
             <xref target="I-D.ietf-mmusic-msid" />, Section 3.2, and
-            their values stored, ignoring any "appdata" field.</t>
+            their values stored, ignoring any "appdata" field. If no "a=msid"
+            attributes are present, a "default" msid-id value is created for
+            the session, if one hasn't already been created, and this "default"
+            value is stored.</t>
 
             <t>Any "a=imageattr" attributes MUST be parsed as specified
             in
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -3520,9 +3520,9 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             specified in
             <xref target="I-D.ietf-mmusic-msid" />, Section 3.2, and
             their values stored, ignoring any "appdata" field. If no "a=msid"
-            attributes are present, a "default" msid-id value is created for
-            the session, if one hasn't already been created, and this "default"
-            value is stored.</t>
+            attributes are present, a random msid-id value is generated for a
+            "default" MediaStream for the session, if not already present, and
+            this value is stored.</t>
 
             <t>Any "a=imageattr" attributes MUST be parsed as specified
             in


fippo:
also tagging @jan-ivar (who might be unhappy with 'default' which matches Chrome behaviour but not Firefox behaviour (it generates a uuid)

draft-ietf-rtcweb-jsep.xml
```diff
@@ -3519,7 +3519,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             <t>If present, "a=msid" attributes MUST be parsed as
             specified in
             <xref target="I-D.ietf-mmusic-msid" />, Section 3.2, and
-            their values stored, ignoring any "appdata" field.</t>
+            their values stored, ignoring any "appdata" field. If no "a=msid"
+            attributes are present, a "default" msid-id value is created for
```
jan-ivar: By "default" do you literally mean the` id === "default"`, or figuratively some value chosen by the UA? Why would it need to be a specific string?

Wouldn't that limit UAs from using different ids for independent peer connections, either causing msid-less tracks from *all* peer connections on the page to end up in the same stream, or different streams with identical ids?

juberti: When I used quotes around "default", the intention was the same as in https://tools.ietf.org/html/draft-ietf-mmusic-msid-16, Section 3.1; there's no official *default* MediaStream but we create a catch-all. However, I see that this could be interpreted the way you describe. Will reword.

juberti: PTAL

juberti:
ping @fluffy, @ekr

seanturner:
@fluffy @ekr PTAL - Barring any objections this one ought to get merged.

juberti:
Folks, I think this one is pretty uncontroversial - planning to land this next week if there are no objections.

PR #863, opened 2019-01-28
ICE reconciliation

juberti:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3013,10 +3015,10 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
...
             described in
             <xref target="I-D.ietf-mmusic-ice-sip-sdp"></xref>,
-            Section 3.2.1.2. Note, however,
-            that the m= line protocol need not match the default
-            candidate, because this protocol value must instead match
-            what was supplied in the offer, as described above.</t>
+            Section 3.2.1.2. Note that in certain cases, the m= line protocol
+            may not match that of the default candidate, because the m= line
+            protocol value MUST match what was supplied in the offer, as
+            described above.</t>
```
rshpount: Should we say here what address and port should be used in this case in c= and m= line? I would say that dummy values should be used for address and port.

juberti: As noted, it has to be the IP/port of the default candidate.

juberti: dummy values are only used if no candidates have yet been gathered.

rshpount:
@juberti What happens in the following situation?
1. Browser client is placed behind the firewall which only allows TCP connections 
2. No TURN servers are configured
3. Client establishes connection to a server which supports ICE TCP
4. ICE nomination completes
5. Client browser app calls createOffer with no options (ICE restart option is not present)

In this case ICE nomination is complete and the TCP candidate is nominated. No other candidates should be present in the offer.

juberti:
I believe the other candidates can still be present in the updated offer, as noted in the JSEP examples section.

rshpount:
@juberti What example specifically are you talking about?

Also, please take a look at RFC 5245 section 9.1.2.2 (https://tools.ietf.org/html/rfc5245#section-9.1.2.2):

The agent MUST include candidate attributes for candidates matching the default destination for each component of the media stream, and MUST NOT include any other candidates.

And in mmusic-ice-sip-sdp Section 3.4.1.2.2 (https://tools.ietf.org/html/draft-ietf-mmusic-ice-sip-sdp-24#section-3.4.1.2.2):

In addition, the offerer only includes SDP candidates representing the local candidates of the nominated candidate pair.  The offerer MUST NOT include any other SDP candidate attributes in the subsequent offer.

Does JSEP overwrites this somewhere?

juberti:
example: https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-25#section-7.2

JSEP seems to say that a=candidate lines MUST always be added:
      If the m= section is not bundled into another m= section, for each
      candidate that has been gathered during the most recent gathering
      phase (see Section 3.5.1), an "a=candidate" line MUST be added, as
      defined in [I-D.ietf-mmusic-ice-sip-sdp], Section 4.1.  If
      candidate gathering for the section has completed, an "a=end-of-
      candidates" attribute MUST be added, as described in
      [I-D.ietf-ice-trickle], Section 9.3.  If the m= section is bundled
      into another m= section, both "a=candidate" and "a=end-of-
      candidates" MUST be omitted.

rshpount:
I guess it was not immediately obvious to me if this section referrers to offers sent while ICE nomination process or when it is completed. RFC5245 and ice-sip-sdp both make a distinction between the two. The ICE nomination status in the example is also not clear.

In any case, if overwriting a MUST from a referenced RFC was intentional, it should be explicit. I think I am not the only person who is confused by this right now.

Also, what is the reason to keep old candidates once ICE nomination is completed? When does the browser release the unused candidates?

juberti:
It wasn't intentional; it sort of assumed that renomination could occur at any time (which was my view on ICE at the time this was written), which meant you would want the additional candidates

rshpount:
Since this was not intentional, non-obvious, and somewhat contradicts RFC5245/RFC8445/ice-sip-sdp  specifications, what do you want to do here?

One way to solve this is to make this compliant with RFC8445/ice-sip-sdp, and to specify that once ICE nomination completes, only nominated candidates are kept and implementation can release other candidates. Subsequent offer after the end of ICE nomination includes only nominated candidates. If TCP candidate is a default in this case, proto should be TCP based.

Another way to deal with this is to specify how JSEP nomination process is different from normal ICE. For instance ICE nomination from the browser is never considered complete, so that candidates are never released and all candidates are always included in the offer. In this case, m= and c= line can be assigned to one of the UDP candidates even if TCP candidate is the default. This way proto is always UDP based and browser based client will never generate TCP based proto.

Current specification is kind of between these two points which makes it ambiguous.

draft-ietf-rtcweb-jsep.xml
cdh4u: What is the reason for the "candidate in active use" deviation? If ICE checking has completed for multiple pairs, the endpoints can switch between those pairs. Which pair is the "active"?

Also, we should use proper ICE terminology. Instead of saying "If ICE checking has completed for multiple pairs" we should say "If ICE has produced one ore more valid pairs", or something like that...

juberti: @cdh4u, not sure which text you are referring to here.

cdh4u: The text on row 2368.

juberti: ah, ok. The goal here is to avoid any ambiguity, I believe. It would be the most-recently-used pair for sending media.

cdh4u:
> My guess is that the consensus view for JSEP would be that it match 8445, and not include extra candidates. Chrome supports renomination, and would include the extra candidates accordingly, but that's not something that JSEP needs to adhere to (especially because there is no active standards-track renomination spec).

Just for my own interest, does Chrome send keep-alives and consensus checks for all candidates that it includes in subsequent offers? If not, how does Chrome check that a pair still works if it wants to renominate?

draft-ietf-rtcweb-jsep.xml
cdh4u: > ah, ok. The goal here is to avoid any ambiguity, I believe. It would be the most-recently-used pair for sending media.

I think that should be clarified. Let's not introduce new ICE terminology - is it messy enough already :) 

But, before nomination takes place, endpoints can send media on ANY valid pair, and the pair used for sending media can change. I think the text should say *WHY* JSEP deviates from draft-ice-sdp in this regard.

rshpount: One of the reasons why ice-sip-sdp specified default and not the "most-recently-used pair", was that most-recently used pair can change between offer and answer. Furthermore, candidates used by different end points during ICE nomination can belong to different candidate pairs and even used different transport (for instance if one of UDP connectivity check packets got lost due to packet loss). This results in candidates on the m= line and even protocols which do not match. Stating that default candidate MUST be one of the UDP candidates, makes c=/m= line change less often and be more consistent. Also, changing c=/m= line this often serves little purpose when both end points support ICE (as they should for JSEP).

Regarding the protocol and candidate mismatch in the answer, this can result in ICE mismatch and negotiation failure. See  https://tools.ietf.org/html/draft-ietf-mmusic-ice-sip-sdp-24#section-3.3.3 for more details. Once again, putting dummy value is safer there (but can still cause issues with older implementations).

cdh4u: > One of the reasons why ice-sip-sdp specified default and not the "most-recently-used pair", was that most-recently used pair can change between offer and answer. Furthermore, candidates used by different end points during ICE nomination can belong to different candidate pairs and even used different transport (for instance if one of UDP connectivity check packets got lost due to packet loss). This results in candidates on the m= line and even protocols which do not match. Stating that default candidate MUST be one of the UDP candidates, makes c=/m= line change less often and be more consistent. Also, changing c=/m= line this often serves little purpose when both end points support ICE (as they should for JSEP).
> 
> Regarding the protocol and candidate mismatch in the answer, this can result in ICE mismatch and negotiation failure. See https://tools.ietf.org/html/draft-ietf-mmusic-ice-sip-sdp-24#section-3.3.3 for more details. Once again, putting dummy value is safer there (but can still cause issues with older implementations).

@rshpount, I agree with you, and that's why I would like to know why JSEP does things differently.

juberti:
Updated this PR to incorporate the feedback regarding what candidate should be put into the m=/c= during ICE negotiation (now JSEP says to simply follow sip-sdp).

draft-ietf-rtcweb-jsep.xml
```diff
@@ -2359,20 +2359,14 @@ candidate:1 1 UDP 1694498815 192.0.2.33 10000 typ host
             be generated according to the same rules as for an initial
             offer.</t>
 
-            <t>Each "m=" and c=" line MUST be filled in with the port,
-            protocol, and address of the default candidate for the m=
+            <t>Each "m=" and c=" line MUST be filled in with the port, profile,
```
rshpount: I've just noticed that this document refers to SDP protocol as profile. If you look at the references for the documents that define "profiles" in JSEP, they all refer to them as protocols. For SRTP profile is actually something different and refers to SRTP key parameters.

Also, default candidate does not have profile (or transport), it has transport, port, and (connection-)address. So each "m=" and "c=" MUST be filled in with port, protocol, and address to match port, transport, and address of the default candidate for the m= section, as described in ice-sip.sdp.

juberti: It's kind of ambiguous; https://tools.ietf.org/html/rfc7850 uses "profile", "proto", and "protocol". 

Overall this change would extend outside of this specific block of text; as such, I think you should raise it on the mailing list and we can consider it outside of the scope of this review.

rshpount: I will take the first comment to the list.

The second comment still applies: Default candidate does not have a profile (or protocol) as in UDP/TLS/SRTP/AVPF. It has transport (UDP or TCP), port and address.  What we are trying to say is that profile in m= line must use the same transport as default candidate, but we cannot say that m= line must use the profile from default candidate, since no profile is specified in the candidate.

juberti: OK, that makes sense, will adjust accordingly.

juberti: I think I addressed the issue. PTAL.

rshpount: This looks good.

PR #944, opened 2020-08-12
Rfced 12 pranswer

ajeanmahoney:
used suggestion 2, deleted comments, fixes #882, issue 12

PR #945, opened 2020-08-12
Rfced 13 handshaking

### LINKED ISSUE #883: 13) [rfced] Section 4.1.8.1:  As it appears that "ICE and DTLS
13) [rfced] Section 4.1.8.1:  As it appears that "ICE and DTLS
handshaking have" means "ICE handshaking and DTLS handshaking have,"
we updated this sentence accordingly.  Please let us know if this is
incorrect (i.e., if the text refers to one handshaking process, in
which case "have" should be "has").

Original:
 Upon acceptance by the user, the
 application will plug in the tracks it has acquired, which, because
 ICE and DTLS handshaking have likely completed by this point, can
 start transmitting immediately.

Currently:
 Upon acceptance by the user, the
 application will plug in the tracks it has acquired, which, because
 ICE handshaking and DTLS handshaking have likely completed by this
 point, can start transmitting immediately.

ajeanmahoney:
authors ok with change, deleted comment, fixes #883, issue 13


juberti:
LGTM once #944 changes are addressed

PR #946, opened 2020-08-12
Rfced 15 16 17 18 sec4nits

### LINKED ISSUE #885: 15) [rfced] Section 4.1.10:  Please confirm that "local media
15) [rfced] Section 4.1.10:  Please confirm that "local media
state" and "local resources" (as opposed to remote) are correct in the
context of setRemoteDescription.  (We ask because we see identical
wording in Section 4.1.9 ("setLocalDescription").)

Original:
 This API changes the local media state; among other things, it sets
 up local resources for sending and encoding media.

### LINKED ISSUE #886: 16) [rfced] Section 4.1.16:  Should "The effects of this method
16) [rfced] Section 4.1.16:  Should "The effects of this method
call" be "The effects of calling this method," and should "This call"
be "Calling this method"?

Original:
 The effects of this method call
 depend on when it is invoked, and differ depending on which specific
 parameters are changed:
...
 This call may result in a change to the state of the ICE Agent.

### LINKED ISSUE #887: 17) [rfced] Section 4.2.2:  Will "stopTransceiver" be clear to
17) [rfced] Section 4.2.2:  Will "stopTransceiver" be clear to
readers?  We could not find this string elsewhere in this document,
anywhere else in this cluster of documents, in any published RFC, or
in google searches.

Original:
 The stopped property indicates whether the transceiver has been
 stopped, either by a call to stopTransceiver or by applying an answer
 that rejects the associated m= section.

### LINKED ISSUE #888: 18) [rfced] Section 4.2.5:  We see "direction attribute" but not
18) [rfced] Section 4.2.5:  We see "direction attribute" but not
"directional attribute" in RFC 3264.  Will this text be clear to
readers?  (We ask because we also see "A direction attribute,
determined by applying the rules regarding the offered direction
specified in [RFC3264], Section 6.1" in Section 5.3.1 of this
document.)

Original:
 More specifically, it
 indicates the [RFC3264] directional attribute of the associated m=
 section in the last applied answer (including provisional answers),
 with "send" and "recv" directions reversed if it was a remote
 answer.

ajeanmahoney:
Several small updates:

- Issue 15: local media state, local resources correct, removed comment. Fixes #885
- Issue 16: updated wording to use 'calling this method', removed comments, fixes #886
- Issue 17: changed stopTransceiver to stop, removed comments, fixes #887
- Issue 18: changed 'directional attribute' to 'direction attribute', removed comments, fixes #888

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1570,22 +1570,9 @@ Possibly:
           <t>The setConfiguration method allows the global
           configuration of the PeerConnection, which was initially set
           by constructor parameters, to be changed during the session.
-          The effects of this method call depend on when it is invoked,
+          The effects of calling this method depend on when it is invoked,
           and they will differ, depending on which specific parameters are
-          changed:
-
-<!-- [rfced] Section 4.1.16:  Should "The effects of this method call"
-be "The effects of calling this method," and should "This call" be
-"Calling this method"?
-
-Original:
- The effects of this method call
- depend on when it is invoked, and differ depending on which specific
- parameters are changed:
-...
- This call may result in a change to the state of the ICE Agent. -->
-
-</t>
+          changed: </t>
           <ul spacing="normal">
             <li>Any changes to the STUN/TURN servers to use affect the
               next gathering phase. If an ICE gathering phase has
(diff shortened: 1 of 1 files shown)


juberti:
LGTM once #944 lands

Issue #893, opened 2020-07-26
23) [rfced] Section 5.2.1:  This is the only bullet item in this


juberti:
23) [rfced] Section 5.2.1:  This is the only bullet item in this
list that (1) is a sentence fragment, unlike the rest and (2) does
not contain "RFC 2119" language.  We have updated as follows, but
please let us know if further changes are needed.

Original:
 o  If WebRTC identity is being used, an "a=identity" line as
    described in [I-D.ietf-rtcweb-security-arch], Section 5.

Currently:
 *  If WebRTC identity is being used, an "a=identity" line, as
    described in [RFC8827], Section 5, needs to be included.

juberti:
I think what we want is

 `an "a=identity" line MUST be added, as described in [RFC8827], Section 5.'

juberti:
@ajeanmahoney please close this issue once you consider it resolved.

ajeanmahoney:
@mskucherawy normative text change, please review.

ajeanmahoney:
Removed editor-ready because it's already been addressed and added the discuss label since that seems like a reasonable label for AD attention.

juberti:
ping @mskucherawy

mskucherawy:
Yep, that's approved.

PR #952, opened 2020-08-19
Rfced 19 21 23 30 wording

### LINKED ISSUE #889: 19) [rfced] Section 5.1.1:  Does SDES refer to "source
19) [rfced] Section 5.1.1:  Does SDES refer to "source
description" or "security description"?  Neither [RFC4568] nor RFC
8827 [I-D.ietf-rtcweb-security-arch] mention "SDES" or "source
description" (as used in RFC 8852 and other documents in this
cluster). 

Original:
 The SDES SRTP keying mechanism from [RFC4568] MUST NOT be used, as
 discussed in [I-D.ietf-rtcweb-security-arch].

### LINKED ISSUE #891: 21) [rfced] Section 5.2.1:  Equations and other math-related
21) [rfced] Section 5.2.1:  Equations and other math-related
items: Per <https://www.rfc-editor.org/materials/FAQ-xml2rfcv3.html>,
xml2rfc v3 provides the ability to use superscript (among other
features previously not available).  Please note that superscript
will display in the .html and .pdf files but not in the .txt file
(where it appears as "(2^(63))-1").  We added the "superscript"
version of "(2**63)-1" in parentheses here, to show you how it would
appear.  Please let us know if you would like to use the superscript
feature.

Also, if there are other items in this document that could use
similar treatment, please let us know.

Original:
 The sess-id MUST be representable by a 64-bit signed
 integer, and the value MUST be less than (2**63)-1.

Currently:
  .xml (to be updated per your preference):
    ... less than (2**63)-1 ((2<sup>63</sup>)-1).

### LINKED ISSUE #900: 30) [rfced] Section 5.2.1:  We had trouble sorting out the "and"
30) [rfced] Section 5.2.1:  We had trouble sorting out the "and"
... "or" relationships here.  If the suggested text is not correct,
please clarify.

Original:
 o  If the bundle policy for this PeerConnection is set to "max-
    bundle", and this is not the first m= section, or the bundle
    policy is set to "balanced", and this is not the first m= section
    for this media type, an "a=bundle-only" line.

Suggested:
 o  If (1) the bundle policy for this PeerConnection is set to "max-
    bundle" and this is not the first "m=" section or (2) the bundle
    policy is set to "balanced" and this is not the first "m=" section
    for this media type, an "a=bundle-only" line.

ajeanmahoney:
Update contains the following:

- clarified SDP security descriptions mechanism. Removed comment. Fixes #889, issue 19
- added superscript. Removed comment. Fixes #891, issue 21
- updated to use normative text. Removed comment. Fixes issue #893, issue 23
- authors ok with suggested wording. Removed comment. Fixes #900, issue 30


draft-ietf-rtcweb-jsep.xml
```diff
@@ -1901,31 +1890,9 @@ Examples from original:
...
- The sess-id MUST be representable by a 64-bit signed
- integer, and the value MUST be less than (2**63)-1.
-
-Currently:
-  .xml (to be updated per your preference):
-    ... less than (2**63)-1 ((2<sup>63</sup>)-1). -->
-
- It is <bcp14>RECOMMENDED</bcp14> that the
+           value <bcp14>MUST</bcp14> be less than
+	    (2<sup>63</sup>)-1.
```
juberti, line 1894: Can you remove the parentheses here?

ajeanmahoney: Parens removed, PR updated

PR #955, opened 2020-08-24
Rfced 38 43 47 49 wording

### LINKED ISSUE #908: 38) [rfced] Section 5.3.1:  Does this text mean that the offered
38) [rfced] Section 5.3.1:  Does this text mean that the offered
media formats are allowed (in which case it should say "they are
allowed") or are not allowed (in which case "and, if applicable"
should be "or, if applicable")?

Original:
 o  None of the offered media formats are supported and, if
    applicable, allowed by codec preferences.

### LINKED ISSUE #913: 43) [rfced] Section 5.8.1:  We found this sentence confusing;
43) [rfced] Section 5.8.1:  We found this sentence confusing;
should "t=" remain in this list of lines not used by this
specification?

Original:
 The "i=", "u=", "e=", "p=", "t=", "r=", "z=", and "k=" lines are
 not used by this specification; they MUST be checked for syntax
 but their values are not used.

We ask because we see (for example):

o  A "t=" line MUST be added, as specified in [RFC4566], Section 5.9;
    both <start-time> and <stop-time> SHOULD be set to zero, e.g. "t=0
    0".
...
 o  The "s=" and "t=" lines MUST stay the same.
...
 t=0 0

### LINKED ISSUE #917: 47) [rfced] Section 5.10:  We found this sentence confusing, as
47) [rfced] Section 5.10:  We found this sentence confusing, as
we could not tell what "as prescribed by [RFC8445], Section 9" and
"generate an error" refer to.  Please confirm that the citation is
correct and will be clear to readers.

Original:
 o  If the current local description indicates an ICE restart, and
    either the ICE ufrag or password has not changed from the previous
    remote description, as prescribed by [RFC8445], Section 9,
    generate an error.

### LINKED ISSUE #919: 49) [rfced] Section 5.11:  As it appears that "is also enabled"
49) [rfced] Section 5.11:  As it appears that "is also enabled"
refers to silence suppression, we updated this sentence as follows.
Please let us know if this is incorrect.

Original:
 *  If the m= section is of type audio, and silence suppression was
    configured for the send media format as a result of processing
    the remote description, and is also enabled for that format in
    the local description, use silence suppression for outgoing
    media, in accordance with the guidance in Section 5.2.3.2.

Currently:
 -  If (1) the "m=" section is of type "audio" and (2) silence
    suppression was configured for the send media format as a
    result of processing the remote description and is also enabled
    for that format in the local description, use silence
    suppression for outgoing media, in accordance with the guidance
    in Section 5.2.3.2.

ajeanmahoney:
Updated with the following:

- clarified offered media formats, removed comment. Fixes #908, issue 38
- dropped the phrase 'not used by this spec...', removed comment. Fixes #913, issue 43
- clarified error condition, removed comment. Fixes #917, issue 47
- clarified silence suppression, removed comment. Fixes #919, issue 49


draft-ietf-rtcweb-jsep.xml
```diff
@@ -3953,34 +3909,14 @@ Original:
...
 	       the outgoing RTP streams, as indicated in
 	       <xref target="RFC8851" sectionFormat="comma" section="4"/>.</li>
-		 <li>If (1) the "m=" section is of type "audio" and (2)&nbsp;silence
-	       suppression was configured for the send media format as a
-	       result of processing the remote description and is also
-	       enabled for that format in the local description, use
+		 <li>If the "m=" section is of type "audio", and silence
+	       suppression was (1) configured for the send media format as a
+	       result of processing the remote description and (2) also
+	       enabled for that format in the local description, then use
```
juberti, line 3915: suggest dropping the 'then', as it's not used in the other bullets.

ajeanmahoney: 'then' removed.

Issue #901, opened 2020-07-26
31) [rfced] Sections 5.2.1 and 5.3.1:  Please confirm that


juberti:
31) [rfced] Sections 5.2.1 and 5.3.1:  Please confirm that
"9 IN IP4 0.0.0.0" (and not "IN IP4 0.0.0.0") is correct in these
two items.

Original:
 o  An "a=rtcp" line, as specified in [RFC3605], Section 2.1,
    containing the dummy value "9 IN IP4 0.0.0.0", because no
    candidates have yet been gathered.
...
 Otherwise, an "a=rtcp" line, as
 specified in [RFC3605], Section 2.1, containing the dummy value "9
 IN IP4 0.0.0.0" (because no candidates have yet been gathered).

juberti:
This is the correct value; the complete line would be `a=rtcp:9 9 IN IP4 0.0.0.0`.

juberti:
@ajeanmahoney

ajeanmahoney:
I don't think it needs further clarification, but perhaps using the term "default" (used in RFC8840) or "placeholder" instead of "dummy"? (9 instances)

juberti:
'default', to match RFC8840, SGTM. 

Note that the trickle-ICE reference to Section 5.1 (in JSEP, S 5.3.1) looks like it needs to be updated:
```as indicated in [I-D.ietf-ice-trickle], Section 5.1.```

ajeanmahoney:
[I-D.ietf-ice-trickle] is now RFC 8838.  I'm not finding a [RFC8838], Section 5.1 reference. Perhaps another Section 5.1 reference?

juberti:
ok, I see that you've already updated the reference properly - I was looking at jsep-26. Switching to ```default``` should work fine everywhere then.

PR #956, opened 2020-08-24
Rfced 31 dummy value

ajeanmahoney:
replaced the term 'dummy' with 'default' per comments on #901, issue 31


juberti:
Since this also contains the changes from #955, will wait until that PR is merged.

PR #989, opened 2020-12-12
update W3C webertc ref

fluffy:


fluffy:
@ajeanmahoney - One of the last names in this references is Boström. I have no idea how that works with unicode / ascii etc. Can you check the characters of that name are correct. Thanks

fluffy:
This PR is wrong - don't merge

fluffy:
This has been replaced by https://github.com/rtcweb-wg/jsep/pull/991

PR #991, opened 2020-12-13
fix up W3C ref on rfced-61 version 

fluffy:


fluffy:
FYI - we should get the Proposed Recommendation version this today and I will update and all going well the Recommendation version ( semi equivalent to an RFC) should be Jan 26

fluffy:
The PR version published and I updated to point at that

juberti (approved):

ajeanmahoney:
Regarding the name Boström, this can be included as-is in the XML:
`<author fullname="Henrik Boström" initials="H." surname="Boström" role="editor">`

ajeanmahoney:
@fluffy I fixed Boström in PR #999

PR #999, opened 2021-01-11
updated dates and C238 refs, updated Boström in the [W3C.webrtc] ref.…

### LINKED ISSUE #998: Updates to C238 references
The references for the C238 docs need to be checked and updated - author names and titles may have been updated, and the dates definitely need updating.

ajeanmahoney:
updated dates and C238 refs (dates and order of seriesInfo), updated Boström in the [W3C.webrtc] ref (see #991). Fixes #998

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -55,7 +55,7 @@ affiliation in various documents within the cluster.  Please review and let us
 know if any updates are needed. 
 -->
 
-    <date month="July" year="2020"/>
+    <date month="January" year="2021"/>
 
     <keyword>webrtc</keyword>
     <keyword>sdp</keyword>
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -5103,10 +5103,10 @@ Datagram Transport Layer Security (DTLS) and Transport Layer Security (TLS)</tit
               <organization>Cisco</organization>
             </author>
             <author
-                fullname="Henrik Bostr&#x00F6;m"
```
fluffy, line 5106: I can live with this either way but the problem with this change it is very hard to ensure that the character you replaced it with is the the right one. Or some smart tool did not change it along the way in the way that microsoft word replace some characters. My read of the RFCs was that the #x00F6 form was prefered but I could be very wrong about that. 

This seems like something there should be clear guidance on. If you know what is the approved of way, I am good with either was I don't know what we are supposed to do this this. I do know that I can't look at this diff in this tool and tell that it was replaced with the correct character or just one that looks similar.

ajeanmahoney: The current RPC practice is to insert UTF-8 into attributes such as fullname.  See the following for more details:
https://www.rfc-editor.org/materials/FAQ-xml2rfcv3.html#name-how-do-i-insert-non-ascii-c

FWIW  I took the output of xml2rfc, which rendered #x00F6 to ö, as input for this change.

juberti (approved):

draft-ietf-rtcweb-jsep.xml
fluffy: OK - that is nuts but fine with me if that is what people are doing. Thanks

PR #1002, opened 2021-01-14
Rfced 210114

### LINKED ISSUE #848: canTrickleIceCandidiates when sending initial offer
JSEP-24 says:

        "null:  No SDP has been received from the other side, so it is not
         known if it can handle trickle.  This is the initial value before
         setRemoteDescription() is called.²

         (The same is said in the W3C WebRTC spec)

         ...and:

        "However, applications can use the canTrickleIceCandidates property to determine whether their  peer can actually do Trickle ICE, i.e., whether it is safe to send an initial offer or answer followed later by candidates as they are gathered. As "true" is the only value that definitively indicates remote Trickle ICE support, an application which compares canTrickleIceCandidates against "true" will by default attempt Half Trickle on initial offers and Full Trickle on subsequent interactions with a Trickle ICE-compatible agent."

QUESTION: Since the value will be NULL before completion of setRemoteDescription (before SDP has been received from the other side) how can the value be used by applications for initial offers?

SUGGESTION: Remove the text saying that canTrickleIceCandidates can be used when creating an initial offer.

### LINKED ISSUE #873: 3) [rfced] Eric, we note that Mozilla and RTFM are listed as your
3) [rfced] Eric, we note that Mozilla and RTFM are listed as your
affiliation in various documents within the cluster.  Please review
and let us know if any updates are needed.

ajeanmahoney:
Ekr ok with his contact info. Removed comment. Fixes #873
added parenthetical, fixes #848

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -50,11 +50,6 @@
       </address>
     </author>
 
-<!-- [rfced] Eric, we note that Mozilla and RTFM are listed as your
-affiliation in various documents within the cluster.  Please review and let us
-know if any updates are needed. 
--->
-
     <date month="January" year="2021"/>
 
     <keyword>webrtc</keyword>
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1483,7 +1483,7 @@ Suggested:
           gathered. As "true" is the only value that definitively
           indicates remote Trickle ICE support, an application that
           compares canTrickleIceCandidates against "true" will by
-          default attempt Half Trickle on initial offers (as canTrickleIceCandidates will be null) and Full
+          default attempt Half Trickle on initial offers and Full
           Trickle on subsequent interactions with a Trickle
           ICE-compatible agent.</t>
         </section>
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -1483,7 +1483,7 @@ Suggested:
           gathered. As "true" is the only value that definitively
           indicates remote Trickle ICE support, an application that
           compares canTrickleIceCandidates against "true" will by
-          default attempt Half Trickle on initial offers and Full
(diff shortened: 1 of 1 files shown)


ekr (approved):
LGTM.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1488,7 +1483,7 @@ Suggested:
           gathered. As "true" is the only value that definitively
           indicates remote Trickle ICE support, an application that
           compares canTrickleIceCandidates against "true" will by
-          default attempt Half Trickle on initial offers and Full
+          default attempt Half Trickle on initial offers (as canTrickleIceCandidates will be null) and Full
           Trickle on subsequent interactions with a Trickle
```
juberti, line 1487:
```suggestion
          Trickle on subsequent interactions, when negotiating with a Trickle
```

juberti: Added this tweak to make it clear that we're talking about a trickle-compatible agent for both the initial and subsequent interactions. LMK if you agree.

juberti (approved):

draft-ietf-rtcweb-jsep.xml
ekr: Not sure it's necessary but it seems fine.

juberti: Rereading this change, I think this makes the sentence a bit clumsier and harder to understand, so at this point in time I think we should just keep the original text.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -1488,7 +1483,7 @@ Suggested:
           gathered. As "true" is the only value that definitively
           indicates remote Trickle ICE support, an application that
           compares canTrickleIceCandidates against "true" will by
-          default attempt Half Trickle on initial offers and Full
+          default attempt Half Trickle on initial offers (as canTrickleIceCandidates will be null) and Full
```
juberti, line 1486:
```suggestion
          default attempt Half Trickle on initial offers and Full
```

PR #1007, opened 2021-01-16
Don't use a=bundle-only for bundled sections in any offers or answers

### LINKED ISSUE #843: BUNDLE answers should contain bundle-only atttribs
per @cdh4u, seems like answers should now be handled differently:

> The initial offer is as before, i.e., you assign a unique address, SDP attributes etc to each bundled m- section.
>
> But, in the answer, and in every subsequent offer/answer, you only assign the BUNDLE port to the bundled m- section represented by the offerer/answerer BUNDLE-tag. To every other m- section you assign a zero port value and a bundle-only attribute (previously the bundle-only attribute was only defined for offers).

juberti:
Fixes #843. 

I tried to keep the changes minimal, so I kept the construct "marked as bundle-only", even though that marking now is conceptual rather than actual (i.e., in SDP).

Note that this leaves some ambiguity regarding the m= line port for bundled m= sections; it may be useful to spell that out directly, and perhaps also state that zero/a=bundle-only is not to be used.

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -511,7 +511,8 @@ setRemote(ANSWER) |                                   |
           needed by new/recycled/restarting "m=" sections; other "m="
           sections continue to use their existing candidates. Also, if
           an "m=" section is bundled (either by a successful bundle
-          negotiation or by being marked as bundle-only), then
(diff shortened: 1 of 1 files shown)


fluffy (approved):
LGTM

juberti:
Outdated by events.

PR #1015, opened 2021-01-19
Add Disclaimer as provided by Murray

ekr:


draft-ietf-rtcweb-jsep.xml
```diff
@@ -205,6 +205,37 @@
         example. Providing createOffer/createAnswer avoids this
         problem.</t>
       </section>
+      <section anchor="sec.disclaimer" numbered="true" toc="default">
+        <name>Contradiction regarding bundle-only "m=" sections</name>
```
juberti, line 209: Should this be in caps, like the other headings? @ajeanmahoney 
```suggestion
        <name>Contradiction Regarding Bundle-only "m=" Sections</name>
```

draft-ietf-rtcweb-jsep.xml
```diff
@@ -205,6 +205,37 @@
...
+        and <xref target="RFC8843" format="default"/> respectively)
+        Rather than delaying publication further to come
+        to a resolution, the documents are being published as they were
+        originally approved.  The IETF intends to restart work on these
+        technologies, and revised versions of these documents will be
+        published as soon as a resolution becomes available.
+        </t>
+        <t>
+        The specific issue involves the handling of "m=" sections that
+        are designated as bundle-only, as discussed in [rfced: This RFC], Section
```
juberti, line 223: should this be an xref?

ekr: Fine by me, but I will defer to Murray.

juberti: @mskucherawy

draft-ietf-rtcweb-jsep.xml
```diff
@@ -205,6 +205,37 @@
         example. Providing createOffer/createAnswer avoids this
         problem.</t>
       </section>
+      <section anchor="sec.disclaimer" numbered="true" toc="default">
+        <name>Contradiction Regarding Bundle-only "m=" Sections</name>
```
juberti, line 209: Based on the discussion, seems like my suggestion should be reverted.

juberti (approved):
Updated text with RFC Editor version, approving

PR #1017, opened 2021-07-10
max-bundle -> must-bundle

### LINKED ISSUE #843: BUNDLE answers should contain bundle-only atttribs

juberti:
Fixes #843

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *~
 lib
+venv
 foo
 bar
 old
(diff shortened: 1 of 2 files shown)


fluffy (approved):
LGTM

PR #1024, opened 2022-03-29
Update W3C ref

seanturner:
Addresses GENART review:
https://mailarchive.ietf.org/arch/msg/rtcweb/DcHcqjiwRZdfjZzbMJMDofd8YXU/

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -5081,7 +5081,7 @@ Datagram Transport Layer Security (DTLS) and Transport Layer Security (TLS)</tit
 <xi:include href="https://xml2rfc.ietf.org/public/rfc/bibxml/reference.RFC.3960.xml"/>
 
         <reference anchor="W3C.webrtc"
-                   target="https://www.w3.org/TR/2020/PR-webrtc-20201215/">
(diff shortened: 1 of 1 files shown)


ekr (approved):
LGTM. @fluffy @juberti ?

fluffy:
Just trying to see how this comes out in th .txt final file - still looking at this

fluffy:
In the .txt file, the refernce from this PR comes out like 


   [W3C.webrtc]
              Jennings, C., Ed., Boström, H., Ed., and J. Bruaroey, Ed.,
              "WebRTC 1.0: Real-time Communication Between Browsers",
              World Wide Web Consortium Recommendation, January 2021,
              <https://www.w3.org/TR/2021/REC-webrtc-20210126/>.

This looks correct to me as it is a stable reference to the REC.

fluffy (changes requested):
I agree with the change from the PR to the dated REC. 

I do not think it should be moved from informative to normative.

seanturner:
@fluffy see #1026.

juberti:
Sounds like we should close this in favor of #1026. I agree with Cullen that informative is probably the right category here as the W3C spec largely depends on JSEP, rather than vice versa.

seanturner:
Yes feel free to close this out in favor of #1026.

PR #1025, opened 2022-04-27
Addressing opsdir comments

seanturner:
Addressing [opsdir comments](https://mailarchive.ietf.org/arch/msg/rtcweb/6a-xbYS2dihJXRRqaGoSNDQVRNQ/).

### DIFF
--- a/draft-ietf-rtcweb-jsep.xml
+++ b/draft-ietf-rtcweb-jsep.xml
@@ -205,11 +205,11 @@
           When <xref target="RFC8829"/> was published, inconsistencies regarding BUNDLE
           <xref target="RFC8843"/> operation were identified with regard to
           both the specification text as well as implementation behavior. The
-          former concern was addressed via an update to <xref target="RFC8843"/>.
+          former concern was addressed via an update to BUNDLE, see <xref target="RFC9143"/>.
           For the latter concern, it was observed that some implementations 
           implemented the "max-bundle" bundle policy defined in <xref target="RFC8829"/>
           by assuming that bundling had already been negotiated, rather than marking "m=" sections
-          as bundle-only as indicated by the specification.
+          as bundle-only as indicated by the BUNDLE specification.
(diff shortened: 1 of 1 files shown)


draft-ietf-rtcweb-jsep.xml
```diff
@@ -205,11 +205,11 @@
           When <xref target="RFC8829"/> was published, inconsistencies regarding BUNDLE
           <xref target="RFC8843"/> operation were identified with regard to
           both the specification text as well as implementation behavior. The
-          former concern was addressed via an update to <xref target="RFC8843"/>.
+          former concern was addressed via an update to BUNDLE, see <xref target="RFC9143"/>.
```
ekr, line 208:
```suggestion
          former concern was addressed via an update to BUNDLE (see <xref target="RFC9143"/>).
```

ekr (approved):
LGTM modulo the change below. @juberti @fluffy

juberti (approved):
This seems like a good clarification, but there is still some confusion with 8843 vs 9143 references. I've added some suggested text to resolve.

draft-ietf-rtcweb-jsep.xml
```diff
@@ -205,11 +205,11 @@
           When <xref target="RFC8829"/> was published, inconsistencies regarding BUNDLE
           <xref target="RFC8843"/> operation were identified with regard to
```
juberti, line 206: I'm guessing we want to leave this one as RFC8843?

seanturner: I think this one we do want to leave alone because it is referring to the "old" version.

juberti: OK. I will update accordingly and merge.

juberti: PTAL at the updated text for this PR

juberti: @seanturner

